PR #2, opened 2019-07-30
check compiler page was given the expected PROT_FLAGS

carenas:
in HardenedBSD[1] and probably other security enhanced platforms, the
PROT_FLAGS passed to mmap won't be respected in some cases, leading
to the allocation of pages that lack execution permission and therefore
will trigger the application to segfault later.

detect the problem earlier by doing an mprotect of the recently allocated
page with the same permissions and if denied, discard it instead and
report an allocation problem upstream.

[1] https://github.com/HardenedBSD/hardenedBSD/wiki#pageexec-and-mprotect-aka-noexec

### DIFF
--- a/sljit_src/sljitExecAllocator.c
+++ b/sljit_src/sljitExecAllocator.c
@@ -137,6 +137,7 @@ static SLJIT_INLINE int get_map_jit_flag()
 static SLJIT_INLINE void* alloc_chunk(sljit_uw size)
 {
 	void *retval;
+	const int prot = PROT_READ | PROT_WRITE | PROT_EXEC;
 
 #ifdef MAP_ANON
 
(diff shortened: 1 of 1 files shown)


carenas:
this change was tested on HardenedBSD, macOS and Linux (Debian) and has the added benefit of only needing the MPROTECT "entitlement" when running as the "testing" will also set them straight even if mmap changed them behind our back

zherczeg:
Sljit has an executable allocator for hardened systems:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitProtExecAllocator.c

Its major flaw is that it does not support fork(), since the tmp file is not duplicated. Do you think it would be possible to fix this?

carenas:
> Do you think it would be possible to fix this?

no; at the end the application should control this process; trying to work around the restrictions in the library "transparently" is why the design won't work at the end.

PR #12, opened 2020-01-15
Hjl/cet/master

hjl-tools:
Intel Control-flow Enforcement Technology (CET):

https://software.intel.com/en-us/articles/intel-sdm

contains shadow stack (SHSTK) and indirect branch tracking (IBT).  When
SHSTK is enabled, return address on normal stack must match the one on
shadow stack.  When IBT is enabled, the indirect branch target must
start with ENDBR instruction.  When CET is enabled at the compile time
with "gcc -fcf-protection -mshstk", this patch updates sljit with:

1. Emit ENDBR32/ENDBR64 at function entry.  ENDBR32/ENDBR64 are NOPs on
non-CET machines.
2. Add SLJIT_ENDBR to sljit_emit_op0 to generate ENDBR32/ENDBR64 for x86
processors.
3. Before JIT return, if shadow stack is enabled, adjust shadow stack
so that the top of show stack matches the top of normal stack.  NB: It
is safe since shadow stack is read-only, you can only skip stack frame,
not jump to anywhere else.
4. Fix X86_32 emit_x86_instruction to properly encode (%ebp) so that
we can use TMP_REG1 as scratch register to unwind show stack in
sljit_emit_return.  NB: ECX is unusable since it isn't a scratch
register in sljit.

When indirect branch tracking (IBT) from Intel Control-flow Enforcement
Technology (CET) is enabled, all indirect branch targets must start with
ENDBR32/ENDBR64.  This patch calls sljit_emit_op0 (compiler, SLJIT_ENDBR)
to generate ENDBR32/ENDBR64 at indirect branch targets for x86 processors
in sljitTest.c and  regexJIT.c.

Tested with
```
$ CC="gcc -Wl,-z,cet-report=error -fcf-protection -mshstk" make
```
on x86-32 and x86-64 Linux CET machines.

### DIFF
--- a/sljit_src/sljitConfigInternal.h
+++ b/sljit_src/sljitConfigInternal.h
@@ -199,6 +199,14 @@
 #if defined(__CET__)
 #define SLJIT_CONFIG_X86_CET 1
 #endif
+#if (defined SLJIT_CONFIG_X86_CET && SLJIT_CONFIG_X86_CET)
+#if defined(__GNUC__)
+#if !defined (__SHSTK__)
+#error "-mshstk is needed to compile with -fcf-protection"
+#endif
+#include <x86intrin.h>
+#endif
+#endif
 #endif
 
 /**********************************/
--- a/sljit_src/sljitNativeX86_32.c
+++ b/sljit_src/sljitNativeX86_32.c
@@ -310,26 +310,39 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_return(struct sljit_compiler *comp
 		SLJIT_SP, 0, SLJIT_SP, 0, SLJIT_IMM, compiler->local_size));
 #endif
 
-	size = 2 + (compiler->scratches > 7 ? (compiler->scratches - 7) : 0) +
+	size = (compiler->scratches > 9 ? (compiler->scratches - 9) : 0) +
 		(compiler->saveds <= 3 ? compiler->saveds : 3);
+	if (size) {
+		inst = (sljit_u8*)ensure_buf(compiler, 1 + size);
+		FAIL_IF(!inst);
+
+		INC_SIZE(size);
+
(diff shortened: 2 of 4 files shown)


zherczeg:
I laso have another question:
If I have two functions, A and B, and recursively call them 3 times. The stack looks like the following:

`[stack top] main A B A B A B [current stack bottom]`

And I want to return to the third B function:

`[stack top] main A B [current stack bottom]`

Does the current code correctly figures this out?

hjl-tools:
> I think the SLJIT_ENDBR part is ready, it could even go in as a separate patch.

I will make it a separate patch.

> As for the rest, I still miss the control over CET. It might be possible that you need a shadow stack update before a fast return, and in other cases you don't need to do it when you return from a function. Would it be possible to create another op0 operation which can be emitted before a return if shadow stack update is needed?

For

sljit_emit_enter
...
sljit_emit_return

and

sljit_emit_fast_enter
...
sljit_emit_fast_return

We need to pop shadow stack if there are exceptions before return.   Shadow stack must
be adjusted right before "RET" when emitting return so that we know the location of return
address on stack.   We can add an option to sljit_emit_return and sljit_emit_fast_return to
indicate if shadow stack should be checked.

hjl-tools:
> I also have another question:
> If I have two functions, A and B, and recursively call them 3 times. The stack looks like the following:
> 
> `[stack top] main A B A B A B [current stack bottom]`
> 
> And I want to return to the third B function:
> 
> `[stack top] main A B [current stack bottom]`
> 
> Does the current code correctly figures this out?

Yes, it works fine.  You can add a testcase to sljitTest.c.

zherczeg:
Ok, I will create a new testcase. But only probably tomorrow.

zherczeg:
I have added the promised test. It even revealed some issues when code is generated for CPUs with delay slot.

PR #13, opened 2020-01-16
x86: Add SLJIT_ENDBR to sljit_emit_op0 to generate ENDBR32/ENDBR64

hjl-tools:
Intel Control-flow Enforcement Technology (CET):

https://software.intel.com/en-us/articles/intel-sdm

contains indirect branch tracking (IBT). When IBT is enabled, the
indirect branch target must start with ENDBR instruction.  When IBT
is enabled at the compile time, this patch does:

1. Emit ENDBR32/ENDBR64 at function entry.  ENDBR32/ENDBR64 are NOPs on
non-CET machines.
2. Add SLJIT_ENDBR to sljit_emit_op0 to generate ENDBR32/ENDBR64 for x86
processors.

PR #14, opened 2020-01-17
x86: Enable shadow stack support

hjl-tools:
Intel Control-flow Enforcement Technology (CET):

https://software.intel.com/en-us/articles/intel-sdm

contains shadow stack (SHSTK).  When SHSTK is enabled, return address on
normal stack must match the one on shadow stack.  When SHSTK is enabled
at the compile time, this patch updates sljit with:

1. Before JIT return, if shadow stack is enabled, adjust shadow stack
so that the top of show stack matches the top of normal stack.  NB: It
is safe since shadow stack is read-only, you can only skip stack frame,
not jump to anywhere else.
2. Fix X86_32 emit_x86_instruction to properly encode (%ebp) so that
we can use TMP_REG1 as scratch register to unwind show stack in
sljit_emit_return.  NB: ECX is unusable since it isn't a scratch
register in sljit as shown in sljit test:

$ ./bin/sljit_test

test51 case 2 failed
SLJIT tests: 1 (2%) tests are FAILED on x86 32bit (little endian + unaligned) ABI:fastcall (with fpu)

Tested with

$ CC="gcc -Wl,-z,cet-report=error -fcf-protection -mshstk" make

on x86-32 and x86-64 Linux CET machines.

test67 isn't compatible with CET since it uses sljit_emit_fast_return to
return from JIT.  We can't adust shadow stack in sljit_emit_fast_return
since it modifies flags, but sljit_emit_fast_return must preserve flags
as checked by test62.

### DIFF
--- a/sljit_src/sljitLir.c
+++ b/sljit_src/sljitLir.c
@@ -945,6 +945,7 @@ static const char* op2_names[] = {
 
 static const char* op_src_names[] = {
 	(char*)"fast_return",
+	(char*)"skip_frames_before_fast_return",
 };
 
 static const char* fop1_names[] = {
@@ -1317,10 +1318,12 @@ static SLJIT_INLINE CHECK_RETURN_TYPE check_sljit_emit_op_src(struct sljit_compi
 	sljit_s32 src, sljit_sw srcw)
 {
 #if (defined SLJIT_ARGUMENT_CHECKS && SLJIT_ARGUMENT_CHECKS)
-	CHECK_ARGUMENT(op >= SLJIT_FAST_RETURN && op <= SLJIT_FAST_RETURN);
+	CHECK_ARGUMENT(op >= SLJIT_FAST_RETURN
+		       && op <= SLJIT_SKIP_FRAMES_BEFORE_FAST_RETURN);
 	FUNCTION_CHECK_SRC(src, srcw);
 
-	if (op == SLJIT_FAST_RETURN)
+	if (op == SLJIT_FAST_RETURN
+	    || op == SLJIT_SKIP_FRAMES_BEFORE_FAST_RETURN)
 	{
 		CHECK_ARGUMENT(src != SLJIT_IMM);
 		compiler->last_flags = 0;
--- a/sljit_src/sljitLir.c
+++ b/sljit_src/sljitLir.c
@@ -926,7 +926,8 @@ static void sljit_verbose_fparam(struct sljit_compiler *compiler, sljit_s32 p, s
 
 static const char* op0_names[] = {
 	(char*)"breakpoint", (char*)"nop", (char*)"lmul.uw", (char*)"lmul.sw",
-	(char*)"divmod.u", (char*)"divmod.s", (char*)"div.u", (char*)"div.s"
+	(char*)"divmod.u", (char*)"divmod.s", (char*)"div.u", (char*)"div.s",
+	(char*)"endbr", (char*)"skip_frames_before_return"
 };
 
 static const char* op1_names[] = {
@@ -1162,9 +1163,10 @@ static SLJIT_INLINE CHECK_RETURN_TYPE check_sljit_emit_op0(struct sljit_compiler
 #if (defined SLJIT_ARGUMENT_CHECKS && SLJIT_ARGUMENT_CHECKS)
 	CHECK_ARGUMENT((op >= SLJIT_BREAKPOINT && op <= SLJIT_LMUL_SW)
 		|| ((op & ~SLJIT_I32_OP) >= SLJIT_DIVMOD_UW && (op & ~SLJIT_I32_OP) <= SLJIT_DIV_SW)
-		|| op == SLJIT_ENDBR);
+		|| (op >= SLJIT_ENDBR && op <= SLJIT_SKIP_FRAMES_BEFORE_RETURN));
 	CHECK_ARGUMENT(GET_OPCODE(op) < SLJIT_LMUL_UW || GET_OPCODE(op) >= SLJIT_ENDBR || compiler->scratches >= 2);
-	if (GET_OPCODE(op) >= SLJIT_LMUL_UW && GET_OPCODE(op) <= SLJIT_DIV_SW)
+	if ((GET_OPCODE(op) >= SLJIT_LMUL_UW && GET_OPCODE(op) <= SLJIT_DIV_SW)
+	    || op == SLJIT_SKIP_FRAMES_BEFORE_RETURN)
 		compiler->last_flags = 0;
 #endif
 #if (defined SLJIT_VERBOSE && SLJIT_VERBOSE)
--- a/sljit_src/sljitNativeX86_32.c
+++ b/sljit_src/sljitNativeX86_32.c
@@ -890,3 +890,37 @@ static sljit_s32 emit_fast_return(struct sljit_compiler *compiler, sljit_s32 src
 	RET();
 	return SLJIT_SUCCESS;
 }
+
+static sljit_s32 skip_frames_before_return(struct sljit_compiler *compiler)
+{
+	sljit_s32 size, base;
+	sljit_s32 has_f64_aligment;
+
+	SLJIT_ASSERT(compiler->args >= 0);
+	SLJIT_ASSERT(compiler->local_size > 0);
+
+#if !defined(__APPLE__)
+	has_f64_aligment = compiler->options & SLJIT_F64_ALIGNMENT;
+#else
+	has_f64_aligment = 0;
+#endif
+
+	size = compiler->local_size;
+	if (has_f64_aligment) {
+		/* Use ECX as base register to check return address.
+		   ECX is a scratch register for normal return. */
+		base = 3;
+		/* mov base, [esp + compiler->local_size].  */
+		EMIT_MOV(compiler, base, 0, SLJIT_MEM1(SLJIT_SP), size);
+		/* Return address is at [base + saveds_size]. */
+		size = 0;
+	} else
+		base = SLJIT_SP;
+
+	size += ((compiler->scratches > 9 ? (compiler->scratches - 9) : 0)
+		+ (compiler->saveds <= 3 ? compiler->saveds : 3))
+		* sizeof(sljit_uw) + sizeof(sljit_uw);
+
+	/* Adjust shadow stack if needed.  */
+	return adjust_shadow_stack(compiler, SLJIT_UNUSED, 0, base, size);
+}
--- a/sljit_src/sljitNativeX86_64.c
+++ b/sljit_src/sljitNativeX86_64.c
@@ -897,3 +897,19 @@ static sljit_s32 emit_mov_int(struct sljit_compiler *compiler, sljit_s32 sign,
 
 	return SLJIT_SUCCESS;
 }
+
+static sljit_s32 skip_frames_before_return(struct sljit_compiler *compiler)
+{
+	sljit_s32 tmp, size;
+
+	size = compiler->local_size;
+	tmp = compiler->scratches;
+	if (tmp >= SLJIT_FIRST_SAVED_REG)
+		size += (tmp - SLJIT_FIRST_SAVED_REG + 1) * sizeof(sljit_uw);
+	tmp = compiler->saveds < SLJIT_NUMBER_OF_SAVED_REGISTERS ? (SLJIT_S0 + 1 - compiler->saveds) : SLJIT_FIRST_SAVED_REG;
+	if (SLJIT_S0 >= tmp)
+		size += (SLJIT_S0 - tmp + 1) * sizeof(sljit_uw);
+
+	/* Adjust shadow stack if needed.  */
+	return adjust_shadow_stack(compiler, SLJIT_UNUSED, 0, SLJIT_SP, size);
+}
--- a/sljit_src/sljitNativeX86_common.c
+++ b/sljit_src/sljitNativeX86_common.c
@@ -702,6 +702,10 @@ static SLJIT_INLINE sljit_s32 emit_sse2_store(struct sljit_compiler *compiler,
 static SLJIT_INLINE sljit_s32 emit_sse2_load(struct sljit_compiler *compiler,
 	sljit_s32 single, sljit_s32 dst, sljit_s32 src, sljit_sw srcw);
 
+static sljit_s32 emit_cmp_binary(struct sljit_compiler *compiler,
+	sljit_s32 src1, sljit_sw src1w,
+	sljit_s32 src2, sljit_sw src2w);
+
 static SLJIT_INLINE sljit_s32 emit_endbranch(struct sljit_compiler *compiler)
 {
 #if (defined SLJIT_CONFIG_X86_CET && SLJIT_CONFIG_X86_CET)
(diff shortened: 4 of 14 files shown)


zherczeg:
Similar to #13, I would prefer this feature as user controlled. As for sljit_emit_return, we could use a flag which tells that the shadow stack needs to be cleaned. As for fast calls, we would need another function, which prepares the shadow stack before return, but does not return, so the user still can do some tasks (e.g. setting status flag). If I understand correctly, you can update the shadow stack anytime, the only thing matters is that when the return is executed, the shadow stack must know the location.

If you are ok with this change, I might delete the fast return function and create a new one where you can pass an opcode and has a source. I might move the prefetch instruction here as well later.

hjl-tools:
> Similar to #13, I would prefer this feature as user controlled. As for sljit_emit_return, we could use a flag which tells that the shadow stack needs to be cleaned. As for fast calls, we would need another function, which prepares the shadow stack before return, but does not return, so the user still can do some tasks (e.g. setting status flag). If I understand correctly, you can update the shadow stack anytime, the only thing matters is that when the return is executed, the shadow stack must know the location.

Adding a flag to sljit_emit_return should work.  A new function to prepare for fast return can
make shadow stack work.

I can only update shadow stack anytime if the return address is known or at a known location
on normal stack.  Since return address is passed to fast return, it isn't a problem.

> If you are ok with this change, I might delete the fast return function and create a new one where you can pass an opcode and has a source. I might move the prefetch instruction here as well later.

Yes, please.  Thanks.

zherczeg:
A patch is landed which introduces `check_sljit_emit_op_src`. You can add a new opcode here which updates the shadow stack according to the src value passed to this function. Would it be possible to add another `sljit_emit_op0` opcode which updates the shadow stack before `sljit_emit_return`? This way we would not need to add a flag (and we could do things after the shadow stack is updated but before we return).

Another question: do you use the ebp adressing on 32 bit frequently? Because if you don't, we could just emit the instruction directly, so there is no need to change the generic instruction emitter. I worry a bit about the FIXME comment.

hjl-tools:
> A patch is landed which introduces `check_sljit_emit_op_src`. You can add a new opcode here which updates the shadow stack according to the src value passed to this function. Would it be possible to add another `sljit_emit_op0` opcode which updates the shadow stack before `sljit_emit_return`? This way we would not need to add a flag (and we could do things after the shadow stack is updated but before we return).

I don't think `check_sljit_emit_op_src` help here since it can't tell me if shadow stack needs
adjustment.    `sljit_emit_op0` can work.  One can add it before return to adjust shadow
stack.

> Another question: do you use the ebp adressing on 32 bit frequently? Because if you don't, we could just emit the instruction directly, so there is no need to change the generic instruction emitter. I worry a bit about the FIXME comment.

I don't think (%ebp) will be used in other places and I can encode (%ebp) directly.

BTW, should it be an error when (%ebp/r13) is used in the generic instruction emitter?
It took me a while to figure what went wrong.

zherczeg:
> I don't think `check_sljit_emit_op_src` help here since it can't tell me if shadow stack needs
> adjustment. `sljit_emit_op0` can work. One can add it before return to adjust shadow
> stack.

As for fast return, the return address is not stored on the stack. It is provided in a register or memory location. This fits better for CPUs with a link register. I suspect you need to know the return address, hence I recommended to add a new opcode to `check_sljit_emit_op_src`. As for normal return, the return address is on the stack, you just need to calculate its position.

> BTW, should it be an error when (%ebp/r13) is used in the generic instruction emitter?
> It took me a while to figure what went wrong.

Yes, an SLJIT_ASSERT could be added here. As far as I know the r13 is not an issue, because the REX prefix enough to determine that a register >= 8 is used, and you don't need any special casing.

hjl-tools:
> > I don't think `check_sljit_emit_op_src` help here since it can't tell me if shadow stack needs
> > adjustment. `sljit_emit_op0` can work. One can add it before return to adjust shadow
> > stack.
> 
> As for fast return, the return address is not stored on the stack. It is provided in a register or memory location. This fits better for CPUs with a link register. I suspect you need to know the return address, hence I recommended to add a new opcode to `check_sljit_emit_op_src`. As for normal return, the return address is on the stack, you just need to calculate its position.

I added SLJIT_FAST_RETURN_SKIP_FRAMES and use it in in test67 to skip stack frames
for fast return.

> > BTW, should it be an error when (%ebp/r13) is used in the generic instruction emitter?
> > It took me a while to figure what went wrong.
> 
> Yes, an SLJIT_ASSERT could be added here. As far as I know the r13 is not an issue, because the REX prefix enough to determine that a register >= 8 is used, and you don't need any special casing.

I added SLJIT_ASSERT for unsupported (%esp) and hand coded "mov 0x0(%ebp),%ebp".
Everything works now.

sljit_src/sljitNativeX86_32.c
```diff
@@ -310,26 +310,39 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_return(struct sljit_compiler *comp
...
 	FAIL_IF(!inst);
-
 	INC_SIZE(size);
 
-	if (compiler->saveds > 0 || compiler->scratches > 11)
-		POP_REG(reg_map[SLJIT_S0]);
-	if (compiler->saveds > 1 || compiler->scratches > 10)
-		POP_REG(reg_map[SLJIT_S1]);
-	if (compiler->saveds > 2 || compiler->scratches > 9)
-		POP_REG(reg_map[SLJIT_S2]);
```
zherczeg: Do we need these changes? I thought the assertion below is enough.

sljit_src/sljitNativeX86_common.c
```diff
@@ -2220,6 +2363,8 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_op_src(struct sljit_compiler *comp
 	CHECK_EXTRA_REGS(src, srcw, (void)0);
 
 	switch (op) {
+	case SLJIT_FAST_RETURN_SKIP_FRAMES:
+		FAIL_IF(adjust_shadow_stack(compiler, src, srcw, 0));
```
zherczeg: Maybe I wasn't clear, but I want this as a separate operation, and not part of `SLJIT_FAST_RETURN`.

Ideally you could do the following:
sljit_emit_op_src(compiler, SLJIT_SKIP_FRAMES_BEFORE_FAST_RETURN, src, srcw);
/* Do other things, e.g. set flags, since SLJIT_SKIP_FRAMES_BEFORE_FAST_RETURN may destroy them. */
sljit_emit_op_src(compiler, SLJIT_FAST_RETURN, src, srcw);

This provides quite a flexibility since users can do the shadow stack update anytime, not when the return is actually performed.

sljit_src/sljitNativeX86_32.c
hjl-tools: We have

```
POP some registers    <-- Separate instruction buffer
Adjust shadow stack if needed <-- Separate instruction buffer
POP TMP_REG1
RETURN
```

When allocating the instruction buffer for `POP some registers`, the buffer size must be
precise.  There must be no unused space in  `POP some registers` instruction buffer.
The bug is latent since a single instruction buffer used to covere:

```
POP some registers    <-- Separate instruction buffer
POP TMP_REG1
RETURN
```

with some unused space after RETURN.

zherczeg: Ah, I am sorry. I thought this code is related to emit_x86_instruction. So we make `adjust_shadow_stack` mandatory for `sljit_emit_return`? Could we make it an op0 operation, which can be run by the user when it is necessary?

hjl-tools: adjust_shadow_stack must know exactly where return address is on stack.  How can a user
tell and pass the info to the op0 operation?

sljit_src/sljitNativeX86_common.c
hjl-tools: Fixed.

sljit_src/sljitNativeX86_32.c
zherczeg: The op0 has no operand. The point is, the user pass no info, but you can compute everything you need. See sljit_set_context:

https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitNativeX86_32.c#L247

`compiler->local_size` contains the stack frame size except the saved registers and the return address. The following code shows how the stack is adjusted before return:

https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitNativeX86_32.c#L302

So you can load the return address by adding a constant to esp:
mov ebp, [esp + compiler->local_size + saveds_size]

Note: compiler->local_size + saveds_size is a compile time constant.

saveds_size for x86-32:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitNativeX86_32.c#L278

When SLJIT_F64_ALIGNMENT flag is set, it is a bit more complicated because you need two loads:
mov ebp, [esp + compiler->local_size]
mov ebp, [ebp + saveds_size]

As for 64 bit, it is easier because you need only one load all the time (formulas are different though).

zherczeg: `saveds = (2 + (scratches > 9 ? (scratches - 9) : 0) + (saveds <= 3 ? saveds : 3)) * sizeof(sljit_uw);`

The 2 in this formula represents the return adress and ebp, so you might need to add only 1:
`saveds = (1 + (scratches > 9 ? (scratches - 9) : 0) + (saveds <= 3 ? saveds : 3)) * sizeof(sljit_uw);`

But please check, I just give the general idea, haven't checked it yet.

hjl-tools: Can you add a testcase with sljit_emit_return with frames skipped?  The current
testcases don't cover this case.

zherczeg: I have added a new test.

hjl-tools:
```
static void test68(void)
{
        /* Test skipping returns from fast calls (return type is normal). */
        executable_code code;
        struct sljit_compiler *compiler;
        struct sljit_jump *call, *jump;
        struct sljit_label *label;
        int i;

        if (verbose)
                printf("Run test67\n");
                                   ^^^^^ Should be test68.
````

Comment (zherczeg) [reply sljit_src/sljitNativeX86_32.c]: Thanks. I will fix it later.

Comment (hjl-tools) [reply sljit_src/sljitNativeX86_32.c]: I added SLJIT_SKIP_FRAMES_BEFORE_RETURN.   It passed testes.

File: sljit_src/sljitNativeX86_32.c [hunk 3]
```diff
@@ -370,6 +370,8 @@ static sljit_u8* emit_x86_instruction(struct sljit_compiler *compiler, sljit_s32
 	SLJIT_ASSERT((flags & (EX86_PREF_F2 | EX86_PREF_F3)) != (EX86_PREF_F2 | EX86_PREF_F3)
 		&& (flags & (EX86_PREF_F2 | EX86_PREF_66)) != (EX86_PREF_F2 | EX86_PREF_66)
 		&& (flags & (EX86_PREF_F3 | EX86_PREF_66)) != (EX86_PREF_F3 | EX86_PREF_66));
+	/* We don't support (%esp). */
+	SLJIT_ASSERT(!(b & SLJIT_MEM) || immb || reg_map[b & REG_MASK] != 5);
```
zherczeg: Isn't this in another patch?

sljit_src/sljitNativeX86_32.c
```diff
@@ -893,3 +895,36 @@ static sljit_s32 emit_fast_return(struct sljit_compiler *compiler, sljit_s32 src
...
+		/* Return address is at [base + saveds_size]. */
+		size = 0;
+	} else
+		base = SLJIT_SP;
+
+	size += ((compiler->scratches > 9 ? (compiler->scratches - 9) : 0)
+		 + (compiler->saveds <= 3 ? compiler->saveds : 3)) * 4 + 4;
+
+	/* Adjust shadow stack if needed.  */
+	return adjust_shadow_stack(compiler, SLJIT_UNUSED, 0, base, size);;
```
zherczeg: Single semicolon is enough.

sljit_src/sljitNativeX86_64.c
```diff
@@ -897,3 +897,19 @@ static sljit_s32 emit_mov_int(struct sljit_compiler *compiler, sljit_s32 sign,
...
 }
+
+static sljit_s32 skip_frames_before_return(struct sljit_compiler *compiler)
+{
+	sljit_s32 tmp, size;
+
+	size = compiler->local_size;
+	tmp = compiler->scratches;
+	if (tmp >= SLJIT_FIRST_SAVED_REG)
+		size += (tmp - SLJIT_FIRST_SAVED_REG + 1) * 8;
```
zherczeg: Use `sizeof(sljit_uw)` instead of 8.

sljit_src/sljitNativeX86_64.c
```diff
@@ -897,3 +897,19 @@ static sljit_s32 emit_mov_int(struct sljit_compiler *compiler, sljit_s32 sign,
...
+	tmp = compiler->saveds < SLJIT_NUMBER_OF_SAVED_REGISTERS ? (SLJIT_S0 + 1 - compiler->saveds) : SLJIT_FIRST_SAVED_REG;
+	if (SLJIT_S0 >= tmp)
+		size += (SLJIT_S0 - tmp + 1) * 8;
```
zherczeg: Ditto.

sljit_src/sljitLir.c
```diff
@@ -1161,9 +1161,10 @@ static SLJIT_INLINE CHECK_RETURN_TYPE check_sljit_emit_op0(struct sljit_compiler
 #if (defined SLJIT_ARGUMENT_CHECKS && SLJIT_ARGUMENT_CHECKS)
 	CHECK_ARGUMENT((op >= SLJIT_BREAKPOINT && op <= SLJIT_LMUL_SW)
 		|| ((op & ~SLJIT_I32_OP) >= SLJIT_DIVMOD_UW && (op & ~SLJIT_I32_OP) <= SLJIT_DIV_SW)
-		|| op == SLJIT_ENDBR);
+		|| (op >= SLJIT_ENDBR && op <= SLJIT_SKIP_FRAMES_BEFORE_RETURN));
 	CHECK_ARGUMENT(GET_OPCODE(op) < SLJIT_LMUL_UW || GET_OPCODE(op) >= SLJIT_ENDBR || compiler->scratches >= 2);
-	if (GET_OPCODE(op) >= SLJIT_LMUL_UW && GET_OPCODE(op) <= SLJIT_DIV_SW)
+	if ((GET_OPCODE(op) >= SLJIT_LMUL_UW && GET_OPCODE(op) <= SLJIT_DIV_SW)
+	    || op == SLJIT_SKIP_FRAMES_BEFORE_RETURN)
```
zherczeg: Please add the name to `op0_names`

sljit_src/sljitLir.c
```diff
@@ -1317,10 +1318,12 @@ static SLJIT_INLINE CHECK_RETURN_TYPE check_sljit_emit_op_src(struct sljit_compi
...
 {
 #if (defined SLJIT_ARGUMENT_CHECKS && SLJIT_ARGUMENT_CHECKS)
-	CHECK_ARGUMENT(op >= SLJIT_FAST_RETURN && op <= SLJIT_FAST_RETURN);
+	CHECK_ARGUMENT(op >= SLJIT_FAST_RETURN
+		       && op <= SLJIT_SKIP_FRAMES_BEFORE_FAST_RETURN);
 	FUNCTION_CHECK_SRC(src, srcw);
 
-	if (op == SLJIT_FAST_RETURN)
+	if (op == SLJIT_FAST_RETURN
+	    || op == SLJIT_SKIP_FRAMES_BEFORE_FAST_RETURN)
```
zherczeg: Please add the name to `op_src_names`

sljit_src/sljitNativeX86_32.c
hjl-tools: Fixed.

sljit_src/sljitNativeX86_32.c
hjl-tools: Fixed.

sljit_src/sljitNativeX86_64.c
hjl-tools: Fixed,

sljit_src/sljitLir.c
hjl-tools: Fixed.

sljit_src/sljitLir.c
hjl-tools: Fixed.

sljit_src/sljitNativeX86_64.c
hjl-tools: Fixed.

zherczeg:
I have added some updates, please check them.

One more thing came to my mind. We silently overwrite ecx for x86-32 in some cases. Could we push/pop this register?

zherczeg:
`[esp+0]` is also reserved for temporaries, we could also use this memory location.

hjl-tools:
> `[esp+0]` is also reserved for temporaries, we could also use this memory location.

See https://github.com/zherczeg/sljit/pull/19

PR #16, opened 2020-01-19
Patches for VS2019

kiyolee:
A few patches to get things compile with VS2019.
1. VS2019 is more stringent about error C4146.
2. A minor compiler warning when calling sljit_set_target() in sljitTest.c.
3. One of the macros would generate code like "if (true) {}", suppressing C4127 to silence that.


test_src/sljitTest.c
```diff
@@ -31,6 +31,11 @@
 #include <stdlib.h>
 #include <string.h>
 
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable: 4127) /* conditional expression is constant */
+#endif
```
zherczeg: As far as I know VSStudio is an IDE, can't you set this in the IDE?

kiyolee:
More precisely I should refer to MSVC instead of VS. You can indeed set that in vcxproj, but if you use cl.exe directly to compile, you will still see the same warning. The #pragma is to suppress the warning however you compile the file.

Sent on the go with Vodafone
Get Outlook for Android<https://aka.ms/ghei36>

________________________________

zherczeg:
Ok, it does not seem to affect other systems.

PR #19, opened 2020-01-24
X86-32: Preserve ECX when skipping frames for return

hjl-tools:
We need to preserve ECX when skipping frames for normal return.  Use
[esp] to save and restore ECX.

### DIFF
--- a/sljit_src/sljitNativeX86_32.c
+++ b/sljit_src/sljitNativeX86_32.c
@@ -895,8 +895,12 @@ static sljit_s32 emit_fast_return(struct sljit_compiler *compiler, sljit_s32 src
 
 static sljit_s32 skip_frames_before_return(struct sljit_compiler *compiler)
 {
-	sljit_s32 size, base;
+	sljit_s32 size, saved_size;
 	sljit_s32 has_f64_aligment;
+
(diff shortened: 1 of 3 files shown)


sljit_src/sljitNativeX86_32.c
```diff
@@ -909,9 +909,20 @@ static sljit_s32 skip_frames_before_return(struct sljit_compiler *compiler)
 
 	size = compiler->local_size;
 	if (has_f64_aligment) {
-		/* Use ECX as base register to check return address.
-		   ECX is a scratch register for normal return. */
-		base = 3;
+		/* Find a scratch register for normal return. */
```
zherczeg: I have an idea which could eliminate all scratch register uses. Move the return value to [esp]
Intel style code:
```
mov ebp, [esp + local_size]
mov ebp, [ebp + saved_registers_size]
mov [esp], ebp
```
Then you can use:
adjust_shadow_stack(compiler, SLJIT_UNUSED, 0, SLJIT_SP, size)

Actually you can remove the base argument, since it will be always SLJIT_SP.

hjl-tools: Fixed.

hjl-tools:
> LGTM

Thanks.

PR #21, opened 2020-01-30
VS2019 patches

kiyolee:
1. Fix VS2019 compile error C4146.
2. Mark some unused arguments to suppress compile warnings.

### DIFF
--- a/test_src/sljitTest.c
+++ b/test_src/sljitTest.c
@@ -5438,7 +5438,7 @@ static void test56(void)
 	sljit_emit_op2(compiler, SLJIT_ADD32 | SLJIT_SET_OVERFLOW, SLJIT_R0, 0, SLJIT_R0, 0, SLJIT_IMM, -(91 << 12));
 	sljit_emit_op_flags(compiler, SLJIT_MOV, SLJIT_MEM1(SLJIT_S0), 11 * sizeof(sljit_sw), SLJIT_OVERFLOW32);
 
-	sljit_emit_op1(compiler, SLJIT_MOV32, SLJIT_R0, 0, SLJIT_IMM, (sljit_sw)-0x80000000);
+	sljit_emit_op1(compiler, SLJIT_MOV32, SLJIT_R0, 0, SLJIT_IMM, -0x7fffffff-1);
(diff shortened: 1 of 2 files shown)


test_src/sljitTest.c
```diff
@@ -5438,7 +5438,7 @@ static void test56(void)
 	sljit_emit_op2(compiler, SLJIT_ADD32 | SLJIT_SET_OVERFLOW, SLJIT_R0, 0, SLJIT_R0, 0, SLJIT_IMM, -(91 << 12));
 	sljit_emit_op_flags(compiler, SLJIT_MOV, SLJIT_MEM1(SLJIT_S0), 11 * sizeof(sljit_sw), SLJIT_OVERFLOW32);
 
-	sljit_emit_op1(compiler, SLJIT_MOV32, SLJIT_R0, 0, SLJIT_IMM, (sljit_sw)-0x80000000);
+	sljit_emit_op1(compiler, SLJIT_MOV32, SLJIT_R0, 0, SLJIT_IMM, -(sljit_sw)0x80000000);
```
zherczeg: This was in your previous patch but this is incorrect. 0x80000000 does not fit into a signed 32-bit integer, so GCC rightfully complains that `0x80000000` cannot be converted to sljit_sw. However `-0x80000000` is a perfectly valid int32 number (actually it is the lowest value).

kiyolee: I guess the problem for VS2019 is that 0x80000000 is considered being unsigned integer and so apply minus operator would cause the compile error. An alternative way to fix this for VS2019 is to remove the minus sign which is kind of redundant as 0x80000000 == -0x80000000. Also checked even the casting (sljit_sw) is unnecessary, VS2019 happily accept the plain value 0x80000000 without any warning.

kiyolee: Ah, for GCC is -0x80000000 is an integer value by itself. VS2019 consider that as minus operator operated on an unsigned integer value. Interesting.

kiyolee: Found an alternative way to please VS2019 while maintaining logical accuracy. It is to write the value as `(-0x7fffffff-1)`. Looks like this is how INT_MIN is defined in Windows, Linux and FreeBSD, perhaps for similar reason to please various compilers.

zherczeg: That looks ok.

kiyolee:
Updated the patch with better way to fix C4146.

PR #29, opened 2020-04-06
netbsd: implement non file backed alternative for sljitProtExecAllocator

carenas:
the experimental sljitProtExecAllocator uses file backed maps to
workaround W^X policies and is therefore incompatible with fork()

NetBSD allows anonymous pages to have multiple pages with different
permissions by using mremap with the MAP_REMAPDUP flag so use that
instead.

### DIFF
--- a/sljit_src/sljitProtExecAllocator.c
+++ b/sljit_src/sljitProtExecAllocator.c
@@ -96,10 +96,9 @@ struct chunk_header {
 #endif
 #endif
 
+#if !(defined(__NetBSD__) && defined(MAP_REMAPDUP))
 int mkostemp(char *template, int flags);
-#if !defined(__NetBSD__)
 char *secure_getenv(const char *name);
-#endif
 
 static SLJIT_INLINE int create_tempfile(void)
 {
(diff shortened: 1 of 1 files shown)


sljit_src/sljitProtExecAllocator.c
```diff
@@ -204,6 +203,36 @@ static SLJIT_INLINE struct chunk_header* alloc_chunk(sljit_uw size)
...
 	return retval;
 }
+#else
+static SLJIT_INLINE struct chunk_header* alloc_chunk(sljit_uw size)
+{
+	struct chunk_header *retval;
+	void *maprx;
+
+	retval = (struct chunk_header *)mmap(NULL, size,
+			PROT_MPROTECT(PROT_EXEC|PROT_WRITE|PROT_READ),
```
zherczeg, line 213: Can you set all those flags in a secure environment?

sljit_src/sljitProtExecAllocator.c
```diff
@@ -204,6 +203,36 @@ static SLJIT_INLINE struct chunk_header* alloc_chunk(sljit_uw size)
...
+			MAP_ANON, -1, 0);
+
+	if (retval == MAP_FAILED)
+		return NULL;
+
+	maprx = mremap(retval, size, NULL, size, MAP_REMAPDUP);
```
zherczeg, line 219: Is this code NetBSD spcefic, or any system with MAP_REMAPDUP can be supported (at least in theory)?

sljit_src/sljitProtExecAllocator.c
carenas: this call is not asking for those permissions yet, indeed the mmap will return 
a page without any permissions and those will be set by mprotect below.

PROT_MPROTECT is specially designed for NetBSD's secure environment (available since version 8) and indicates an "intention", it is also NetBSD specific and documented for this specific use on its man pages for man and mremap

sljit_src/sljitProtExecAllocator.c
carenas: MAP_REMAPDUP is NetBSD specific and available since version 8, hopefully other systems could reuse this code if they implement MAP_REMAPDUP or a equivalent but none is available as of now, hence why the whole code segment is ifdef to __NetBSD__

indeed, even in NetBSD, version 7 (soon to be EOL) or older don't have MAP_REMAPDUP and will use the original file based code instead (tested with 7.2 in i386)

zherczeg: In the original code there was a guard around `secure_getenv`.  Is this not needed in older NetBSD systems anymore?

carenas: nothing has changed regarding portability of that function, which is why that old code guard is still needed and will be used in NetBSD 7 or older.

a slightly more sophisticated version of it will be needed for OpenBSD or HardenedBSD but that change is orthogonal to this one

Issue #31, opened 2020-04-08
Use after free when sljit_free_compiler


carenas:
seems to be easily reproducible in test57, with the problem "going away" by moving the call to the end of the function as shown by c66c38f586b755844ebe39956288336bfb63ff2f in the asan branch from my fork

instrumenting the code to print the pointers that are being passed in the loop in sljitLir.c:473 shows the problem is being triggered by the call (in a Fedora Rawhide x86_64 with asan) :

```
Run test57
0x6210000dd500
=================================================================
==151==ERROR: AddressSanitizer: heap-use-after-free on address 0x6210000dd518 at pc 0x000000428e30 bp 0x7ffe3fb99060 sp 0x7ffe3fb99050
READ of size 8 at 0x6210000dd518 thread T0
    #0 0x428e2f in sljit_get_label_addr sljit_src/sljitLir.h:1384
    #1 0x428e2f in test57 test_src/sljitTest.c:5515
    #2 0x428e2f in sljit_test test_src/sljitTest.c:6652
    #3 0x7feb57130041 in __libc_start_main (/lib64/libc.so.6+0x27041)
    #4 0x4013cd in _start (/usr/src/bin/sljit_test+0x4013cd)

```
the issue persists, even if building with SLJIT_SINGLE_THREADED

```diff
diff --git a/sljit_src/sljitLir.c b/sljit_src/sljitLir.c
index 86772cc..7bcc882 100644
--- a/sljit_src/sljitLir.c
+++ b/sljit_src/sljitLir.c
@@ -462,6 +462,7 @@ SLJIT_API_FUNC_ATTRIBUTE void sljit_free_compiler(struct sljit_compiler *compile
        while (buf) {
                curr = buf;
                buf = buf->next;
+               printf("%p\n", curr);
                SLJIT_FREE(curr, allocator_data);
        }
```

carenas:
it will seem the bug was already there even before the last changes, but was probably not easy to notice as in most cases it would still work.

the original issue was noticed in OpenBSD which is a little more strict on its own memory management and was therefore failing the test

carenas:
fix included as third patch (586a48113abd9302a087ef015174d924ce330189) in branch compat in my fork

zherczeg:
Thank you for finding this issue. Your analysis is correct, the test has a bug. Could you open a PR with the proposed patch?

carenas:
[...]

Issue #1, opened 2019-04-20
How to reproduce PCRE 8.42 buffer overread in JIT?


hityc2019:
I have some question in  recurring vulnerability in the webisite : https://vcs.pcre.org/pcre/code/trunk/pcre_jit_compile.c?r1=1740&r2=1739&pathrev=1740 .
 It is hard to understand how to cause buffer overflow. 
Could you please provide test code or analyze the cause of the vulnerability in detail? 
Thanks!

zherczeg:
I think this was discussed in the pcre dev channels. May I close this issue?

hityc2019:
Of course you can!

PR #32, opened 2020-04-08
test: fix test57 not to use freed memory

carenas:
otherwise failing in SE systems (ex: OpenBSD and HardenedBSD)
as well as reported by asan as shown by:

==151==ERROR: AddressSanitizer: heap-use-after-free on address 0x6210000dd518 at pc 0x000000428e30 bp 0x7ffe3fb99060 sp 0x7ffe3fb99050
READ of size 8 at 0x6210000dd518 thread T0
    #0 0x428e2f in sljit_get_label_addr sljit_src/sljitLir.h:1384
    #1 0x428e2f in test57 test_src/sljitTest.c:5515
    #2 0x428e2f in sljit_test test_src/sljitTest.c:6652
    #3 0x7feb57130041 in __libc_start_main (/lib64/libc.so.6+0x27041)
    #4 0x4013cd in _start (/usr/src/bin/sljit_test+0x4013cd)

label addresses must be copied out for using as they will be
wiped out once the compiler is destroyed.

### DIFF
--- a/test_src/sljitTest.c
+++ b/test_src/sljitTest.c
@@ -5479,6 +5479,8 @@ static void test57(void)
 	executable_code code;
 	struct sljit_compiler* compiler = sljit_create_compiler(NULL);
 	struct sljit_label* labels[5];
+	sljit_p addr[5];
+	int i;
 
 	if (verbose)
 		printf("Run test57\n");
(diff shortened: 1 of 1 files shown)


carenas:
fixes #31 

tested in OpenBSD 6.5, 6.6 (amd64) and HardenedBSD HEAD (amd64) with additional patches
tested in Fedora Rawhide (x86_64) with ASAN

PR #33, opened 2020-04-08
Add compatibility shim for using secure_getenv in BSD

carenas:
sets the basic framework by adding an autoconf compatible define that could be used to indicate a secure_getenv implementation from the system should be used instead;  NetBSD and OpenBSD are instructed to use a replacement function in the first patch with HardenedBSD (AKA FreeBSD) added in the third.

the intermediate patch is a refactoring of the function that creates the tmpfile and that addresses:

* makes sure TMPDIR is a working option to minimize the risk of failing to allocate the executable map
* avoids calling open (with O_TMPFILE) with an "" directory


sljit_src/sljitProtExecAllocator.c
```diff
@@ -96,51 +102,47 @@ struct chunk_header {
...
+			}
+		}
+	}
+#ifdef P_tmpdir
+	if (!tmp_name_len) {
+		len = strlen(P_tmpdir);
+		if (len > 0 && len < sizeof(tmp_name)) {
+			strcpy(tmp_name, P_tmpdir);
 			tmp_name_len = len;
 		}
```
carenas, line 134: this would be nicer using instead strlcpy

sljit_src/sljitProtExecAllocator.c
```diff
@@ -96,51 +102,47 @@ struct chunk_header {
...
+		strcpy(tmp_name, "/tmp");
+		tmp_name_len = 4;
+	}
 
 	SLJIT_ASSERT(tmp_name_len > 0 && tmp_name_len < sizeof(tmp_name));
 
-	while (tmp_name_len > 0 && tmp_name[tmp_name_len - 1] == '/') {
-		tmp_name_len--;
-		tmp_name[tmp_name_len] = '\0';
-	}
```
carenas, line 146: why is this code needed instead of just removing a trailing '/'?

at least the new version won't convert "////" -> ""

sljit_src/sljitProtExecAllocator.c
```diff
@@ -84,7 +84,13 @@ struct chunk_header {
        as it only uses local variables
 */
 
+#if !(defined(__NetBSD__) && defined(MAP_REMAPDUP))
```
carenas, line 87: this assumes `#include <sys/mman.h>` previously

sljit_src/sljitProtExecAllocator.c
```diff
@@ -96,51 +102,47 @@ struct chunk_header {
...
-
+	struct stat st;
 	char tmp_name[256];
-	size_t tmp_name_len;
+	size_t tmp_name_len = 0;
 	char *dir;
 	size_t len;
 
-#ifdef P_tmpdir
-	len = (P_tmpdir != NULL) ? strlen(P_tmpdir) : 0;
```
carenas, line 113: P_tmpdir is a macro (AKA a string literal) not a (char *) therefore couldn't be NULL

zherczeg:
This looks like a lot of effort for an allocator which is experimental and never intended to support all systems (and does not support fork). Lets discuss these changes in detail, because it seems the patch contains multiple things.

1) Shall we use getenv instead of the secure version? Or systems without secure_getenv can use the normal allocator.
2) `at least the new version won't convert "////" -> ""` why is that a problem?

sljit_src/sljitConfig.h
```diff
@@ -59,6 +59,13 @@ extern "C" {
 /* #define SLJIT_CONFIG_AUTO 1 */
 /* #define SLJIT_CONFIG_UNSUPPORTED 1 */
 
+/* --------------------------------------------------------------------- */
+/*  Compat                                                               */
+/* ----------------------------------------------------------------------*/
+
+/* Define to 1 if your system has secure_getenv function */
+/* #undef HAVE_SECURE_GETENV */
```
zherczeg, line 67: This file is for sljit configuration, not system cability configuration. This should be done autmatically if needed.

carenas: this is just a fallback from when sljit is compiled standalone or it is used inside a codebase that doesn't have a facility for system configuration.

when sljit is used as part of another code that has that capability (ex: pcre) then those variables will be set automatically and used accordingly.

tested with both ./configure and cmake, but for the later you will need to use pcre trunk

zherczeg: There is a section `Custom defines` above for this purpose.

carenas: I was considering "custom defines" as something that should be always empty upstream and only to be used for application (meaning the one that is include sljit, not the library itself, which in this case includes the ProtExecAllocator) specific code.

indeed, my intention is to use this section more as you could see in carenas/sljit@041f6b7eb7a1f43f8e1b9bb31e666fdbe756b22b which is a fix for pcre bug 2455

carenas:
1. moving from secure_getenv to getenv() degrades the security (ex: would allow any user to create files as root by pointing setting TMPDIR and then invoking a suid application using sljit)
2. passing "" (empty string) as the first parameter for open(O_TMPFILE) fails with -1; that is the case for example if TMPDIR=/ which is otherwise a valid input and wouldn't get mangled with the new code, but would with the old one

zherczeg:
What about making `secure_getenv` mandatory? If a system does not have it, it should use the normal allocator which does not allocate files. I am worried about adding a lot of new includes and a function which might be present.

carenas:
FWIW the design allows removing the compat code completely once the real function is available in the OS (for a better example of how that would work can see the shim for memfd_create in carenas/sljit@ef0c27bc4e50c63bc6de983447503b1b72142fee.

note also that each OS is enable explicitly so any other system trying to build this code will just get a link error (for Android I even make it explicit, so if the user doesn't pass "-DHAVE_MEMFD_CREATE" either in the command line or most likely from a config include managed with autotools or cmake, then the build just breaks)

agree that the shim adds complexity and need maintenance but the cat was already out of the bag when the original #ifdef was added for NetBSD (which luckily now will be only used in the soon to be EOL 7.x) but in the other hand will allow more users/platforms to run this code (hopefully, understanding the trade offs), after all and against most advice opensuse and netbsd are currently building pcre using this experimental allocator in their default packages, and I wouldn't be surprised there might be more

zherczeg:
My problems are not technical, rather design. The string handling fxes are ok, as long as no new include files are needed. But I really don't want to define any public (exported) function without sljit prefix in the code. It would be better if we could recommend a helper library which provides these functions. I think providing missing libc functions is not the task of a jit library, but I suspect these functions are needed for many other software tools so somebody has already created a "libc supplement" for BSD systems. Maybe it is even a core library.

zherczeg:
This seems created for exactly this purpose: https://www.gnu.org/software/gnulib/

carenas:
FWIW gnulib is GPL so it might not be a ideal source from code to put together with BSD, but agree that sljit is not the place to implement a compatibility library, which is why by design I was expecting the "shim" to be compiled out as soon as the user provided the right defines and were mostly there only to prevent the build to break and as an "available fallback" and mostly meant to discourage people reacting to a broken build with a patch like the one that degraded security in NetBSD by ignoring secure_getenv()

note that the fact the defines where autotools compatible also meant that they will be removing the shim code as soon as some sort of configuration applied to the code could consider it unnecessary, and the fact they were tied to specific systems, meant they won't silently "patch" the code in systems that don't really need them (ex: Solaris, macOS, Windows) where the build will be broken.

finally, the added headers are mostly there to make sure the right functions are included and are most likely no ops (since the code built before (maybe except for stdio.h which defines P_tmpdir and might had not been effective before in non debug builds or string.h which then will end up using the gcc builtin strcpy instead)

carenas:
superseded by #37 which itself is a WIP and likely to be superseded by #34 and #39

PR #34, opened 2020-04-10
protexec: allow using memfd_create for file backing creation

carenas:
memfd_create is available in linux since kernel 3.17 with support
added in glibc (2.27), musl (1.1.20) and bionic (30).

it allows creating a tmpfs backed file without having to worry about
file system support, path location or permission or even file name
so is ideal for the kind of file that the ProtExecutionAllocator
needs to back its maps.

this solves the problemss reported in PCRE's bug 2455[1] where a JIT
enabled call that was compiled to use this experimental allocator
will fail because the selected TMPDIR was non executable (ironically
something commong in SE systems)

### DIFF
--- a/sljit_src/sljitProtExecAllocator.c
+++ b/sljit_src/sljitProtExecAllocator.c
@@ -108,6 +108,13 @@ static SLJIT_INLINE int create_tempfile(void)
 	char *dir;
 	size_t len;
 
+#ifdef HAVE_MEMFD_CREATE
+	/* this is a GNU extension, make sure to use -D_GNU_SOURCE */
+	fd = memfd_create("sljit", MFD_CLOEXEC);
+	if (fd != -1)
+		return fd;
+#endif
+
 #ifdef P_tmpdir
 	len = (P_tmpdir != NULL) ? strlen(P_tmpdir) : 0;
 


sljit_src/sljitProtExecAllocator.c
```diff
@@ -109,6 +109,13 @@ static SLJIT_INLINE int create_tempfile(void)
 	char *dir;
 	size_t len;
 
+#ifdef HAVE_MEMFD_CREATE
+	/* this is a GNU extension, make sure to use -D_GNU_SOURCE */
+	fd = memfd_create("sljit", MFD_CLOEXEC);
+	if (fd != -1)
+		return fd;
+#endif
```
zherczeg, line 117: Shouldn't we return regardless of the result? If we continue the execution we just create another tmp file.

zherczeg:
According to, http://man7.org/linux/man-pages/man2/memfd_create.2.html , the file is visible in /proc/self/fd/. Only the name or the content of the file? Can it be read or write somehow?

zherczeg:
Anyway this new api sounds like a progress from Linux side, although the BSD mmap dup is still a much nicer solution for multiple mapping. I don't understand why they didn't add fork support for a memory mapped file :(

sljit_src/sljitProtExecAllocator.c
carenas: if fd == -1, then no file was created and then we continue with the fallback to try with open and if that fails as well once more with mkostemp.

note the same logic was used for the test with open(O_TMPFILE)

carenas:
the content is read/write through the mmap by the application.

externally (just like the other backing files) the path in /proc/self/fd could be open as well if having the right permissions

zherczeg:
This mode seems superior to the previous method, we could make it mandatory. This allocator does not need to work on any old sytem. Maybe at least we could drop the getenv support, which reduces external influence.

carenas:
implemented in #35, but be aware this will effectively break NetBSD 7.2 and some old versions of linux that were working before and as such should be spelled out in any release notes IMHO

and additional commit will be needed to help with anyone that will need to provide their own compat layer as well to (at least for PCRE the closing of bug 2553).

zherczeg:
I am ok with this change. Seems a small enough change. We could add a define for secure_getenv on BSD since the function seems a one liner. This could be enough for the next release of pcre. We can think about more complex changes afterwards.

carenas:
in line with #38, something like carenas/sljit@ f7b94b9cac37a7175b9f23420c50d8dda3cd2668 will be eventually needed on top

presume you mean that and the refactoring should wait until after the release of PCRE 10.35?, as for the getenv change will prepare yet another pull request then with what I hope would be a suitable approach based on your comments

PR #35, opened 2020-04-11
protexec: use memfd_create for file backing creation

carenas:
memfd_create is available in linux since kernel 3.17 with support
added in glibc (2.27), musl (1.1.20) and bionic (30).

it allows creating a tmpfs backed file without having to worry about
file system support, path location, permission or even file name
so is ideal for the kind of file that the ProtExecutionAllocator
needs to back its maps.

this solves the problems reported in PCRE's bug 2455[1] where a JIT
enabled call that was compiled to use this experimental allocator
will fail because the selected TMPDIR was non executable (ironically
something common in SE systems)

[1] https://bugs.exim.org/show_bug.cgi?id=2455

### DIFF
--- a/sljit_src/sljitProtExecAllocator.c
+++ b/sljit_src/sljitProtExecAllocator.c
@@ -83,98 +83,14 @@ struct chunk_header {
        as it only uses local variables
 */
 
-#include <fcntl.h>
-
-#ifndef O_NOATIME
-#define O_NOATIME 0
-#endif
-
-#ifdef __O_TMPFILE
-#ifndef O_TMPFILE
-#define O_TMPFILE	(__O_TMPFILE | O_DIRECTORY)
-#endif
-#endif
-
 #if !(defined(__NetBSD__) && defined(MAP_REMAPDUP))
-int mkostemp(char *template, int flags);
-char *secure_getenv(const char *name);
-
-static SLJIT_INLINE int create_tempfile(void)
-{
-	int fd;
-
-	char tmp_name[256];
-	size_t tmp_name_len;
-	char *dir;
-	size_t len;
-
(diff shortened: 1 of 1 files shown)


zherczeg:
Wow this is so much cleaner! You don't even need an fd.

I would not want to break recent systems, so what about keeping the old code, but marking it as obsolote and not recomennding it unless you have no other choice. This old code could be guarded by a `SLJIT_LEGACY_PROT_ALLOCATOR`. We could allow it to be the default for now (unless you have memfd_create), but  at some point it needs to be explicitly enabled when most OSes has a better choice.

carenas:
an alternative for older linux systems is a compatibility layer like the one in carenas/sljit@7da2cd8d63e8cb5d98cc0ca1ca8b8f2e0a8cf7b3

interestingly while testing it found that FreeBSD 13 will also have memfd_create

zherczeg:
It would be good to have a compatibility library which is dedicated to port libc to old systems (similar to glib) but with bsd license. This would be a good project.

carenas:
since we are going to be keeping create_tempfile() then it would be better to merge #34, but considering the bugs that were raised in #33 it might be better add it on top of something like my branch carenas/getenv@46982816c6534b4b4d2a8fa02ce01b29667a6a36.

note that none of those 3 commits have a guard like you suggested, as IMHO would make the code more difficult to follow and they don't seem to need it (after the two HAVE*) guards keep the code for building (even without the need for any compat code) but let me know it that makes sense.

alternatively we could have 3 different PRs (2 each compatibility function and 1 for the refactoring)

carenas:
implemented the pull with 3 commits in #37

PR #37, opened 2020-04-12
protexec: refactor create_tempfile()

carenas:
refactor create_tempfile to address:

- restrict additional dependencies to NetBSD < 8 or Linux
- error checking for TMPDIR to avoid failing to create the map files when invalid
- cleaner logic for failback so there are not double copies
- make sure P_tmpdir is used even in non debug/verbose builds
- simpler trimming logic for trailing '/' that keeps a working path to avoid breaking the open call

### DIFF
--- a/sljit_src/sljitProtExecAllocator.c
+++ b/sljit_src/sljitProtExecAllocator.c
@@ -83,21 +83,20 @@ struct chunk_header {
        as it only uses local variables
 */
 
+#if !(defined(__NetBSD__) && defined(MAP_REMAPDUP))
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
 
 #ifndef O_NOATIME
 #define O_NOATIME 0
 #endif
 
-#ifdef __O_TMPFILE
+/* this is a linux extension available since kernel 3.11 */
 #ifndef O_TMPFILE
-#define O_TMPFILE	(__O_TMPFILE | O_DIRECTORY)
+#define O_TMPFILE 020200000
 #endif
-#endif
-
-#if !(defined(__NetBSD__) && defined(MAP_REMAPDUP))
-int mkostemp(char *template, int flags);
 
 #ifdef __NetBSD__
 /*
@@ -114,11 +113,10 @@ char *secure_getenv(const char *name);
 static SLJIT_INLINE int create_tempfile(void)
 {
 	int fd;
-
 	char tmp_name[256];
-	size_t tmp_name_len;
+	size_t tmp_name_len = 0;
 	char *dir;
-	size_t len;
+	struct stat st;
 #if defined(SLJIT_SINGLE_THREADED) && SLJIT_SINGLE_THREADED
 	mode_t mode;
 #endif
@@ -132,49 +130,44 @@ static SLJIT_INLINE int create_tempfile(void)
 	}
 #endif
 
-#ifdef P_tmpdir
-	len = (P_tmpdir != NULL) ? strlen(P_tmpdir) : 0;
-
-	if (len > 0 && len < sizeof(tmp_name)) {
-		strcpy(tmp_name, P_tmpdir);
-		tmp_name_len = len;
-	}
-	else {
-		strcpy(tmp_name, "/tmp");
-		tmp_name_len = 4;
-	}
-#else
-	strcpy(tmp_name, "/tmp");
-	tmp_name_len = 4;
-#endif
-
 	dir = secure_getenv("TMPDIR");
 
 	if (dir) {
-		len = strlen(dir);
-		if (len > 0 && len < sizeof(tmp_name)) {
-			strcpy(tmp_name, dir);
-			tmp_name_len = len;
+		tmp_name_len = strlen(dir);
+		if (tmp_name_len > 0 && tmp_name_len < sizeof(tmp_name)) {
+			if ((stat(dir, &st) == 0) && S_ISDIR(st.st_mode))
+				strcpy(tmp_name, dir);
 		}
+	}
+
+#ifdef P_tmpdir
+	if (!tmp_name_len) {
+		tmp_name_len = strlen(P_tmpdir);
+		if (tmp_name_len > 0 && tmp_name_len < sizeof(tmp_name))
+			strcpy(tmp_name, P_tmpdir);
+	}
+#endif
+	if (!tmp_name_len) {
+		strcpy(tmp_name, "/tmp");
+		tmp_name_len = 4;
 	}
 
 	SLJIT_ASSERT(tmp_name_len > 0 && tmp_name_len < sizeof(tmp_name));
 
-	while (tmp_name_len > 0 && tmp_name[tmp_name_len - 1] == '/') {
-		tmp_name_len--;
-		tmp_name[tmp_name_len] = '\0';
-	}
+	if (tmp_name[tmp_name_len - 1] == '/')
+		tmp_name[--tmp_name_len] = '\0';
 
-#ifdef O_TMPFILE
-	fd = open(tmp_name, O_TMPFILE | O_EXCL | O_RDWR | O_NOATIME | O_CLOEXEC, 0);
+	/*
+	 * the previous trimming might had left an empty string if TMPDIR="/"
+	 * so work around the problem below as suggested by @zherczeg
+	 */
+	fd = open(tmp_name_len ? tmp_name : "/",
+		O_TMPFILE | O_EXCL | O_RDWR | O_NOATIME | O_CLOEXEC, 0);
 	if (fd != -1)
 		return fd;
-#endif
 
 	if (tmp_name_len + 7 >= sizeof(tmp_name))
-	{
 		return -1;
-	}
 
 	strcpy(tmp_name + tmp_name_len, "/XXXXXX");
 #if defined(SLJIT_SINGLE_THREADED) && SLJIT_SINGLE_THREADED


sljit_src/sljitProtExecAllocator.c
```diff
@@ -172,11 +182,11 @@ static SLJIT_INLINE int create_tempfile(void)
...
 
 	if (tmp_name_len + 7 >= sizeof(tmp_name))
-	{
 		return -1;
-	}
 
-	strcpy(tmp_name + tmp_name_len, "/XXXXXX");
+	if (!(tmp_name_len == 1 && *tmp_name == '/'))
+		tmp_name[tmp_name_len++] = '/';
+	strcpy(tmp_name + tmp_name_len, "XXXXXX");
```
zherczeg, line 189: I think `//XXXXXX` is valid, so we might not need to change this.

sljit_src/sljitProtExecAllocator.c
```diff
@@ -83,20 +83,16 @@ struct chunk_header {
        as it only uses local variables
 */
 
+#if !(defined(__NetBSD__) && defined(MAP_REMAPDUP))
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <stdio.h>
```
zherczeg, line 89: do we need stdio?

sljit_src/sljitProtExecAllocator.c
```diff
@@ -153,17 +149,31 @@ static SLJIT_INLINE int create_tempfile(void)
...
 		if (len > 0 && len < sizeof(tmp_name)) {
-			strcpy(tmp_name, dir);
+			if ((stat(dir, &st) == 0) && S_ISDIR(st.st_mode)) {
+				strcpy(tmp_name, dir);
+				tmp_name_len = len;
+			}
+		}
+	}
+
+#ifdef P_tmpdir
```
zherczeg, line 159: there is a `#ifdef P_tmpdir` above which does the same thing, does it?

sljit_src/sljitProtExecAllocator.c
carenas: without this include then P_tmpdir wouldn't be defined.
ironically, most of the time when build for testing; stdio gets included already and it will seem to work, until the code gets deployed in production.

sljit_src/sljitProtExecAllocator.c
carenas: still the new code avoids the unnecessary double "//" which might be problematic if this code ever gets ported to Novell or any other system where that makes a difference:

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_13

sljit_src/sljitProtExecAllocator.c
```diff
@@ -132,38 +128,36 @@ static SLJIT_INLINE int create_tempfile(void)
...
-	tmp_name_len = 4;
-#endif
-
 	dir = secure_getenv("TMPDIR");
 
 	if (dir) {
 		len = strlen(dir);
 		if (len > 0 && len < sizeof(tmp_name)) {
-			strcpy(tmp_name, dir);
+			if ((stat(dir, &st) == 0) && S_ISDIR(st.st_mode)) {
```
zherczeg, line 136: Is this new code disallows passing an invalid directory to silently disabling JIT compiling?

zherczeg:
There is something which came to my mind. If tmp directory points to a network file system, is it possible to read/write the machine code from the "other side"? If yes, allowing to change the tmp directory seems quite dangerous.

sljit_src/sljitProtExecAllocator.c
carenas: didn't consider that a feature, but yes if TMPDIR points to an invalid directory then we will fallback to use the other 2 options instead of failing.

I was considering this instead to be a reliability fix, so that something as important as the memory needed for JIT will be most likely to succeed than before.

IMHO, if we want to provide some sort of external flag to disable JIT we should define either a specific ENV variable or some other flag than a broken TMPDIR

sljit_src/sljitProtExecAllocator.c
carenas: this code is to revert the trimming of "/" that was done from the directory to use in lines 159-160 and that was technically not needed, since a directory with trailing "/" seems to work just fine for that open call (tested in Debian 10 with ext4 and tmpfs) as shown by:

 ```
#define _GNU_SOURCE
#include <stdio.h>
#include <fcntl.h>
#include <unistd.h>
#include <stdlib.h>

int main(int argc, char *argv[])
{
  int fd = open(argv[1], O_TMPFILE|O_EXCL|O_RDWR|O_NOATIME|O_CLOEXEC, 0);
  if (fd < 1) {
    perror("temp file creation failed");
    exit(1);
  }

  close(fd);
  return 0;
}
```

note also that in modern Linux, that open call should succeed MOST (the only exception I can't think of are misconfigured docker containers) of the time since support for O_TMPFILE is wide spread and is definitely preferred to falling back to mkostemp, which has still a race condition in threaded mode.

either way, I think that trimming the trailing "/" and using only one "/" are cleaner so would rather have the code as proposed

**Comment by carenas** [comment]
most of the scenarios where that would be possible were prevented with 02a0762fcac0bdc7fe4e1b728200a8d5fb8d2faa and in some way also some of the previous changes I proposed.

in most cases (specially in SE systems), TMPDIR would point instead to a private directory instead of using a publicly readable /tmp or /var/tmp (depending on the system) and so IMHO using TMPDIR is actually an improvement.

if you would rather fail if the file system is not safe (should only affect mkostemp) then we could add a check with either a whitelist or a blacklist (probably preferred) but it wouldn't be perfect either way

File: sljit_src/sljitProtExecAllocator.c (hunk 1 above)
Comment (zherczeg) [reply sljit_src/sljitProtExecAllocator.c:173]: Maybe I was not clear.

The following code is correct:
```
if (tmp_name_len > 0 && tmp_name[tmp_name_len - 1] == '/')
		tmp_name[--tmp_name_len] = '\0';
strcpy(tmp_name + tmp_name_len, "/XXXXXX");
```

You added the extra stuff, because there is an `#ifdef` part which needs it. However this code might be not exist so we do extra work for nothing.

If we use the code inside the ifdef:
`fd = open(tmp_name_len > 0 ? tmp_name : "/", ...`

We get the best from both worlds.

Plus `if (tmp_name_len + 7 >= sizeof(tmp_name))` does not need to be updated as well.

Comment (carenas) [reply sljit_src/sljitProtExecAllocator.c:173]: the first code has a bug, but as you correctly said it is mostly relevant outside the ifdef for O_TMPFILE, since it will break that call by passing "" as the directory.  you are correct that your proposed change avoids that but IMHO just works around the bug by making an even more unexpected (and undocumented) implementation.

the whole point of "refactoring" is to change the code, which is why I have to admit I am not sure if we should keep known bugs just to reduce the number of changed lines as an objective.

notice also that TMPDIR="/" is an edge case, and as I mentioned earlier the whole trimming of trailing "/" wasn't even needed (at least on my test system) but just because it works shouldn't be IMHO an incentive to not do the right thing.

finally; all this code could go away soon, because Linux could start using memfd_create (except for really old versions, and just because their libc didn't have the supported call, considering kernel support is ancient), and NetBSD >= 8 is already in a better/cleaner code path.

Comment (zherczeg) [reply sljit_src/sljitProtExecAllocator.c:173]: You have correctly identified the bug, nice job! But there is a simpler (and I think nicer) solution to fix it.

File: sljit_src/sljitProtExecAllocator.c [hunk 5]
```diff
@@ -83,22 +83,21 @@ struct chunk_header {
...
 #ifndef O_NOATIME
 #define O_NOATIME 0
 #endif
 
-#ifdef __O_TMPFILE
+/* this is a linux extension available since kernel 3.11 */
 #ifndef O_TMPFILE
-#define O_TMPFILE	(__O_TMPFILE | O_DIRECTORY)
-#endif
+#define O_TMPFILE 020200000
```
zherczeg, line 98: This is something new. Basically we try to pass this flag, and see what happens? I am ok with it, just want to make sure I understand correctly. These flags are harmless on other systems, don't they?

carenas: that is valid only in Linux. and wouldn't be needed if the code was compiled with -D_GNU_SOURCE as it will come included from fcntl.h

in NetBSD will just return an error because it is a directory and continue with the fallback

zherczeg:
Do you have any other plans, or can I do an update in pcre2?

carenas:
wasn't expecting this latest code to go into 10.35 per your previous comments, and eventhough it would only really affect NetBSD 7.2 (soon to be EOL) it might be worth #ifdef that open call anyway for correctness (see pull #41)

the only thing left around this code was the issue you raised about some sort of paranoid mode where additional checks are made to make sure that the TMPFILE used is safe and impossible to be leaked, but under the current implementation that would be only affecting NetBSD 7.2 in multithreaded mode, which shouldn't be using this code to begin with and that could sidestep the issue with a properly configured private TMPDIR, or a very broken linux setup (probably not that uncommon, considering the prevalence of containers with linux) which could do also a similar workaround even if MEMFD_CREATE is not enabled with the 10.35 release of PCRE

carenas:
on a related note, had noticed also a bug in PCRE2 where the threading support wasn't not properly done and that resulted in PCRE2 always using the non threaded SLJIT so it might be problematic to merge this AS-IS as it would make pcre itself not thread safe when used by a threaded application; and IMHO that is probably not worth looking for during the 10.35 RC period

sljit_src/sljitProtExecAllocator.c
carenas: is this a feature (runtime disabling of JIT) worth considering?, I can see some uses of it (ex: in git where a JIT enabled pcre being used by `git grep` might be problematic, but that is just because git (at least for PCRE2) uses the JIT fastpath from PCRE which doesn't provide the normal fallback.

are there any other users of sljit where runtime disabling of JIT might be needed?, this and the paranoid mode you suggested might be even somehow related at least on use cases

zherczeg:
I may misunderstand your comment, but SLJIT_SINGLE_THREADED is never defined in PCRE, so it always use the thread safe code.

carenas:
my bad, I left a important part out of that by not mentioning static builds (corrected), which is what cmake does by default (configure does both, and therefore prefers dynamic when linking).

note also the problem seems to be platform specific, as NetBSD doesn't show any pthread references in their binaries even when using configure and a dynamic object, while macOS shows them in both.

hopefully it is just user error, but wouldn't want risking to introduce a problem with PCRE this late in the release cycle which is why I wanted to raise the flag earlier (and while I still have no definitive reason, or otherwise I would had filed a bug for pcre already), will keep you updated as I troubleshoot otherwise.

carenas:
OK, got around looking at it again; definitely user error, and therefore you are encouraged to merge.

That means I also need a feature to force pcre2 to build the static version of sljit to avoid the known race condition for the tmpfile in NetBSD 7.2, or get working on that paranoid mode you asked about.

BTW there is a bug in MAP_REMAPDUP[1] which I hadn't yet got around fixing and that would affect otherwise 10.35, so should merge at least 83f4525 which is a shame, because I was really hoping the fork() bug would be gone at least for NetBSD when using this experimental allocator.

it is still rare to hit; after all NetBSD has been using a python with libffi with the same code since NetBSD 8.0 but since PCRE gets embedded in more places and some of those processes are more likely to stick around (ex: php inside a preforked apache) it might be also more likely to affect us.

[1] https://mail-index.netbsd.org/netbsd-bugs/2020/04/15/msg067149.html

PR #38, opened 2020-04-12
protexec: restrict file system permissions in the map files

carenas:
the backing file contents should not be accessible outside the running
process so make sure there are no RW permissions in the file created
for any user as early as possible.

### DIFF
--- a/sljit_src/sljitProtExecAllocator.c
+++ b/sljit_src/sljitProtExecAllocator.c
@@ -83,6 +83,7 @@ struct chunk_header {
        as it only uses local variables
 */
 
+#include <sys/stat.h>
 #include <fcntl.h>
 
 #ifndef O_NOATIME
@@ -118,12 +119,17 @@ static SLJIT_INLINE int create_tempfile(void)
 	size_t tmp_name_len;
 	char *dir;
 	size_t len;
+#if defined(SLJIT_SINGLE_THREADED) && SLJIT_SINGLE_THREADED
+	mode_t mode;
+#endif
 
 #ifdef HAVE_MEMFD_CREATE
 	/* this is a GNU extension, make sure to use -D_GNU_SOURCE */
 	fd = memfd_create("sljit", MFD_CLOEXEC);
-	if (fd != -1)
+	if (fd != -1) {
+		fchmod(fd, 0);
 		return fd;
+	}
 #endif
 
 #ifdef P_tmpdir
@@ -160,7 +166,7 @@ static SLJIT_INLINE int create_tempfile(void)
 	}
 
 #ifdef O_TMPFILE
-	fd = open(tmp_name, O_TMPFILE | O_EXCL | O_RDWR | O_NOATIME | O_CLOEXEC, S_IRUSR | S_IWUSR);
+	fd = open(tmp_name, O_TMPFILE | O_EXCL | O_RDWR | O_NOATIME | O_CLOEXEC, 0);
 	if (fd != -1)
 		return fd;
 #endif
@@ -171,10 +177,18 @@ static SLJIT_INLINE int create_tempfile(void)
 	}
 
 	strcpy(tmp_name + tmp_name_len, "/XXXXXX");
+#if defined(SLJIT_SINGLE_THREADED) && SLJIT_SINGLE_THREADED
+	mode = umask(0777);
+#endif
 	fd = mkostemp(tmp_name, O_CLOEXEC | O_NOATIME);
+#if defined(SLJIT_SINGLE_THREADED) && SLJIT_SINGLE_THREADED
+	umask(mode);
+#else
+	fchmod(fd, 0);
+#endif
 
 	if (fd == -1)
-		return fd;
+		return -1;
 
 	if (unlink(tmp_name)) {
 		close(fd);


sljit_src/sljitProtExecAllocator.c
```diff
@@ -145,7 +146,7 @@ static SLJIT_INLINE int create_tempfile(void)
 	}
 
 #ifdef O_TMPFILE
```
zherczeg, line 148: I have been thinking whether we should make this flag mandatory? Or drop this part?

carenas: O_TMPFILE is Linux specific, so if mandatory will break NetBSD
as far as memfd_create is not mandatory then using this is a better option than going all the way to mkostemp so discarding it wouldn't be recommended either IMHO

sljit_src/sljitProtExecAllocator.c
```diff
@@ -83,6 +83,7 @@ struct chunk_header {
        as it only uses local variables
 */
 
+#include <sys/stat.h>
```
zherczeg, line 86: Is this available in all posix OSes?

sljit_src/sljitProtExecAllocator.c
```diff
@@ -118,12 +119,17 @@ static SLJIT_INLINE int create_tempfile(void)
...
 
 #ifdef HAVE_MEMFD_CREATE
 	/* this is a GNU extension, make sure to use -D_GNU_SOURCE */
 	fd = memfd_create("sljit", MFD_CLOEXEC);
-	if (fd != -1)
-		return fd;
+	if (fd != -1) {
+		if (fchmod(fd, 0) == 0)
+			return fd;
+		else
```
zherczeg, line 130: No need else here.

sljit_src/sljitProtExecAllocator.c
```diff
@@ -160,7 +166,8 @@ static SLJIT_INLINE int create_tempfile(void)
 	}
 
 #ifdef O_TMPFILE
-	fd = open(tmp_name, O_TMPFILE | O_EXCL | O_RDWR | O_NOATIME | O_CLOEXEC, S_IRUSR | S_IWUSR);
+	fd = open(tmp_name,
+		O_TMPFILE | O_EXCL | O_RDWR | O_NOATIME | O_CLOEXEC, 0);
```
zherczeg, line 170: No need to break this into two lines.

sljit_src/sljitProtExecAllocator.c
```diff
@@ -171,10 +178,12 @@ static SLJIT_INLINE int create_tempfile(void)
 	}
 
 	strcpy(tmp_name + tmp_name_len, "/XXXXXX");
+	mode = umask(0777);
```
zherczeg, line 181: Is this thread safe?

sljit_src/sljitProtExecAllocator.c
```diff
@@ -118,12 +119,17 @@ static SLJIT_INLINE int create_tempfile(void)
...
 	size_t len;
+	mode_t mode;
 
 #ifdef HAVE_MEMFD_CREATE
 	/* this is a GNU extension, make sure to use -D_GNU_SOURCE */
 	fd = memfd_create("sljit", MFD_CLOEXEC);
-	if (fd != -1)
-		return fd;
+	if (fd != -1) {
+		if (fchmod(fd, 0) == 0)
```
zherczeg, line 128: It looks like this function is not available everywhere. Is this a new hidden requirement?

sljit_src/sljitProtExecAllocator.c
carenas: no; will better wrap within SLJIT_SINGLE_THREADED and leave the original version that used fchmod afterwards for threaded versions of the code.

note that this call is (under the sljit context) guarded by the allocator lock even in those cases.

sljit_src/sljitProtExecAllocator.c
carenas: OK, but I was expecting that in the extremely unlikely case when fchmod could fail here it was still probably better to fallback to use a different way to get the tmpfile than continue using one file that might be possible to hijack externally IMHO

sljit_src/sljitProtExecAllocator.c
carenas: it is part of POSIX and definitely more broadly available than memfd_create, and so available in Linux and NetBSD which AFAIK where the only ones that might be using this code.

eitherway will wrap around, specially since it will be also used below instead of umask for the threaded version.

sadly this makes this code less safe (IMHO) than the regular allocator which is kind of ironic for it being used in SE systems and might be worth documenting with some comments at the top (alongside all other know deficiencies)

sljit_src/sljitProtExecAllocator.c
carenas: yes, eventhough I was only considering Linux and NetBSD for this codebase
https://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/stat.h.html

sljit_src/sljitProtExecAllocator.c
zherczeg: I always considered this allocator as a workaround, rather than a better option. I had discussions about linux people before, and a key feature of SELinux (and friends) is NOT allowing jit at all, because runtime code generation is considered extremely dangerous. With this allocator you just cheat the system, exploit one of its weakness, and creates an even bigger danger than the normal allocator. So the name of the allocator is kind of misleading :)

zherczeg: Btw lets make this requirement mandatory. We don't need to support all existing systems with this allocator.

carenas: IMHO the fact that this is being broadly used regardless probably hints that it should had been remained more hidden and its deficiencies spelled it more prominently

PR #39, opened 2020-04-12
netbsd: secure_getenv compat

carenas:
avoid calling getenv in scenarios where it is not safe (like when
running setuid/setgid)

this should only be effective in NetBSD < 8, as anything that has
MAP_REMAPDUP is using a different implementation of the alloc_chunk
function that doesn't have this deficiency.

### DIFF
--- a/sljit_src/sljitProtExecAllocator.c
+++ b/sljit_src/sljitProtExecAllocator.c
@@ -97,7 +97,18 @@ struct chunk_header {
 
 #if !(defined(__NetBSD__) && defined(MAP_REMAPDUP))
 int mkostemp(char *template, int flags);
+
+#ifdef __NetBSD__
+/*
+ * this is a workaround for NetBSD < 8 that lacks a system provided
+ * secure_getenv function.
+ * ideally this should never be used, as the standard allocator is
+ * a preferred option for those systems and should be used instead.
+ */
+#define secure_getenv(name) issetugid() ?  NULL : getenv(name)
+#else
 char *secure_getenv(const char *name);
+#endif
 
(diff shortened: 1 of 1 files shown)


sljit_src/sljitProtExecAllocator.c
```diff
@@ -97,7 +97,24 @@ struct chunk_header {
...
+ * secure_getenv function.
+ * ideally this should never be used, as the standard allocator is
+ * a preferred option for those systems and should be used instead.
+ */
+static SLJIT_INLINE char *secure_getenv(const char *name)
+{
+	int secure_execution_required = getuid() != geteuid()
+					|| getgid() != getegid();
+
+	return secure_execution_required ? NULL : getenv(name);
```
zherczeg, line 113: This is ok, but what about making it a define?
```
#define secure_getenv(name) \
  (cond) ? NULL : getenv(name);
```

krytarowski:
If this is for NetBSD<8 then is it worth the patch? NetBSD-7 will go EOL soon.

zherczeg:
I am ok with the patch, but I don't know how netbsd maintenace works. Most OSes only do security updates after a release, so they never use the latest code for an old revision. Users can do anything though.

carenas:
the original code introduces a security issue that this fixes, tragically it is included in pkg_src so as such affects currently all NetBSD versions.

the comment about NetBSD < 8 not being affected also requires that the sljitProtExecAllocator be patched with 9a3d3998dcaf151f5c562f8dde457790fe0d167c, that was merged first but hasn't been released yet

PR #40, opened 2020-04-14
netbsd: cleanup alloc_chunk() in sljitProtExec for NetBSD >= 8

carenas:
in line with the current semantics make the map SHARED so it could
be used (with the same limitations and risks) like the old function
after fork()

Fixes: 9a3d3998dcaf151f5c562f8dde457790fe0d167c

### DIFF
--- a/sljit_src/sljitProtExecAllocator.c
+++ b/sljit_src/sljitProtExecAllocator.c
@@ -220,28 +220,26 @@ static SLJIT_INLINE struct chunk_header* alloc_chunk(sljit_uw size)
 static SLJIT_INLINE struct chunk_header* alloc_chunk(sljit_uw size)
 {
 	struct chunk_header *retval;
-	void *maprx;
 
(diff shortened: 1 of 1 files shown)


sljit_src/sljitProtExecAllocator.c
```diff
@@ -220,28 +220,26 @@ static SLJIT_INLINE struct chunk_header* alloc_chunk(sljit_uw size)
 static SLJIT_INLINE struct chunk_header* alloc_chunk(sljit_uw size)
 {
 	struct chunk_header *retval;
-	void *maprx;
 
 	retval = (struct chunk_header *)mmap(NULL, size,
-			PROT_MPROTECT(PROT_EXEC|PROT_WRITE|PROT_READ),
-			MAP_ANON, -1, 0);
+			PROT_READ|PROT_WRITE|PROT_MPROTECT(PROT_EXEC),
```
zherczeg, line 225: Please add spaces around binary operators.

PR #46, opened 2020-07-04
Adapt to NetBSD >= 8 only

krytarowski:
NetBSD-7 is EOL as of June 29th 2020:

http://mail-index.netbsd.org/netbsd-announce/2020/06/29/msg000316.html

While there, change to MAP_PRIVATE from MAP_SHARED.


zherczeg:
@carenas could you review this patch?

krytarowski:
Ah, I see. Well, I will reject this patch then.

PR #47, opened 2020-07-04
Add NetBSD PaX MPROTECT support in sljitExecAllocator.c

krytarowski:
However this.. crashes:

```
(gdb) c
Continuing.
Run test1

Program received signal SIGSEGV, Segmentation fault.
0x00007c591e56c020 in ?? ()
(gdb) bt
#0  0x00007c591e56c020 in ?? ()
#1  0x000000000040176c in test1 () at test_src/sljitTest.c:173
#2  0x00000000004225fc in sljit_test (argc=2, argv=0x7f7fff719608) at test_src/sljitTest.c:6602
#3  0x00000000004011cc in main (argc=2, argv=0x7f7fff719608) at test_src/sljitMain.c:83
```

I need some help to fix this.

### DIFF
--- a/sljit_src/sljitExecAllocator.c
+++ b/sljit_src/sljitExecAllocator.c
@@ -92,6 +92,64 @@ static SLJIT_INLINE void free_chunk(void *chunk, sljit_uw size)
 	VirtualFree(chunk, 0, MEM_RELEASE);
 }
 
+#elif defined(__NetBSD__)
+#include <stddef.h> /* max_align_t */
+
+#define ptr_overhead (sizeof(max_align_t))
+
+#define PTR_ADD(p, d) ((void *)((uintptr_t)(p) + (d)))
+
+static sljit_uw page_size;
+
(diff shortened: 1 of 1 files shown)


krytarowski:
OK. I got it. PaX MPROTECT supports only two mappings RW and RX, they can point to the same physical memory, but there is no support for concurrent RWX.

I don't know how to fix this? Or just assume that this is not compatible and move on? Are there any shortages with this allocator without `SLJIT_PROT_EXECUTABLE_ALLOCATOR`?

krytarowski:
@zherczeg do you think is it possible to use `mprotect(2)` to toggle RW -> RX in sljitExecAllocator.c? We have got a problem with RWX mapping at once. But there is a shopped chunk into blocks and it's not all clear how to achieve this.

zherczeg:
There is a protected allocator, which maps the region twice with different attributes.

krytarowski:
I know. But having a single region pointer and mprotect protection bits would be nice too.

Issue #51, opened 2020-07-10
macOS 11 support


carenas:
as suggested in :

  https://developer.apple.com/documentation/apple_silicon/porting_just-in-time_compilers_to_apple_silicon

where the arm64 version of that OS will enforce some of the constrains included with the Hardened Runtime (which is partially why the then undocumented MAP_JIT was added starting in macOS 10.14) and therefore result in a crash under the current implementation.

Suggested-by: @stuartcarnie

PR #52, opened 2020-07-10
macos: add support for W^X pages in Big Sur

carenas:
incomplete (missing any fallbacks or compatibility) and untested
but useful as a POC.

Fixes: #51

### DIFF
--- a/sljit_src/sljitNativeARM_64.c
+++ b/sljit_src/sljitNativeARM_64.c
@@ -255,6 +255,7 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 
 	code = (sljit_ins*)SLJIT_MALLOC_EXEC(compiler->size * sizeof(sljit_ins));
 	PTR_FAIL_WITH_EXEC_IF(code);
+	pthread_jit_write_protect_np(0);
 	buf = compiler->buf;
 
 	code_ptr = code;
@@ -379,6 +380,7 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 	code = (sljit_ins *)SLJIT_ADD_EXEC_OFFSET(code, executable_offset);
 	code_ptr = (sljit_ins *)SLJIT_ADD_EXEC_OFFSET(code_ptr, executable_offset);
 
+	pthread_jit_write_protect_np(1);
 	SLJIT_CACHE_FLUSH(code, code_ptr);
 	return code;
 }


stuartcarnie:
I have experimented with this and it appears to work, however, we have to add calls to `pthread_jit_write_protect_np` anywhere we are writing to these executable pages. An example is in `sljit_malloc_exec`:

https://github.com/zherczeg/sljit/blob/07562a3c589ccffb2c8e286e2df3d0694be8861a/sljit_src/sljitExecAllocator.c#L296

* `test_exec_allocator` fails, as it calls `memset` on the pages, which are now `RX`.
* `test1` .. `test10` work
* `test11` fails, as it calls `sljit_set_const` to update the memory, which is sill `RX`.

I could keep sprinkling calls to `pthread_jit_write_protect_np` to make more tests pass, but that seems like the wrong approach. I would imagine a new API is in order to switch from write to execute mode, allowing calls to APIs like `sljit_set_const` to be executed as a batch.

stuartcarnie:
@zherczeg / @carenas here is an initial idea / pass at JIT support for Si 

https://github.com/stuartcarnie/sljit/commit/783ff6baaa3f4231b64e7b788301f9fe8819d4ad

The benefit of this approach is that the thread is in EXEC mode at rest and only briefly changes to WRITE mode when mutating the JITted code or recycling memory pages.

This is liberally stolen from mono, which use two macros that assert the desired access permissions (WRITE or EXEC) within the scope of a function. We declare WRITE behavior in the alloc and generate methods, for example. These macros do nothing for other platforms. 

https://github.com/Unity-Technologies/mono/pull/1309/commits/3f2278e081484e6616ee5892895f999610453ca9#

I've run the `sljit_test` and `regex_test` binaries on DTK hardware to verify it works. There may need to be an audit to ensure the macros are performing the assertions in all places.

Also, I included my CMakeLists.txt file for reference only.

zherczeg:
@stuartcarnie thank you for the patch. I think we don't need any automatic switching between modes. You should enable writing before `SLJIT_MALLOC_EXEC` in `sljit_generate_code` and disable it before `SLJIT_CACHE_FLUSH`. You need to do a similar thing either in `sljit_set_jump_addr` and `sljit_set_const` or in `modify_imm64_const`. Since the whole thing seems arm64 and apple specific, I would simply add a macro only to this file, define it at the beginning and undef it at the end.

carenas:
@zherczeg: it might be also a good time to decouple the chunk_header from the executable memory, so management of the chunks can be done without having to change the permission of the executable memory pages.

while the change is mandatory for macos 11 arm64, it also needs to be done for x86; or otherwise it will also fail the same way when linked against a SE binary/system

carenas:
likely superseded by a combination of #53, #60 and #61

PR #53, opened 2020-07-16
JIT support for arm64

stuartcarnie:
This PR introduces JIT support for Si. It makes the following changes:

* Adds macros for `__APPLE__` && `__arm64__`, shamelessly lifted from https://github.com/Unity-Technologies/mono/commit/3f2278e081484e6616ee5892895f999610453ca9
    * These macros are no-ops for other platforms
    * These macros utilize `__attribute__(clean(<function>))` to restore the previous state
* Enables write scope in specific functions that must write to this memory, namely:
    * `sljit_malloc_exec`
    * `sljit_free_exec` and
    * `modify_imm64_const`

Update the `test_exec_allocator` tests to use the `SLJIT_SCOPE_ENABLE_JIT_WRITE` macro to enable write access, as these tests write to the allocated memory returned from the allocator.

### DIFF
--- a/sljit_src/sljitExecAllocator.c
+++ b/sljit_src/sljitExecAllocator.c
@@ -192,6 +192,63 @@ static SLJIT_INLINE void free_chunk(void *chunk, sljit_uw size)
 
 #endif
 
+#if defined(__APPLE__) && defined(__arm64__)
+__thread jit_protect_mode arm_current_jit_protect_mode = JPM_ENABLED;
+
+inline void sljit_arm_jit_write_protect_enable()
+{
+	if (__builtin_available(macOS 11, *)) {
+		if (arm_current_jit_protect_mode != JPM_ENABLED) {
+			pthread_jit_write_protect_np(1);
+			arm_current_jit_protect_mode = JPM_ENABLED;
+		}
+	}
+}
+
+inline void sljit_arm_jit_write_protect_disable()
+{
+	if (__builtin_available(macOS 11, *)) {
+		if (arm_current_jit_protect_mode != JPM_DISABLED) {
+			pthread_jit_write_protect_np(0);
+			arm_current_jit_protect_mode = JPM_DISABLED;
+		}
+	}
+}
+
+#define SLJIT_SCOPE_ENABLE_JIT_WRITE() \
+	__attribute__((unused, cleanup(sljit_arm_restore_jit_protect_mode))) \
+	jit_protect_mode scope_restrict_mode = arm_current_jit_protect_mode; \
+		sljit_arm_jit_write_protect_disable();
+
+#define SLJIT_SCOPE_ENABLE_JIT_EXEC() \
+	__attribute__((unused, cleanup(sljit_arm_restore_jit_protect_mode))) \
+	jit_protect_mode scope_restrict_mode = arm_current_jit_protect_mode; \
+		sljit_arm_jit_write_protect_enable();
+
+inline void sljit_arm_restore_jit_protect_mode(jit_protect_mode* previous_jit_protect_mode)
+{
+	if (*previous_jit_protect_mode == arm_current_jit_protect_mode)
+		return;
+
+	switch (*previous_jit_protect_mode)
+	{
+	case JPM_ENABLED:
+		sljit_arm_jit_write_protect_enable();
+		break;
+	case JPM_DISABLED:
+	case JPM_NONE:
+	default:
+		sljit_arm_jit_write_protect_disable();
+	}
+}
+
+#else
+#define SLJIT_SCOPE_ENABLE_JIT_WRITE()
+#define SLJIT_SCOPE_ENABLE_JIT_EXEC()
+#endif /* __APPLE__ && __arm64__ */
+
+
+
 /* --------------------------------------------------------------------- */
 /*  Common functions                                                     */
 /* --------------------------------------------------------------------- */
@@ -254,6 +311,8 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_malloc_exec(sljit_uw size)
 	sljit_uw chunk_size;
 
 	allocator_grab_lock();
+	SLJIT_SCOPE_ENABLE_JIT_WRITE();
+
 	if (size < (64 - sizeof(struct block_header)))
 		size = (64 - sizeof(struct block_header));
 	size = ALIGN_SIZE(size);
@@ -323,6 +382,8 @@ SLJIT_API_FUNC_ATTRIBUTE void sljit_free_exec(void* ptr)
 	struct free_block* free_block;
 
 	allocator_grab_lock();
+	SLJIT_SCOPE_ENABLE_JIT_WRITE();
+
 	header = AS_BLOCK_HEADER(ptr, -(sljit_sw)sizeof(struct block_header));
 	allocated_size -= header->size;
 
--- a/sljit_src/sljitNativeARM_64.c
+++ b/sljit_src/sljitNativeARM_64.c
@@ -153,6 +153,8 @@ static SLJIT_INLINE sljit_s32 emit_imm64_const(struct sljit_compiler *compiler, 
 
 static SLJIT_INLINE void modify_imm64_const(sljit_ins* inst, sljit_uw new_imm)
 {
+	SLJIT_SCOPE_ENABLE_JIT_WRITE();
+	
 	sljit_s32 dst = inst[0] & 0x1f;
 	SLJIT_ASSERT((inst[0] & 0xffe00000) == MOVZ && (inst[1] & 0xffe00000) == (MOVK | (1 << 21)));
 	inst[0] = MOVZ | dst | ((new_imm & 0xffff) << 5);
@@ -253,8 +255,10 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 	CHECK_PTR(check_sljit_generate_code(compiler));
 	reverse_buf(compiler);
 
+	SLJIT_SCOPE_ENABLE_JIT_WRITE();
 	code = (sljit_ins*)SLJIT_MALLOC_EXEC(compiler->size * sizeof(sljit_ins));
 	PTR_FAIL_WITH_EXEC_IF(code);
+
 	buf = compiler->buf;
 
 	code_ptr = code;
--- a/sljit_src/sljitLir.h
+++ b/sljit_src/sljitLir.h
@@ -1534,6 +1534,34 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_op_custom(struct sljit_compiler *c
 SLJIT_API_FUNC_ATTRIBUTE void sljit_set_current_flags(struct sljit_compiler *compiler,
 	sljit_s32 current_flags);
 
+#if defined(__APPLE__) && defined(__arm64__)
+typedef enum {
+    JPM_NONE,
+    JPM_ENABLED,
+    JPM_DISABLED,
+} jit_protect_mode;
+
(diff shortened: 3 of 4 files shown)


sljit_src/sljitExecAllocator.c
```diff
@@ -192,6 +192,63 @@ static SLJIT_INLINE void free_chunk(void *chunk, sljit_uw size)
...
+		sljit_arm_jit_write_protect_disable();
+
+#define SLJIT_SCOPE_ENABLE_JIT_EXEC() \
+	__attribute__((unused, cleanup(sljit_arm_restore_jit_protect_mode))) \
+	jit_protect_mode scope_restrict_mode = arm_current_jit_protect_mode; \
+		sljit_arm_jit_write_protect_enable();
+
+inline void sljit_arm_restore_jit_protect_mode(jit_protect_mode* previous_jit_protect_mode)
+{
+	if (*previous_jit_protect_mode == arm_current_jit_protect_mode)
```
zherczeg, line 230: I still don't get why we need to remember the previous mode. I don't know any nested compiling in sljit.

stuartcarnie: @zherczeg ahh right – this state is internal to sljit, so I'll remove it 👍

stuartcarnie: However, I would point out that `sljit_malloc_exec` requires the switch from `RX` to `RW` as it writes to those pages.

stuartcarnie: @zherczeg regarding:

> @stuartcarnie thank you for the patch. I think we don't need any automatic switching between modes. You should enable writing before `SLJIT_MALLOC_EXEC` in `sljit_generate_code` and disable it before `SLJIT_CACHE_FLUSH`. You need to do a similar thing either in `sljit_set_jump_addr` and `sljit_set_const` or in `modify_imm64_const`. Since the whole thing seems arm64 and apple specific, I would simply add a macro only to this file, define it at the beginning and undef it at the end.

Is missing some important areas:

1. `sljit_free_exec` also requires write access to the pages to add them to the free list
2. `SLJIT_MALLOC_EXEC` is used directly in the tests, meaning significant changes to those

For example, `test_exec_allocator` expects to write to memory:

https://github.com/zherczeg/sljit/blob/96d945e76e614a8bcc70d3c95acbc97e85086a89/test_src/sljitTest.c#L92-L98

Remembering the state ensures the `SLJIT_MALLOC_EXEC` does not remove the disabled write protection on exit for these tests. 

I would argue that given `SLJIT_MALLOC_EXEC` is the narrowest scope that requires RW access, it is most secure to keep `SLJIT_SCOPE_ENABLE_JIT_WRITE` macro in the `sljit_malloc_exec` function.

zherczeg: These are artifical tests, and they should not define requirements for the implementation. The easiest way to handle them is disabling them on Apple systems. A better approach would be to add the necessary macro calls which are simply nops on other systems.

stuartcarnie: @zherczeg ok, I can do that. Thanks for the feedback.

You can appreciate my concern that if I have to make a number of changes to files that look like tests, it indicates a change in behavior that may break other clients of this code. I naturally trust your confidence that these tests are not representative of a typical consumer of sljit

zherczeg: `SLJIT_MALLOC_EXEC` only used in `test_exec_allocator`. Probably disabling execution at the beginning of the test and enable it at the end should be enough (that test does not execute anything). If you need modification in any other test let me know.

sljit_src/sljitNativeARM_64.c
```diff
@@ -253,8 +255,10 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 	CHECK_PTR(check_sljit_generate_code(compiler));
 	reverse_buf(compiler);
 
+	SLJIT_SCOPE_ENABLE_JIT_WRITE();
 	code = (sljit_ins*)SLJIT_MALLOC_EXEC(compiler->size * sizeof(sljit_ins));
 	PTR_FAIL_WITH_EXEC_IF(code);
```
zherczeg, line 260: `PTR_FAIL_WITH_EXEC_IF` does early return. This might affect the enabling logic.

stuartcarnie: The macros use a `__attribute__(cleanup(<function>))` to ensure state is restore at all return points.

sljit_src/sljitExecAllocator.c
```diff
@@ -192,6 +192,63 @@ static SLJIT_INLINE void free_chunk(void *chunk, sljit_uw size)
 
 #endif
 
+#if defined(__APPLE__) && defined(__arm64__)
```
carenas, line 195: this is also needed in x86, to test will need to have the test application linked against the Hardened Runtime though.

once the '__arm64__' define is removed it might make more sense to get all this code together with the rest of the __APPLE__ specific code in line 97

sljit_src/sljitExecAllocator.c
```diff
@@ -192,6 +192,63 @@ static SLJIT_INLINE void free_chunk(void *chunk, sljit_uw size)
...
+__thread jit_protect_mode arm_current_jit_protect_mode = JPM_ENABLED;
```
carenas, line 196: this will likely break if not linking with pthread (should also `#define SLJIT_SINGLE_THREADED 1` in sljitConfig.h

sljit_src/sljitExecAllocator.c
```diff
@@ -192,6 +192,63 @@ static SLJIT_INLINE void free_chunk(void *chunk, sljit_uw size)
...
+
+inline void sljit_arm_jit_write_protect_enable()
+{
+	if (__builtin_available(macOS 11, *)) {
```
carenas, line 200: this is likely to be also needed for some versions of iOS (and friends), probably even older than macOS 11, since it would seem this "feature" was first implemented there and its hardware support might be even 2 generations old

sljit_src/sljitExecAllocator.c
```diff
@@ -192,6 +192,63 @@ static SLJIT_INLINE void free_chunk(void *chunk, sljit_uw size)
...
+		if (arm_current_jit_protect_mode != JPM_ENABLED) {
+			pthread_jit_write_protect_np(1);
+			arm_current_jit_protect_mode = JPM_ENABLED;
+		}
+	}
+}
+
+inline void sljit_arm_jit_write_protect_disable()
+{
+	if (__builtin_available(macOS 11, *)) {
```
carenas, line 210: this is most likely breaking when using gcc (or other compilers) instead of XCode/clang

sljit_src/sljitNativeARM_64.c
carenas: while the use of the cleanup attribute might had been useful where this code was lifted from, as shown in #52, there are only a few places (even less if the allocator is changed first, so it doesn't keep the chunk headers together with the code) where this need to be added and having it done explicitly makes the code simpler and easier to follow

sljit_src/sljitExecAllocator.c
stuartcarnie: JIT execution on iOS / iPadOS is explicitly forbidden. We could revisit this if that changes in the future

sljit_src/sljitExecAllocator.c
stuartcarnie: This code is guarded for macOS only and gcc is not even sure they will support compiling for Apple Silicon.

See gcc comment here for reference: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96168#c6

carenas: correct, but my point is that we should ideally set the expectations correctly and error out earlier if something other than XCode/clang is used to try to build this in macOS (ex: #error with a clear message) instead of hitting this line and throwing some sort of weird syntax error.

additionally; if we are giving up on using gcc a similar construct could be used to the code above that does MAP_JIT detection and which would make sense adding as a follow up commit.

sljit_src/sljitExecAllocator.c
zherczeg: I hope this patch will work without remembering the previous state, so this problem should go away.

zherczeg:
#60 introduces a new `SLJIT_UPDATE_WX_FLAGS` which could be useful for this patch.

sljit_src/sljitExecAllocator.c
carenas: make sense, there is still the open issue if an ifdef might be required to avoid building failures in iOS (and friends).  is building for iOS something that you could consider testing?

stuartcarnie:
I will rebase when you merge those PRs

carenas:
IMHO the only missing transition should be the one in ```sljit_generate_code()``` and that should be added somewhere after ```PTR_FAIL_WITH_EXEC_IF(code);``` and before the labels and jumps are written, so probably before:

https://github.com/zherczeg/sljit/blob/e670d24390a754ef14a869f43d8f0213e5cc01ef/sljit_src/sljitNativeARM_64.c#L270

```test1()``` will also need additional handling (because the memory is not really writeable when returned) and since you mentioned not wanting to disable the test, then will need to do once a whole RW <-> RX dance for the whole function.

to make sure nothing is broken will be a good idea to also test your changes work when the other allocators are enabled:

```
  $ make EXTRA_CPPFLAGS=-DSLJIT_WX_EXECUTABLE_ALLOCATOR
```

SLJIT_PROT_EXECUTABLE_ALLOCATOR will need code changes to workaround the lack of secure_getenv() but you already know how.

stuartcarnie:
Sounds good, @carenas. Will do the updates and run the tests when those other PRs merge

zherczeg:
If you have time, I think it is worth to resume this work since the necessary infrastructure improvements has been landed.

sljit_src/sljitExecAllocator.c
carenas: FWIW this syntax broke my build in macOS 10.13 (using Xcode) but the alternative used in b0d7c3581f3fbcac797070423d7b15ef237f0a3c should work if I guessed correctly the version number you should have in your headers.

carenas:
to avoid conflicts, it is better to base the changes on #67

the code in the development branch [apple](/carenas/sljit/tree/)up to 2d9eed6ec7d0bb05a6f2d068024bb5a1184aa3a3 should be enough to get this working (obviously untested in real software/hardware)

sljit_src/sljitExecAllocator.c
carenas: not sure how it behaves in real hardware but it is likely that every new page allocated will be non-writeable until the call to the apple API and regardless if it was called before the allocation.

carenas:
with #67 merged, it might be easier to test instead on #90; eitherway we should probably make sure something gets tested and released early enough to at least be included in a PCRE release before Apple Silicon or BigSur get released

zherczeg:
Is this still valid, or shall we close it?

stuartcarnie:
@zherczeg I can look at it again when I return my DTK and receive my Mac Mini

carenas:
@stuartcarnie: I got a Mac Mini with 11.2.2/arm64e and HEAD runs well on it after applying #105

if using PCRE 10.36 you will need #90 (which is also attached in the downstream bug) and should work fine otherwise at least until upgraded to 11.2 or newer where macOS apparently silently fixed a bug and make mmap/mprotect consistent (like it is in HardenedBSD) so a lying mmap can be detected.

carenas:
> shall we close it?

I think it would be better to close it.

I'd been using it for more than a year already, and when I heard someone complain it wasn't supported on HIS mac, I realized all this time people that use macports instead of brew are still waiting for this to close as shown in macports/macports-ports#14675

PR #54, opened 2020-07-19
Fix doc/tutorial examples and improve Makefile rules

nymacro:
This updates the doc/tutorial C files to be compatible with latest APIs.
The Makefile rules are also updated to only depend on dependent files, not the output directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,12 +20,14 @@ CFLAGS += -O2 -Wall
 REGEX_CFLAGS += $(CFLAGS) -fshort-wchar
 LDFLAGS = $(EXTRA_LDFLAGS)
 
-TARGET = sljit_test regex_test
-
 BINDIR = bin
 SRCDIR = sljit_src
 TESTDIR = test_src
 REGEXDIR = regex_src
+EXAMPLEDIR = doc/tutorial
+
+TARGET = $(BINDIR)/sljit_test $(BINDIR)/regex_test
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -37,31 +39,60 @@ SLJIT_LIR_FILES = $(SRCDIR)/sljitLir.c $(SRCDIR)/sljitExecAllocator.c $(SRCDIR)/
 	$(SRCDIR)/sljitNativeTILEGX_64.c \
 	$(SRCDIR)/sljitNativeX86_common.c $(SRCDIR)/sljitNativeX86_32.c $(SRCDIR)/sljitNativeX86_64.c
 
-all: $(BINDIR) $(TARGET)
+.PHONY: all examples
```
zherczeg, line 42: I would not create examples by default.

sljit_src/sljitLir.h
```diff
@@ -699,7 +699,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fast_enter(struct sljit_compiler *
 */
 
 /*
-   IMPORATNT NOTE: memory access MUST be naturally aligned except
+   IMPORTANT NOTE: memory access MUST be naturally aligned except
```
carenas, line 702: s/except/unless

Makefile
nymacro: `all` is still the default make rule and doesn't build `examples` by default.

The phony rule just tells make that these targets aren't file-based targets (i.e. the target doesn't create a file named `all`). Clean should also be added as a phony target.

PR #55, opened 2020-07-19
Haiku/Illumos build fix :

devnexen:
- Haiku does not have madvise call.
- Illumos depends on the flags passed whether access to the old API or the newer one without madivse availability.

### DIFF
--- a/sljit_src/sljitUtils.c
+++ b/sljit_src/sljitUtils.c
@@ -157,6 +157,10 @@ SLJIT_API_FUNC_ATTRIBUTE void SLJIT_FUNC sljit_release_lock(void)
 #include <sys/types.h>
 #include <sys/mman.h>
 
+#ifndef MADV_DOTNEED
+#define madvise posix_madvise
+#define MADV_DOTNEED POSIX_MADV_DOTNEED
+#endif
 #ifndef MAP_ANON
 #ifdef MAP_ANONYMOUS
 #define MAP_ANON MAP_ANONYMOUS


sljit_src/sljitUtils.c
```diff
@@ -157,6 +157,10 @@ SLJIT_API_FUNC_ATTRIBUTE void SLJIT_FUNC sljit_release_lock(void)
 #include <sys/types.h>
 #include <sys/mman.h>
 
+#ifndef MADV_DOTNEED
```
carenas, line 160: is this a typo of MADV_DONTNEED and so could be merged with the ifnded below?

carenas:
> * Illumos depends on the flags passed whether access to the old API or the newer one without madivse availability.

I see no problem building/running 96d945e76e614a8bcc70d3c95acbc97e85086a89 (the commit before this) in Illumos:

$ uname -srvm
SunOS 5.11 illumos-c559157643 i86pc

could you elaborate on what that problem was?, if anything I would think changing all calls to use posix_madvise explicitly might be better than the current define

devnexen:
For Illumos, I mention that depends on the CFLAGS passed (_XOPEN_SOURCE/_POSIX_C_SOURCE or not). Otherwise, we can always move to posix_madvise if you all prefer.

devnexen:
After fixing locally the typo, it is what happens on Illumos
image

carenas:
OK; so the "build problem" for Illumos was just a warning when the feature macros prevent madvise; as for the haiku issue guess it fails to link because there is no madvise (sorry, no haiku for me to test)?

devnexen:
Exactly; no madvise call for Haiku.

carenas:
a "fixed" alternative that just uses posix_madvise() for portability in #57

devnexen:
works for me.

carenas:
> works for me.

does this mean tested, this time?

devnexen:
yes.

PR #57, opened 2020-07-20
utils: avoid madvise portability problems

carenas:
an alternative fix for #55 that avoid typos and simplify logic


carenas:
tested in Linux (Ubuntu 20.04), macOS 10.13.6, FreeBSD 12.1, OpenBSD 6.7 and NetBSD 9
originally tested in illumos (AKA OpenSolaris) and Haiku by @devnexen, but confirmation that there are no regressions would be encouraged

PR #58, opened 2020-07-21
protexec: drop support for NetBSD < 8

carenas:
NetBSD-7 is EOL as of June 29th 2020

originally suggested by @krytarowski in #46

### DIFF
--- a/sljit_src/sljitProtExecAllocator.c
+++ b/sljit_src/sljitProtExecAllocator.c
@@ -83,7 +83,7 @@ struct chunk_header {
        as it only uses local variables
 */
 
-#if !(defined(__NetBSD__) && defined(MAP_REMAPDUP))
+#ifndef __NetBSD__
 #include <sys/stat.h>
 #include <fcntl.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


sljit_src/sljitProtExecAllocator.c
```diff
@@ -98,19 +98,8 @@ struct chunk_header {
...
- * ideally this should never be used, as the standard allocator is
- * a preferred option for those systems and should be used instead.
- */
-#define secure_getenv(name) issetugid() ?  NULL : getenv(name)
-#else
-char *secure_getenv(const char *name);
-#endif
-
 #ifndef _GNU_SOURCE
+char *secure_getenv(const char *name);
```
krytarowski, line 102: I would keep this line out of this `#ifndef _GNU_SOURCE` ifdef in order to prevent unintentional changing of the code for !NetBSD.

carenas: it is just a function definition for whoever forgets to build this on linux with EXTRA_CPPFLAGS='-D_GNU_SOURCE'

it doesn't change the code anywhere, except avoiding a warning and some silly bugs for whoever ignores those when it tries to convert the string into an int

krytarowski: I see, then this is OK. It's also not a NetBSD code block, so I'm relaxed with it myself.

zherczeg:
Please add a note here when the patch is ready.

carenas:
> Please add a note here when the patch is ready.

AFAIK patch is ready.  I'd been mostly just rebasing it, re-reviewing it and re-testing; the only comment was resolved without changes.

It will break building in NetBSD 7.x but that is expected.

zherczeg:
Ok, it was just unclear to me whether you have future plans.

PR #59, opened 2020-07-23
build: reintroduce bin directory dependency

carenas:
after the bin directory was removed as a dependency to cleanup (and fix)
the dependency tracking, the following scenario is possibly breaking
builds:

  $ git clean -fxd
  Removing bin/
  $ make
  cc  -DSLJIT_CONFIG_AUTO=1 -Isljit_src -O2 -Wall -c -o bin/sljitMain.o test_src/sljitMain.c
  error: unable to open output file 'bin/sljitMain.o': 'No such file or directory'
  1 error generated.
  make: *** [bin/sljitMain.o] Error 1

luckily I already fixed this problem in my "makefile" branch as
carenas/sljit@5618ec91b85df8427f38e36f850e1f828501aed4 so
implement it now by adding a sentinel to the bin directory and using
that instead as a dependency.

while at it, fix some other minor issues from that branch that are
hopefully something that we could agree on and might had even been
discussed recently.

CC: @nymacro
Fixes: fcf394a54711946ff938c084d8123e4af9940db2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,8 @@ EXAMPLE_TARGET = $(BINDIR)/func_call $(BINDIR)/first_program $(BINDIR)/branch $(
 
 SLJIT_HEADERS = $(SRCDIR)/sljitLir.h $(SRCDIR)/sljitConfig.h $(SRCDIR)/sljitConfigInternal.h
 
-SLJIT_LIR_FILES = $(SRCDIR)/sljitLir.c $(SRCDIR)/sljitExecAllocator.c $(SRCDIR)/sljitUtils.c \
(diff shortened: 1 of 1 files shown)


nymacro:
Thanks @carenas , I obviously didn't test the make changes well enough.

carenas:
on the contrary, my fault for not pointing it earlier (specially since I knew there was a booby trap, and the least I should have done was point you to my [branch](/carenas/sljit/tree/makefile) which had since been rebased.

FWIW I would rather have a failed build that is easy to fix than not having examples that build, so thanks for fixing that @nymacro

PR #60, opened 2020-07-23
Implement a new W^X allocator.

zherczeg:


sljit_src/sljitWXExecAllocator.c
```diff
@@ -0,0 +1,77 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/*
+   This file contains a simple W^X executable memory allocator for POSIX systems.
+
+   It allocates a separate block for each code block which may waste a lot of memory.
+*/
+
```
carenas, line 32: should include <sys/mman.h> here for safety
unistd.h for sysconf()

sljit_src/sljitWXExecAllocator.c
```diff
@@ -0,0 +1,77 @@
...
+}
+
+#define SLJIT_UPDATE_WX_FLAGS(from, to, enable_exec) \
+	sljit_update_wx_flags((from), (to), (enable_exec))
+
+static void sljit_update_wx_flags(void *from, void *to, sljit_s32 enable_exec)
+{
+	static sljit_uw page_mask = 0;
+
+	sljit_uw addr = (sljit_uw)from;
```
carenas, line 61: wouldn't it be better to use sljit_p here?

sljit_src/sljitWXExecAllocator.c
```diff
@@ -0,0 +1,77 @@
...
+   It allocates a separate block for each code block which may waste a lot of memory.
+*/
+
+SLJIT_API_FUNC_ATTRIBUTE void* sljit_malloc_exec(sljit_uw size)
+{
+	sljit_uw* ptr;
+
+	size += sizeof(sljit_uw);
+	ptr = (sljit_uw*)mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANON, -1, 0);
+
```
carenas, line 39: will add here the NetBSD specific code in a followup

sljit_src/sljitWXExecAllocator.c
```diff
@@ -0,0 +1,77 @@
...
+SLJIT_API_FUNC_ATTRIBUTE void sljit_free_exec(void* ptr)
+{
+	sljit_uw *start_ptr = ((sljit_uw*)ptr) - 1;
+	munmap(start_ptr, *start_ptr);
+}
+
+#define SLJIT_UPDATE_WX_FLAGS(from, to, enable_exec) \
+	sljit_update_wx_flags((from), (to), (enable_exec))
+
+static void sljit_update_wx_flags(void *from, void *to, sljit_s32 enable_exec)
```
carenas, line 57: IMHO might be better to pass a size here instead of "void *end", to avoid having to substract 2 (possible) unrelated pointers which is "Undefined Behaviour"

sljit_src/sljitWXExecAllocator.c
```diff
@@ -0,0 +1,77 @@
...
+	int prot = PROT_READ | (enable_exec ? PROT_EXEC : PROT_WRITE);
+
+	if (page_mask == 0) {
+		/* This is thread safe, since the value is constant. */
+		page_mask = (sljit_uw)(sysconf(_SC_PAGESIZE) - 1);
+	}
+
+	addr &= ~page_mask;
+
+	mprotect((void*)addr, (sljit_uw)to - addr, prot);
```
carenas, line 71: .. or casting the pointers to int

for this allocator, size should be always be the requested size anyway, but if some other allocator try to use this interface (and the system allows it) they might only switch part of the whole allocation in pagesize multiples

sljit_src/sljitWXExecAllocator.c
zherczeg: Cache flush uses the same interface. I would use it for consistency.

sljit_src/sljitWXExecAllocator.c
zherczeg: `sljit_p` is not used internally in the project.

sljit_src/sljitWXExecAllocator.c
zherczeg: I may missing something in your comment but the point is switching only a part of the allocated range.

carenas: mprotect works (and this is architecture/system specific) in pages at a minimum and old versions of it used to throw errors when passing sizes that were not multiples of the system pagesize to make the point clear.

in my implementation (which you copied) I mostly ignored that (since it works in NetBSD anyway, at least in amd64) but I was expecting (and getting) pagesize rounded values from mmap which I also didn't track exactly since munmap/mprotect didn't care either, and because @krytarowski was going to have to fix it when doing the same for bpfjit (in the NetBSD kernel)

this allocator SHOULD switch all the pages allocated between modes, and because if the size is bigger than a page then it will target the wrong page resulting in a segfault.

other allocators that might be using a macro like this might instead try to only change the page that they care for which could be smaller than the allocated size.

the complexities on calculating the end and start (as well as avoiding UB or using ints to do pointer subtraction) is why I suggested instead to pass a size_t even if it would be a different interface than cache clearing.

sljit_src/sljitNativeARM_32.c
```diff
@@ -470,15 +470,23 @@ static SLJIT_INLINE void inline_set_jump_addr(sljit_uw jump_ptr, sljit_sw execut
...
 				SLJIT_CACHE_FLUSH(inst, inst + 1);
 			}
 		} else {
+			if (flush_cache) {
+				SLJIT_UPDATE_WX_FLAGS(inst, inst + 2, 0);
+			}
 			inst[0] = (mov_pc & COND_MASK) | (BL - CONDITIONAL) | (diff & 0xffffff);
 			inst[1] = NOP;
 			if (flush_cache) {
+				SLJIT_UPDATE_WX_FLAGS(inst, inst + 2, 1);
```
carenas, line 489: inst is being updated below, so cache is flushed for a different address than what was just turned RX

same all others below

sljit_src/sljitNativeARM_64.c
```diff
@@ -2034,15 +2035,19 @@ SLJIT_API_FUNC_ATTRIBUTE struct sljit_put_label* sljit_emit_put_label(struct slj
...
+	SLJIT_UPDATE_WX_FLAGS(inst, inst + 4, 1);
 	inst = (sljit_ins *)SLJIT_ADD_EXEC_OFFSET(inst, executable_offset);
 	SLJIT_CACHE_FLUSH(inst, inst + 4);
 }
 
 SLJIT_API_FUNC_ATTRIBUTE void sljit_set_const(sljit_uw addr, sljit_sw new_constant, sljit_sw executable_offset)
 {
 	sljit_ins* inst = (sljit_ins*)addr;
+	SLJIT_UPDATE_WX_FLAGS(inst, inst + 4, 0);
 	modify_imm64_const(inst, new_constant);
```
carenas, line 2049: why not do the RX <-> RX swap once inside modify_imm64_const() instead?

sljit_src/sljitWXExecAllocator.c
krytarowski: I confirm that rounding to PAGE_SIZE is wanted. It is abstracted in userland, but in kernel we are stricter.

carenas:
there is a theoretical race condition where one thread could turn a page RW (ex: for updating a jump address) while another is trying to execute a function and that could result in crashes.

while it might be normally unlikely to happen in practice, might be a good idea to have a per allocation (worst case) or per page (minimum) lock but it will be difficult to enforce, since executing code is done transparently and the writes might affect pages using addresses, that while not the same, map to the same page when rounding.

krytarowski:
@carenas I propose to keep concurrent RW and RX activity as antipattern and bug.

sljit_src/sljitWXExecAllocator.c
```diff
@@ -0,0 +1,80 @@
...
+*/
+
+#include <sys/mman.h>
+#include <unistd.h>
+
+SLJIT_API_FUNC_ATTRIBUTE void* sljit_malloc_exec(sljit_uw size)
+{
+	sljit_uw* ptr;
+
+	size += sizeof(sljit_uw);
```
krytarowski, line 40: I think we should use here `max_align_t` to ensure that the data/code is properly aligned.

sljit_src/sljitWXExecAllocator.c
```diff
@@ -0,0 +1,80 @@
...
+	ptr = (sljit_uw*)mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANON, -1, 0);
```
krytarowski, line 41: It would be nice if we would always have calls to `mmap` that are multiples of `PAGE_SIZE`.

sljit_src/sljitWXExecAllocator.c
krytarowski: Alternatively... can we keep the list of allocations somewhere else in a dedicated structure?

krytarowski: Something similar is done in the RW+RX allocator, that we allocate:

```
struct chunk_header {
        void *executable;
};
```

We could go here for mmap()/malloc()ing:

```
struct chunk_header {
        void *address;
        sljit_uw size;
};
```

and then mmap PAGE_SIZE-rounded chunks.

sljit_src/sljitNativeX86_common.c
```diff
@@ -3117,15 +3121,19 @@ SLJIT_API_FUNC_ATTRIBUTE struct sljit_put_label* sljit_emit_put_label(struct slj
 SLJIT_API_FUNC_ATTRIBUTE void sljit_set_jump_addr(sljit_uw addr, sljit_uw new_target, sljit_sw executable_offset)
 {
 	SLJIT_UNUSED_ARG(executable_offset);
+	SLJIT_UPDATE_WX_FLAGS((void*)addr, (void*)addr + (sizeof(sljit_uw)), 0);
```
carenas, line 3124: for portability (void *) arithmetic should be avoided, probably use sljit_uw for the cast + 1?

same below

sljit_src/sljitNativeX86_common.c
```diff
@@ -3117,15 +3121,19 @@ SLJIT_API_FUNC_ATTRIBUTE struct sljit_put_label* sljit_emit_put_label(struct slj
...
 	sljit_unaligned_store_sw((void*)addr, (sljit_sw) new_target);
 #endif
+	SLJIT_UPDATE_WX_FLAGS((void*)addr, (void*)addr + (sizeof(sljit_uw)), 1);
 }
 
 SLJIT_API_FUNC_ATTRIBUTE void sljit_set_const(sljit_uw addr, sljit_sw new_constant, sljit_sw executable_offset)
 {
 	SLJIT_UNUSED_ARG(executable_offset);
+	SLJIT_UPDATE_WX_FLAGS((void*)addr, (void*)addr + (sizeof(sljit_sw)), 0);
 	sljit_unaligned_store_sw((void*)addr, new_constant);
```
carenas, line 3137: alternatively why not do the SLJIT_UPDATE_WX_FLAGS inside sljit_unaligned_store_sw to cover both sljit_set_jump_{addr, const}()

sljit_src/sljitWXExecAllocator.c
carenas: @krytarowski: max_align_t is C11 and so we will need to add compatibility for older compilers or drop support for little gain IMHO

I was using instead size_t which should be enough in practice (testing it works in all architectures NetBSD supports would help us to confirm that, as mentioned before), but considering sljit_uw is used already in sljit where size_t would had been used I see why this made sense here as well.

do you have any reason to believe some architectures will have problems with code aligned to 8 (in 64-bit) or 4 (in 32-bit) bytes and might need 16 byte alignment from the base of the page?, note that the code ptr is not accessed in ways that could cause alignment issues AFAIK

sljit_src/sljitNativeARM_32.c
zherczeg: This is tricky here. If an allocator uses WX, it does not use dual mapping. The opposite is also true, if an allocator uses dual mapping, it will not change WX flags. The code is also simpler this way.

sljit_src/sljitNativeARM_64.c
zherczeg: It looks like the whole function is a duplication. This could be size reduction optimization.

zherczeg:
> I propose to consider concurrent RW and RX activity as antipattern and bug.

I agree. If you modify the code, and the same code is executed elsewhere, it can lead to a crash. Code modification is not an atomic operation (can be sometimes, but we cannot rely on it).

sljit_src/sljitNativeX86_common.c
zherczeg: This is an unaligned address, and I know the compiler will not complain, but C expects us to use aligned pointers, so I choose this form since it represents a range of data rather than a value.

sljit_src/sljitWXExecAllocator.c
krytarowski: @carenas It depends on code generation. max_align_t is certainly safer and it can be easily implemented:

```
typedef union {
        void *_v;
        long double _ld;
        long long int _ll;
} sljit_max_align_t;

size += sizeof(sljit_max_align_t);
```

carenas:
> > I propose to consider concurrent RW and RX activity as antipattern and bug.
> 
> I agree. If you modify the code, and the same code is executed elsewhere, it can lead to a crash. Code modification is not an atomic operation (can be sometimes, but we cannot rely on it).

I agree (in principle) but note that this is another place where the abstraction details leak out because the crash won't be in the same memory address that was being modified, but somewhere else in the same page which just happened to become RW when jumped into.

if using the original RWX allocator (or even protexec) there wouldn't be a leak, and is definitely a bug to expect it to work but here it will and that should be at least documented somewhere to avoid surprises.

sljit_src/sljitNativeARM_32.c
```diff
@@ -491,28 +499,49 @@ static SLJIT_INLINE void inline_set_jump_addr(sljit_uw jump_ptr, sljit_sw execut
...
 			if (!bl) {
 				if (flush_cache) {
+					SLJIT_UPDATE_WX_FLAGS(inst, inst + 1, 1);
 					inst = (sljit_uw *)SLJIT_ADD_EXEC_OFFSET(inst, executable_offset);
 					SLJIT_CACHE_FLUSH(inst, inst + 1);
 				}
 			} else {
 				inst[1] = BLX | RM(TMP_REG1);
 				if (flush_cache) {
+					SLJIT_UPDATE_WX_FLAGS(inst, inst + 1, 1);
```
carenas, line 515: s/inst + 1/inst + 2/

sljit_src/sljitNativeARM_T2_32.c
```diff
@@ -2366,15 +2368,19 @@ SLJIT_API_FUNC_ATTRIBUTE struct sljit_put_label* sljit_emit_put_label(struct slj
...
 {
 	sljit_u16 *inst = (sljit_u16*)addr;
+	SLJIT_UPDATE_WX_FLAGS(inst, inst + 4, 0);
 	modify_imm32_const(inst, new_target);
+	SLJIT_UPDATE_WX_FLAGS(inst, inst + 4, 1);
 	inst = (sljit_u16 *)SLJIT_ADD_EXEC_OFFSET(inst, executable_offset);
 	SLJIT_CACHE_FLUSH(inst, inst + 4);
 }
 
 SLJIT_API_FUNC_ATTRIBUTE void sljit_set_const(sljit_uw addr, sljit_sw new_constant, sljit_sw executable_offset)
```
carenas, line 2378: this could be consolidated/inlined just like modify_imm64_* for other architectures

sljit_src/sljitNativeX86_common.c
carenas: then better use (char *) or its sljit equivalent (maybe sljit_u8 *), the problem I have is that sizeof(void *) != 1

sljit_src/sljitWXExecAllocator.c
```diff
@@ -0,0 +1,82 @@
...
+
+	if (ptr == NULL) {
```
carenas, line 43: s/NULL/MAP_FAILED

sljit_src/sljitWXExecAllocator.c
```diff
@@ -0,0 +1,82 @@
...
+	sljit_uw *start_ptr = ((sljit_uw*)ptr) - 1;
+	munmap(start_ptr, *start_ptr);
+}
+
+#define SLJIT_UPDATE_WX_FLAGS(from, to, enable_exec) \
+	sljit_update_wx_flags((from), (to), (enable_exec))
+
+static void sljit_update_wx_flags(void *from, void *to, sljit_s32 enable_exec)
+{
+	static sljit_uw page_mask = 0;
```
carenas, line 62: = 0 is implicit, maybe worth moving out to the file level so it could be used as well for the other 2 functions.

there is already one similar code in sljitUtils that might benefit as well

sljit_src/sljitWXExecAllocator.c
```diff
@@ -0,0 +1,82 @@
...
+
+	sljit_uw start = (sljit_uw)from;
+	sljit_uw end = (sljit_uw)to;
+	int prot = PROT_READ | (enable_exec ? PROT_EXEC : PROT_WRITE);
+
+	if (page_mask == 0) {
+		/* This is thread safe, since the value is constant. */
+		page_mask = (sljit_uw)(sysconf(_SC_PAGESIZE) - 1);
```
carenas, line 70: very unlikely but need to protect against sysconf() failing with -1 as it was done in all other places this code exists

sljit_src/sljitWXExecAllocator.c
carenas: @krytarowski the optional rounding available as part of followup PR #63

sljit_src/sljitWXExecAllocator.c
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	start &= ~page_mask;
+	end = (end + page_mask) & ~page_mask;
+
+	mprotect((void*)start, end - start, prot);
```
carenas, line 76: IMHO making this interface return int will allow propagating failures by doing instead `return mprotect()` here

zherczeg:
I did the merge and fixed a couple of things.

sljit_src/sljitNativeX86_common.c
carenas: this needs the same casting change done above to avoid (void *) arithmetic :

`
(void*)(addr + (sizeof(sljit_sw))
`

sljit_src/sljitWXExecAllocator.c
carenas: while I agree will be "nice", I am hard pressed to find a reason why it is needed; specially considering that to do so (as shown in #63) it requires some refactoring (as shown in bfc7c1fff192da921e7fb3949bcfcad107453451)

sljit_src/sljitWXExecAllocator.c
carenas: alternatively we could do the refactoring first and rebase this patch on top of it so this "bug" is never introduced but IMHO it might be too much churn for not much gain.

if you decide to do so squashing also the remaining changes from #63 would be worth too

test_src/sljitTest.c
```diff
@@ -6366,10 +6366,13 @@ static void test64(void)
 	CHECK(compiler);
 	sljit_free_compiler(compiler);
 
+#if !(defined SLJIT_WX_EXECUTABLE_ALLOCATOR && SLJIT_WX_EXECUTABLE_ALLOCATOR)
+	/* WX allocator may not reuse the address. */
```
carenas, line 6370: and neither might the other allocators if a new chunk was needed, hence making this test fragile.

FWIW it seems to fail pretty consistently in OpenBSD and can't figure out why it would work with its use of a memory address that has been made invalid by "free" so it might be worth skipping the whole test instead.

PR #61, opened 2020-07-23
PaX MPROTECT integration

carenas:
tested in NetBSD 9.0 amd64, NetBSD 8.2 i386 and NetBSD 7.2 i386

CC: @krytarowski

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ EXAMPLE_TARGET = $(BINDIR)/func_call $(BINDIR)/first_program $(BINDIR)/branch $(
 SLJIT_HEADERS = $(SRCDIR)/sljitLir.h $(SRCDIR)/sljitConfig.h $(SRCDIR)/sljitConfigInternal.h
 
 SLJIT_LIR_FILES = $(SRCDIR)/sljitLir.c $(SRCDIR)/sljitUtils.c \
-	$(SRCDIR)/sljitExecAllocator.c $(SRCDIR)/sljitProtExecAllocator.c \
+	$(SRCDIR)/sljitExecAllocator.c $(SRCDIR)/sljitProtExecAllocator.c $(SRCDIR)/sljitWXExecAllocator.c \
 	$(SRCDIR)/sljitNativeARM_32.c $(SRCDIR)/sljitNativeARM_T2_32.c $(SRCDIR)/sljitNativeARM_64.c \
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -32,7 +32,7 @@ EXAMPLE_TARGET = $(BINDIR)/func_call $(BINDIR)/first_program $(BINDIR)/branch $(
 SLJIT_HEADERS = $(SRCDIR)/sljitLir.h $(SRCDIR)/sljitConfig.h $(SRCDIR)/sljitConfigInternal.h
 
 SLJIT_LIR_FILES = $(SRCDIR)/sljitLir.c $(SRCDIR)/sljitUtils.c \
-	$(SRCDIR)/sljitExecAllocator.c $(SRCDIR)/sljitProtExecAllocator.c \
+	$(SRCDIR)/sljitExecAllocator.c $(SRCDIR)/sljitProtExecAllocator.c $(SRCDIR)/sljitWXExecAllocator.c \
```
zherczeg, line 35: This is missing from my patch. I am sorry.

carenas: feel free to squash this chunk and I'll rebase over, it is here mainly to easy testing

krytarowski:
The NetBSD specific integration is looking good.

carenas:
> The NetBSD specific integration is looking good.

the W^X allocator seems to work fine in Linux armv8l (Debian sid armel) using ARMv5 which I got through crostini/lxc in an arm64 chromebook

did you get any luck with sparc?; I suspect TileGX might not, but that is unlikely to be a regression

krytarowski:
Sparc should just work.. but if there are concerns I propose to add dedicated sljit-independent ATF NetBSD tests for all architectures and cover sparc too.

Once the patch will stabilize, I will spin a sparc virtual machine.

carenas:
you had concerns with alignment issues and sparc are the best canary architecture for those kind of problems since they don't handle misaligned access in hardware at all, and so will trap and sometimes crash.

I don't think the patch will have any significant changes might warrant waiting otherwise

carenas:
probably needs #63 and other changes like  72ca72c8e22f13d07b1b4125d70e3f07587f51c7 to be "complete", plus github's  automerge looks like a bad idea for repository history so it might be better rebasing this or even resubmitting.

feedback and test reports encouraged

PR #62, opened 2020-07-24
build: avoid -Wunused-parameter like warnings

carenas:
depending on the allocator used, the executable_offset parameter
might go unused.

there are already guards to avoid warnings in this case for x86
so do the same for all other architectures.

### DIFF
--- a/sljit_src/sljitNativeSPARC_32.c
+++ b/sljit_src/sljitNativeSPARC_32.c
@@ -267,6 +267,8 @@ SLJIT_API_FUNC_ATTRIBUTE void sljit_set_jump_addr(sljit_uw addr, sljit_uw new_ta
 {
 	sljit_ins *inst = (sljit_ins *)addr;
 
+	SLJIT_UNUSED_ARG(executable_offset);
+
 	SLJIT_ASSERT(((inst[0] & 0xc1c00000) == 0x01000000) && ((inst[1] & 0xc1f82000) == 0x80102000));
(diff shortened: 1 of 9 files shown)


carenas:
will conflict with #60 though, but either side should be easier to resolve; let me know if you would rather have those conflicts resolved and this rebased on top of that, otherwise

sljit_src/sljitNativeSPARC_32.c
```diff
@@ -276,6 +277,7 @@ SLJIT_API_FUNC_ATTRIBUTE void sljit_set_jump_addr(sljit_uw addr, sljit_uw new_ta
 
 SLJIT_API_FUNC_ATTRIBUTE void sljit_set_const(sljit_uw addr, sljit_sw new_constant, sljit_sw executable_offset)
 {
+	SLJIT_UNUSED_ARG(executable_offset);
```
zherczeg, line 280: I think variable declarations should come first. This is more compatible with older compilers.

zherczeg:
This is a good idea.

carenas:
> This is a good idea.

what is "this"?, rebasing it on top of the wxallocator code and therefore fixing the conflicts?

zherczeg:
Adding these macros. It makes the code nicer. I can land this before other patches.

carenas:
for the PCRE part of this "fix" see :

  https://bugs.exim.org/show_bug.cgi?id=2627

there are also additional changes needed queued in my [pedantic](/carenas/sljit/tree/pedantic) branch and that can be seen in 32397f143f1e2daf7eda50dea3ecbae8be824873

PR #63, opened 2020-07-24
Integration for "NetBSD kernel compatible" wx executable allocator

carenas:
this branch uses #61 and as such the first commit shown here is irrelevant, the next commits are described below and are also slightly independent but is recommended to keep them all while testing or integrating further changes. 

first commit is a generic cleanup as a prerequisite and as such touches also the original allocator, the wxexec part could be squashed, but is not critical for merging otherwise.  if squashed will be simpler if #60 is rebased on top of it as well.

last commit implements the "feature" which should also help for portability where page aligned pointers are needed and the sizes must be multiples of the pagesize.

will keep rebasing this branch as needed

### DIFF
--- a/sljit_src/sljitWXExecAllocator.c
+++ b/sljit_src/sljitWXExecAllocator.c
@@ -45,8 +45,9 @@
 SLJIT_API_FUNC_ATTRIBUTE void* sljit_malloc_exec(sljit_uw size)
 {
 	sljit_uw* ptr;
+	sljit_uw page_mask = (sljit_uw)get_page_alignment();
 
-	size += sizeof(sljit_uw);
+	size = (size + sizeof(sljit_uw) + page_mask) & ~page_mask;
 	ptr = (sljit_uw*)mmap(NULL, size, PROT_READ | PROT_WRITE | SLJIT_PROT_WX,
 				MAP_PRIVATE | MAP_ANON, -1, 0);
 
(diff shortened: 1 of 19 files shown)


sljit_src/sljitWXExecAllocator.c
```diff
@@ -64,17 +64,12 @@ SLJIT_API_FUNC_ATTRIBUTE void sljit_free_exec(void* ptr)
...
-	static sljit_uw page_mask = 0;
+	sljit_uw page_mask = (sljit_uw)get_page_alignment();
 
 	sljit_uw start = (sljit_uw)from;
 	sljit_uw end = (sljit_uw)to;
 	int prot = PROT_READ | (enable_exec ? PROT_EXEC : PROT_WRITE);
 
-	if (page_mask == 0) {
-		/* This is thread safe, since the value is constant. */
-		page_mask = (sljit_uw)(sysconf(_SC_PAGESIZE) - 1);
```
carenas, line 70: include is no longer needed after

https://github.com/zherczeg/sljit/blob/cdd8a685014af2680ac27a0ffe315b89cdc9d4c6/sljit_src/sljitWXExecAllocator.c#L34

carenas:
while testing in OpenBSD 6.7 amd64 had been sporadically getting the following assertion error from sljit_test

  Assertion failed at sljit_src/sljitNativeX86_common.c:617
  Abort trap (core dumped)

the issue is present all the way to #60 and strangely enough doesn't show any other visible "problems" with that assertion removed.

will provide a newer version with portability "fixes" on top to help testing, otherwise

zherczeg:
It is an issue. PATCH_MD should not be set for addr <= HALFWORD_MAX. It works, because the machine instruction works with any data, but compilers should not generate such code.

PR #64, opened 2020-07-26
wxexec: document caveats and recommendations

carenas:


zherczeg:
I did not close this. Seems github bug.

carenas:
guess it makes sense since github has no flow to allow rebasing a pull on top of another branch and the target was deleted without squashing (what I was hoping) or merging.

I will do another PR against master instead, and guess in retrospective everyone else that knew made the right choice by waiting for #60 to merge instead of trying to balance all commits on top.

Issue #56, opened 2020-07-20
Add support for a new allocator with RW <-> RX swapping


krytarowski:
There are architectural issues with handing RWX and double mapping RW + RX in the NetBSD kernel.

The RWX one is not portable to userland implementations through rumpkernel and requires disabled PaX MPROTECT hardening.

The double concurrent mapping RW+RX of the same physical pages has a few issues with cache, internal MMU handling etc.

There is need for another allocator that uses mprotect() [in userland, with a direct equivalent in the kernel] to switch between RW and RX mappings. mprotect() does the right thing with MMU, icache etc.

krytarowski:
CC: @carenas

zherczeg:
The core problem with flipping RX to RW is what happens if two threads use the same page: one for executing code and another for writing instructions. One possible solution is creating a separate allocator, which allocates pages for each code block individually but this may waste a lot of memory.

zherczeg:
Another question: if the OS has issues handling something valid, could it be fixed? Isn't this a security issue?

carenas:
> One possible solution is creating a separate allocator, which allocates pages for each code block individually but this may waste a lot of memory.

and it would still fail in NetBSD if the process fork() and the code needs to be updated, until there is a fix for lib/55177 (probably what is referred as "internal MMU handling" above).

carenas:
@krytarowski could you elaborate on the "issues with cache"?, there is an architecture/platform dependent SLJIT_CACHE_FLUSH() macro that should take care of it, and if it is not working will need fixing independently of what we can do to workaround the PaX MPROTECT issues raised here as the main issue.

the following program is what I used originally to test (independently of sljit), NetBSD's buggy mprotect/mremap workaround for PaX:

`
#include <sys/mman.h>
#include <sys/wait.h>
#include <unistd.h>
#include <string.h>
#include <stdio.h>

const int size = 4096;

int main()
{
  char *maprw = mmap(NULL, size,
                        PROT_READ|PROT_WRITE|PROT_MPROTECT(PROT_EXEC),
                        MAP_ANON|MAP_PRIVATE, -1, 0);
  char *maprx = mremap(maprw, size, NULL, size, MAP_REMAPDUP);
  mprotect(maprx, size, PROT_READ|PROT_EXEC);

  strcpy(maprw, "parent");
  printf("%s = %s\n", maprw, maprx);

  if (fork()) {
[...]

PR #67, opened 2020-07-31
wxexec: better support for SE systems including NetBSD

carenas:
Adds all missing parts to wxexec so it will be supported in NetBSD userspace (with or without MPROTECT) and could be used as a foundation for the macOS Big Sur work

Fixes: #56

### DIFF
--- a/sljit_src/sljitWXExecAllocator.c
+++ b/sljit_src/sljitWXExecAllocator.c
@@ -60,9 +60,56 @@
 
 #ifdef __NetBSD__
 #if defined(PROT_MPROTECT)
+#define check_se_protected(ptr, size) (0)
 #define SLJIT_PROT_WX PROT_MPROTECT(PROT_EXEC)
+#else /* !PROT_MPROTECT */
+#ifdef _NETBSD_SOURCE
+#include <sys/param.h>
+#else /* !_NETBSD_SOURCE */
+typedef unsigned int	u_int;
+#define devmajor_t sljit_s32
+#endif /* _NETBSD_SOURCE */
+#include <sys/sysctl.h>
+#include <unistd.h>
+
+#define check_se_protected(ptr, size) netbsd_se_protected()
+
+static SLJIT_INLINE int netbsd_se_protected(void)
+{
+	int mib[3];
+	int paxflags;
+	size_t len = sizeof(paxflags);
+
+	mib[0] = CTL_PROC;
+	mib[1] = getpid();
+	mib[2] = PROC_PID_PAXFLAGS;
+
+	if (SLJIT_UNLIKELY(sysctl(mib, 3, &paxflags, &len, NULL, 0) < 0))
+		return -1;
+
+	return (paxflags & CTL_PROC_PAXFLAGS_MPROTECT) ? -1 : 0;
+}
 #endif /* PROT_MPROTECT */
+#else /* POSIX */
+#define check_se_protected(ptr, size) generic_se_protected(ptr, size)
+
+static SLJIT_INLINE int generic_se_protected(void *ptr, sljit_uw size)
+{
+	if (SLJIT_LIKELY(!mprotect(ptr, size, PROT_EXEC)))
+		return mprotect(ptr, size, PROT_READ | PROT_WRITE);
+
+	return -1;
+}
 #endif /* NetBSD */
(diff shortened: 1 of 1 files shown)


sljit_src/sljitWXExecAllocator.c
```diff
@@ -27,61 +27,91 @@
...
    restrictions and adding exceptions to your application is not possible.
 */
 
+#include <sys/types.h>
 #include <sys/mman.h>
-#include <unistd.h>
+
+#ifdef __NetBSD__
+#if defined(PROT_MPROTECT)
+#define SLJIT_PROT_WX PROT_MPROTECT(PROT_EXEC)
```
krytarowski, line 56: All supported NetBSD version have `PROT_MPROTECT`.

carenas: but the sljit codebase still supports unsupported versions of NetBSD until #58 is merged

sljit_src/sljitWXExecAllocator.c
```diff
@@ -61,8 +70,9 @@
 SLJIT_API_FUNC_ATTRIBUTE void* sljit_malloc_exec(sljit_uw size)
 {
 	sljit_uw* ptr;
+	sljit_uw page_mask = (sljit_uw)get_page_alignment();
 
-	size += sizeof(sljit_uw);
+	size = (size + sizeof(sljit_uw) + page_mask) & ~page_mask;
```
carenas, line 75: @krytarowski do you have any comments on this and if we could ifdef it out so it only applies to the NetBSD kernel?

krytarowski: I'm waiting for @alnsn to update sljit for the NetBSD kernel.

sljit_src/sljitWXExecAllocator.c
```diff
@@ -45,28 +54,53 @@
...
+#else
+#define SLJIT_PROT_WX 0
+#endif /* PROT_MPROTECT */
+#else /* !NetBSD */
+#define SLJIT_PROT_WX 0
+#endif /* NetBSD */
+
 SLJIT_API_FUNC_ATTRIBUTE void* sljit_malloc_exec(sljit_uw size)
 {
+	static int se_protected = !SLJIT_PROT_WX;
```
zherczeg, line 72: Is this thread safe?

carenas: as much as all other static variables used the same way (note that the initial value is defined at compile time) in the codebase.

while not strictly code safe, it will be set to the same value for any concurrent threads anyway, so it is safe

zherczeg: Are you sure? This variable is set by `se_protected = mprotect...` and because it fails with one allocation, it may be sucessful with another.

carenas: the only way mprotect could fail, is if the system has a policy that prevents PROT_EXEC, and that can't change between threads in normal operation.

the binary could be given an exception (which would apply the next time it is loaded)

on the same line, I am assuming a change of system policy (or enforcement) should require a process reload to be effective, but that is not a "race condition"; but if not valid will invalidate the use of this variable to cache that state.

krytarowski: I think it's better to proactively check for W^X hardening on a per-platform basis.

Here is a code for this in LLVM:

https://github.com/llvm/llvm-project/blob/524f591a395cbb65a67c85b4df111c5f4e9d753e/compiler-rt/lib/sanitizer_common/sanitizer_linux.cpp#L2186

Whenever CheckMPROTECT() fails, we print an error to stdout and abort. This is verbose to users.

carenas: the sysctl() function/headers are itself a BSD extension though and so will fail to build in strict posix mode unless "hacked" : 

https://github.com/carenas/sljit/blob/e6e47a4da434a31894f198813f24a7f1266016ef/sljit_src/sljitWXExecAllocator.c#L68-L71

since that is unlikely to be a problem for most users, in supported (or even recently unsupported) versions it might be ok, and it indeed seems preferable to falling back to mprotect()

zherczeg (approved):
LGTM

Issue #68, opened 2020-08-02
test64 uses invalid memory address for labels and triggers failures due to undefined behaviour and integer wrap around


carenas:
easily to reproduce in OpenBSD (which has ASLR and other memory protections).

while using the WX executable allocator where the code pointer will change on each invocation and come from independent objects; unlike all other allocators, the pointer will be unmapped from the process memory after free and therefore any pointer arithmetic using it will be "Undefined Behaviour".

additionally, the sizes calculated (into label.size) are usually very large and risk wrapping over the upper limit of the max_label parameter in:

https://github.com/zherczeg/sljit/blob/d5f8923bdbc2a9ff3837f7951daad4397a564f10/sljit_src/sljitNativeX86_common.c#L560

which then results in the wrong flags being generated for the label.addr and therefore bad code generated which shows as an assert failure as reported in #63 

disabling the test as suggested in 78ca4769922890248083318421d97249322e52c2 makes sense (at least for the WX executable allocator) but rewriting it to use valid addresses (maybe through sljit_emit_label()) might be worth doing long term.

goes without saying; I am not really sure what put_labels are or where they will be useful, so pointing to some documentation or examples will be also encouraged.

PR #70, opened 2020-08-03
test64: improve reliability by making allocation of executable memory independent of the allocator

carenas:
Fixes: #68

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,17 +57,14 @@ $(BINDIR)/sljitLir.o : $(BINDIR)/.keep $(SLJIT_LIR_FILES) $(SLJIT_HEADERS)
 $(BINDIR)/sljitMain.o : $(TESTDIR)/sljitMain.c $(BINDIR)/.keep $(SLJIT_HEADERS)
 	$(CC) $(CPPFLAGS) $(CFLAGS) -c -o $@ $(TESTDIR)/sljitMain.c
 
-$(BINDIR)/sljitTest.o : $(TESTDIR)/sljitTest.c $(BINDIR)/.keep $(SLJIT_HEADERS)
-	$(CC) $(CPPFLAGS) $(CFLAGS) -c -o $@ $(TESTDIR)/sljitTest.c
-
 $(BINDIR)/regexMain.o : $(REGEXDIR)/regexMain.c $(BINDIR)/.keep $(SLJIT_HEADERS)
 	$(CC) $(CPPFLAGS) $(REGEX_CFLAGS) -c -o $@ $(REGEXDIR)/regexMain.c
 
 $(BINDIR)/regexJIT.o : $(REGEXDIR)/regexJIT.c $(BINDIR)/.keep $(SLJIT_HEADERS) $(REGEXDIR)/regexJIT.h
 	$(CC) $(CPPFLAGS) $(REGEX_CFLAGS) -c -o $@ $(REGEXDIR)/regexJIT.c
 
-$(BINDIR)/sljit_test: $(BINDIR)/.keep $(BINDIR)/sljitMain.o $(BINDIR)/sljitTest.o $(BINDIR)/sljitLir.o
-	$(CC) $(CFLAGS) $(LDFLAGS) $(BINDIR)/sljitMain.o $(BINDIR)/sljitTest.o $(BINDIR)/sljitLir.o -o $@ -lm -lpthread
+$(BINDIR)/sljit_test: $(BINDIR)/.keep $(BINDIR)/sljitMain.o $(TESTDIR)/sljitTest.c $(SRCDIR)/sljitLir.c
+	$(CC) -DSLJIT_MALLOC_EXEC=sljit_test_malloc_exec $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) $(BINDIR)/sljitMain.o $(TESTDIR)/sljitTest.c $(SRCDIR)/sljitLir.c -o $@ -lm -lpthread
 
 $(BINDIR)/regex_test: $(BINDIR)/.keep $(BINDIR)/regexMain.o $(BINDIR)/regexJIT.o $(BINDIR)/sljitLir.o
 	$(CC) $(CFLAGS) $(LDFLAGS) $(BINDIR)/regexMain.o $(BINDIR)/regexJIT.o $(BINDIR)/sljitLir.o -o $@ -lm -lpthread
--- a/sljit_src/sljitNativeX86_common.c
+++ b/sljit_src/sljitNativeX86_common.c
@@ -557,7 +557,7 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 					SLJIT_ASSERT(put_label->label);
 					put_label->addr = (sljit_uw)code_ptr;
 #if (defined SLJIT_CONFIG_X86_64 && SLJIT_CONFIG_X86_64)
-					code_ptr = generate_put_label_code(put_label, code_ptr, (sljit_uw)(SLJIT_ADD_EXEC_OFFSET(code, executable_offset) + put_label->label->size));
+					code_ptr = generate_put_label_code(put_label, code_ptr, (sljit_uw)SLJIT_ADD_EXEC_OFFSET(code, executable_offset) + put_label->label->size);
 #endif
 					put_label = put_label->next;
 					break;
--- a/test_src/sljitTest.c
+++ b/test_src/sljitTest.c
@@ -61,6 +61,8 @@ static sljit_s32 successful_tests = 0;
 static sljit_s32 verbose = 0;
 static sljit_s32 silent = 0;
 
+static void *test64_ptr_hack;
+
 #define FAILED(cond, text) \
 	if (SLJIT_UNLIKELY(cond)) { \
 		printf(text); \
@@ -73,6 +75,19 @@ static sljit_s32 silent = 0;
 		sljit_free_compiler(compiler); \
 		return; \
 	}
+
+/* only relevant for test64
+   requires an sljitLir built with the right -DSLJIT_MALLOC_EXEC flag */
+SLJIT_API_FUNC_ATTRIBUTE void *sljit_test_malloc_exec(sljit_uw size)
+{
+	if (SLJIT_LIKELY(!test64_ptr_hack))
+		return sljit_malloc_exec(size);
+
+	if ((sljit_uw)test64_ptr_hack == 64)
+		test64_ptr_hack = sljit_malloc_exec(size);
+
+	return test64_ptr_hack;
+}
 
 static void cond_set(struct sljit_compiler *compiler, sljit_s32 dst, sljit_sw dstw, sljit_s32 type)
 {
@@ -6283,10 +6298,10 @@ static void test63(void)
 
 static void test64(void)
 {
-	/* Test put label with false labels (small offsets).
-	   This code is allocator implementation dependent. */
+	/* Test put label with absolute label addresses
+	   This code is not thread safe and uses a patched execalloc. */
 	executable_code code;
-	sljit_sw malloc_addr;
+	sljit_uw malloc_addr;
 	struct sljit_label label[4];
 	struct sljit_put_label *put_label[2];
 	struct sljit_compiler* compiler;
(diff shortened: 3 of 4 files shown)


sljit_src/sljitNativeX86_common.c
```diff
@@ -557,7 +557,7 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 					SLJIT_ASSERT(put_label->label);
 					put_label->addr = (sljit_uw)code_ptr;
 #if (defined SLJIT_CONFIG_X86_64 && SLJIT_CONFIG_X86_64)
-					code_ptr = generate_put_label_code(put_label, code_ptr, (sljit_uw)(SLJIT_ADD_EXEC_OFFSET(code, executable_offset) + put_label->label->size));
+					code_ptr = generate_put_label_code(put_label, code_ptr, (sljit_uw)(SLJIT_ADD_EXEC_OFFSET(code, executable_offset)) + put_label->label->size);
```
carenas, line 560: wouldn't the fact that we are trying to use an out of scope address for a label be a reason to trigger a compilation error instead?

sljit_src/sljitNativeX86_64.c
```diff
@@ -76,7 +76,7 @@ static sljit_u8* generate_far_jump_code(struct sljit_jump *jump, sljit_u8 *code_
 
 static sljit_u8* generate_put_label_code(struct sljit_put_label *put_label, sljit_u8 *code_ptr, sljit_uw max_label)
 {
-	if (max_label > HALFWORD_MAX) {
+	if (max_label > HALFWORD_MAX && put_label->label->addr > HALFWORD_MAX) {
```
carenas, line 79: this is obviously incorrect as the rule should also take into consideration put_label->addr and indeed now throws an assert in Ubuntu 20.04 amd64 with this code and using the regular allocator (in test63) but works fine in macOS.

zherczeg:
I have made test64 a long time ago, and don't remember all the details. It seems it directly creates label addresses, which normally no program should do, and set these addresses as put_label targets. The labels has two members: `size` and `addr`, and the values of these are the key to understand this issue. The `addr` is normally computed during code generation, but probably these labels are not touched. 

`generate_put_label_code(put_label, code_ptr, (sljit_uw)(SLJIT_ADD_EXEC_OFFSET(code, executable_offset) + put_label->label->size))`

Perhaps the last argument computing causes the issue. This value might become grater than 2^32, so PATCH_MD is set, while the value in `addr` is less than 2^32, so the bigger space is not needed, and causes an assertion fail.

carenas:
> The `addr` is normally computed during code generation, but probably these labels are not touched.

the assert is triggered at "code generation" (see the backtrace in #63 for details), and yes.
it is IMHO not clear why the values where chosen for the hand made labels (other than maybe to make the test case hit more codepaths for better coverage), but the values used will trigger overflows as explained in #68.
 
> This value might become grater than 2^32, so PATCH_MD is set, while the value in `addr` is less than 2^32, so the bigger space is not needed, and causes an assertion fail.

correct, but it is also not clear from the code around generate_put_label_code(), why the end address of the allocated label by itself would be useful as an indicator that we would need "bigger space".  indeed the following code will usually go into the PATCH_MD path on my tests: 

```C
struct sljit_label *start = sljit_emit_label(c);
/* R1 = &start? */
struct sljit_put_label *put = sljit_emit_put_label(c, SLJIT_R1, 0);
sljit_set_put_label(put, start);
```

and resulting in (with the address changing on every run) : 

```ASM
movabs	rsi, 0x10a98901e
```

zherczeg:
This approach doesn't seem right. The `addr` represents the final physical address of something, but during code generation its value is not known for addresses which has not been reached yet. In this case the `size` attribute is used, which represents a higher bound to `addr`, and computed as `executable_code_start_ptr + size`. Hence the following must be true after the code generation is completed for all addresses: `addr <= executable_code_start_ptr + size`

The test wants to test low addresses (< 4GB), but it needs a right `size` attribute. You could set `addr` and `size` to the same value, since the `<=` must be true in this case. However, what happens if `executable_code_start_ptr + size` overflows resulting bad estimations? The test wants to avoid this by estimating the executable_code_start_ptr address. Parallel test running or random allocator results can hurt this test.

zherczeg:
Hm, it looks like the test even wants more, it wants that `addr == executable_code_start_ptr + size` in all cases, and low `addr` values triggers special code paths.

zherczeg:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitNativeX86_common.c#L535

Look at this code. When a label is reached, not only its `addr` but its `size` is updated as well, so future "predictions" gets the real value. That is why there are no checks for "reached" labels, and their `size` can be used all the time.

carenas:
> In this case the `size` attribute is used, which represents a higher bound to `addr`, and computed as `executable_code_start_ptr + size`.

except that the computated size was done backwards (hence always wrapping around) by subtracting the pointer from the small value provided (which is even worse with 64-bit addresses and will never be less than 4GB)

> The test wants to test low addresses (< 4GB)

then why try to base the size in a pointer that might not be less than 4GB to being with?; if we need addresses lower than 4GB then why not use a small offsets as shown in 418b1978ae11e92185cc6992853ae9e838f3d552; note thought with that commit only the assert has to go.

carenas:
> https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitNativeX86_common.c#L535
> 
> Look at this code. When a label is reached, not only its `addr` but its `size` is updated as well, so future "predictions" gets the real value. That is why there are no checks for "reached" labels, and their `size` can be used all the time.

I think that code doesn't apply for the hand made labels because they are never processed since they were never emitted to begin with.

which is why I considered at the end there was no way to test reliably hand made labels, which is why test64 was useless.

again, not sure what is hitting those lower than 4GB codepaths, but even when removing test64 the coverage doesn't drop.

zherczeg:
It might be the allocator, if it returns with something <4GByte. What we need to figure out is how to fix the test without changing the code.

carenas:
the test now fails consistently in all platforms, but at least this one can be fixed by removing that unnecessary (and potentially invalid and therefore triggering undefined behaviour) malloc pointer.

ironically the fourth commit is needed to avoid a (now reliable) assert when using hand made labels which should also go away once they are converted to emitted labels.

@zherczeg do you have any insights on where to put those emits so the generated labels still hit whatever codepaths the custom labels where supposed to hit, or it doesn't matter now that all addresses will be less than 4GB? (in linux, freebsd and whoever else might provide the MAP_32BIT flag); obviously broken (as it won't compile) in windows but it might had been always broken anyway and will be easier to fix.

sljit_src/sljitNativeX86_common.c
carenas: I can imagine someone creating a label by hand to point somewhere in memory (ex: some IO map) and therefore hitting this codepath if they made a mistake, so still think failing to compile will be helpful here and definitely preferably to writing somewhere else in memory and hopefully crashing soon after

sljit_src/sljitNativeX86_common.c
```diff
@@ -557,7 +573,8 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 					SLJIT_ASSERT(put_label->label);
 					put_label->addr = (sljit_uw)code_ptr;
 #if (defined SLJIT_CONFIG_X86_64 && SLJIT_CONFIG_X86_64)
-					code_ptr = generate_put_label_code(put_label, code_ptr, (sljit_uw)(SLJIT_ADD_EXEC_OFFSET(code, executable_offset) + put_label->label->size));
+					SLJIT_ASSERT(UWORD_MAX - (sljit_uw)code >= put_label->label->size);
```
zherczeg, line 576: `(sljit_uw)code + put_label->label->size >= put_label->label->size`

sljit_src/sljitNativeX86_common.c
```diff
@@ -614,11 +631,11 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 		sljit_unaligned_store_sw((void*)(put_label->addr - sizeof(sljit_sw)), (sljit_sw)put_label->label->addr);
 #else
 		if (put_label->flags & PATCH_MD) {
-			SLJIT_ASSERT(put_label->label->addr > HALFWORD_MAX);
+			SLJIT_ASSERT(put_label->label->addr + put_label->label->size > HALFWORD_MAX);
```
zherczeg, line 634: I think we have discussed that the original assert is correct.

Makefile
```diff
@@ -16,7 +16,7 @@ EXTRA_LDFLAGS=
 endif
 
 CPPFLAGS = $(EXTRA_CPPFLAGS) -DSLJIT_CONFIG_AUTO=1 -Isljit_src
-CFLAGS += -O2 -Wall
+CFLAGS += -g -O0 -Wall
```
zherczeg, line 19: This change is not relevant.

test_src/sljitTest.c
```diff
@@ -6281,10 +6282,33 @@ static void test63(void)
 	successful_tests++;
 }
 
+static SLJIT_FUNC sljit_u8 *test64_malloc_exec(sljit_uw size)
```
zherczeg, line 6285: Seems linux specific.

sljit_src/sljitNativeX86_common.c
carenas: only if we are assuming the compiler doesn't optimize it out

test_src/sljitTest.c
carenas: yes, since I was assuming you would test it there, eventhought it will build in alll POSIX and also do the right kind of allocation in FreeBSD; indeed it is all described in the description for the PR above

sljit_src/sljitNativeX86_common.c
carenas: without this change the assert will still trigger because the label addr won't be patched to the right value unless it is emitted by the compiler.

so it is correct only if not using hand made labels with big sizes and therefore needs this "tweak" to match the one above that assumed addr will always be updated to some value near code

zherczeg: We should not change it because of a test. It is probably better to disable the test if we cannot find a better soultion.

sljit_src/sljitNativeX86_common.c
zherczeg: I have worked on ECMAScript BigInts recently, and v8 js engine heavily rely on similar checks to detect carry, and I also have no trouble with it so far.

zherczeg:
I have an idea:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitConfigInternal.h#L614

We could redefine the testing system to use a custom allocator, which calls these functions. In case of test64, it could return the same area twice. Unless it is too complex to do it.

carenas:
I thought the fact we are finally returning addresses below 4GB was more in line with what the test needs.

If we get it to have a special case built wouldn't that be more important that trying to fix the hacky way it was going around guessing the label addr?, and even if we get that correct, isn't the offset calculation wrong and working in linux out of pure luck? (unless running with multiple threads).

zherczeg:
Yes, the test intends to check addr < 4GB, but the allocator might not satisfy this.

If the allocator returns with the same area twice when nothing else is allocated and you alloc / free / alloc with same size (this is true for all allocators except the newest) the test passes, because it can compute the proper size (the addr part is just a constant, the size is what is difficult).

carenas:
if you apply c89910b, test64() will NEVER pass regardless of what the allocator does, because it always relied in an invalid ptr and undefined behaviour that is now finally triggering an assert

zherczeg:
Obviously, but I would call it tricky not undefined. The trick is there is no signed and unsigned addition in cpu level, not even normal multiply. Only long multiply and division has sign.

carenas:
maybe I am not being clear; after that commit then the pointer operation that was overflowing (which is undefined behaviour) is gone, but still the unsigned word that is holding the address will wrap around and point to somewhere invalid so the assert will trigger and abort before that happens.

the reason why that address was so close to MAX_UINT64 (in 64bit) is because it was created by getting some small number (presumed to be signed by default) and substracting from it a huge one (which is also signed), so the result is a signed negative value that is then treated as unsigned.

no idea what all had to do with testing for addressed lower than 4GB, either.

zherczeg:
I understand what you are talking about, but operations are not undefined. Results are not random, maybe unexpected, but that is different thing if you don't know what you are doing.

zherczeg:
This direction looks better. I think we can do this without weak symbols. We know all allocators use the same C interface, and our custom allocator can be passed as -D macro during compilation. So create an object similar to `$(BINDIR)/sljitLirWeak.o` which gets extra C flags, e.g: `-D'SLJIT_MALLOC_EXEC(size)=sljit_test_malloc_exec(size)'`. Or similar to PCRE, the test could include sljitLir.c directly, and you can do configuration before including it. Or create a test specific header with a similar purpose to `sljitConfig.h`, but maybe doing it in a better way.

test_src/sljitTest.c
```diff
@@ -24,8 +24,13 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-/* Must be the first one. Must not depend on any other include. */
-#include "sljitLir.h"
+#define SLJIT_MALLOC_EXEC(size)	sljit_test_malloc_exec(size)
+#ifdef _WIN32
+static void *sljit_test_malloc_exec(unsigned long long size);
```
carenas, line 29: might need to also special case 64-bit windows

Makefile
```diff
@@ -66,8 +66,8 @@ $(BINDIR)/regexMain.o : $(REGEXDIR)/regexMain.c $(BINDIR)/.keep $(SLJIT_HEADERS)
 $(BINDIR)/regexJIT.o : $(REGEXDIR)/regexJIT.c $(BINDIR)/.keep $(SLJIT_HEADERS) $(REGEXDIR)/regexJIT.h
 	$(CC) $(CPPFLAGS) $(REGEX_CFLAGS) -c -o $@ $(REGEXDIR)/regexJIT.c
 
-$(BINDIR)/sljit_test: $(BINDIR)/.keep $(BINDIR)/sljitMain.o $(BINDIR)/sljitTest.o $(BINDIR)/sljitLir.o
-	$(CC) $(CFLAGS) $(LDFLAGS) $(BINDIR)/sljitMain.o $(BINDIR)/sljitTest.o $(BINDIR)/sljitLir.o -o $@ -lm -lpthread
+$(BINDIR)/sljit_test: $(BINDIR)/.keep $(BINDIR)/sljitMain.o $(TESTDIR)/sljitTest.c $(SRCDIR)/sljitLir.c
```
carenas, line 69: technically we could remove at this point the rule for $(BINDIR)/sljitTest.o that is no longer being used, but it allows for easily testing the broken binary by doing:

```
$ make bin/sljitTest.o
$ cc -DSLJIT_WX_EXECUTABLE_ALLOCATOR -DSLJIT_CONFIG_AUTO=1 -Isljit_src -O2 -Wall  bin/sljitMain.o bin/sljitTest.o bin/sljitLir.o -o bin/broken_test
```

it also makes sure it can build even if there is no SLJIT_MALLOC_EXEC definition even if it is not ideal.

PR #71, opened 2020-08-03
util: consolidate all allocators to use get_page_alignment()

carenas:
while at it, cleanup the comments and avoid mixed declarations with code
or unnecessary use of non sljit types in the original code.

### DIFF
--- a/sljit_src/sljitUtils.c
+++ b/sljit_src/sljitUtils.c
@@ -163,9 +163,6 @@ SLJIT_API_FUNC_ATTRIBUTE void SLJIT_FUNC sljit_release_lock(void)
 #endif /* MAP_ANONYMOUS */
 #endif /* !MAP_ANON */
 
-/* For detecting the page size. */
-#include <unistd.h>
-
 #ifndef MAP_ANON
 
 #include <fcntl.h>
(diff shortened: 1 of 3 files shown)


sljit_src/sljitUtils.c
```diff
@@ -208,6 +205,41 @@ static SLJIT_INLINE sljit_s32 open_dev_zero(void)
...
 #endif /* SLJIT_EXECUTABLE_ALLOCATOR || SLJIT_UTIL_GLOBAL_LOCK */
 
+#if (defined SLJIT_UTIL_STACK && SLJIT_UTIL_STACK) \
+	|| (defined SLJIT_EXECUTABLE_ALLOCATOR && SLJIT_EXECUTABLE_ALLOCATOR)
+
+#ifdef _WIN32
+
+SLJIT_INLINE static sljit_sw get_page_alignment(void) {
+	SYSTEM_INFO si;
+	static sljit_sw sljit_page_align;
```
zherczeg, line 215: When this function is inlined multiple times, is it true that only a single shared static variable is created?

zherczeg:
I might not be clear but I have no issue with inlining the function, my question was only about static variables inside inline functions.

sljit_src/sljitUtils.c
carenas: since there is only compilation unit (with all other code included into sljitLir.c) it should be fine, but in case someone compiles sljitUtil independently it is safer if not inlined (it didn't help much anyway)

carenas:
inlining (at the function level) doesn't affect the way a static variable is instantiated (it is always defined globally but only visible within the function where it is declared)

PR #72, opened 2020-08-06
wxexec: windows support

carenas:
while at it make some minor matching changes to the POSIX code to easy
maintenance.

### DIFF
--- a/sljit_src/sljitWXExecAllocator.c
+++ b/sljit_src/sljitWXExecAllocator.c
@@ -25,7 +25,8 @@
  */
 
 /*
-   This file contains a simple W^X executable memory allocator for POSIX systems.
+   This file contains a simple W^X executable memory allocator for POSIX
+   like systems and windows
 
    It allocates a separate block for each code block and may waste a lot of memory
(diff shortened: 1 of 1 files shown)


zherczeg:
I don't remember anybody complained that normal allocator does not work on Windows. Do we need this?

carenas:
> I don't remember anybody complained that normal allocator does not work on Windows.

I suspect we might, my CI does a windows build (using github's Windows Server 2019 based VM) and for some weird reason the 32-bit version builds fine but doesn't seem to run as you can see (compare it with the output of the x64 run) in : 

  https://github.com/carenas/sljit/runs/936307992?check_suite_focus=true

I only have windows 10 and can't reproduce, but microsoft introduced DEP long ago, and the default setting for server has been moving towards disabling by default for a while (with 32-bit being the most likely to benefit since it has NX hardware support)

this version will be DEP compatible (or so I think) and while it still makes sense to recommend anyone to instead add an exception to their binaries instead, this might be an alternative for anyone that can't.

> Do we need this?

mainly so my CI could run using github actions, I wouldn't recommend it to anyone else

PR #73, opened 2020-08-06
util: cleanup

carenas:
Visual Studio 2019 throws a C4206 (file is empty) warning (only with /W4) while processing sljitUtils.c, which is apparently a bug on their part, and that this refactoring doesn't workaround.

still while a little intrusive, the result is more readable and less code.

as a plus the executable allocator macro makes the use of this util consistent with the rest and could help someone stuck with a non optimizer linker

### DIFF
--- a/sljit_src/sljitUtils.c
+++ b/sljit_src/sljitUtils.c
@@ -28,11 +28,10 @@
 /*  Locks                                                                   */
 /* ------------------------------------------------------------------------ */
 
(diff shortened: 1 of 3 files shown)


sljit_src/sljitUtils.c
```diff
@@ -28,64 +28,65 @@
...
-		allocator_mutex = CreateMutex(NULL, TRUE, NULL);
-	else
-		WaitForSingleObject(allocator_mutex, INFINITE);
+	pthread_mutex_lock(&allocator_mutex);
 }
 
 static SLJIT_INLINE void allocator_release_lock(void)
 {
-	ReleaseMutex(allocator_mutex);
+	pthread_mutex_unlock(&allocator_mutex);
```
zherczeg, line 75: If a new macro is introduced, why don't we drop these inline functions?

PR #74, opened 2020-08-06
build: pedantic

carenas:
mostly fixes some build issues when using C89/C90

### DIFF
--- a/sljit_src/sljitConfigInternal.h
+++ b/sljit_src/sljitConfigInternal.h
@@ -81,7 +81,7 @@ extern "C" {
 
    Other macros:
      SLJIT_FUNC : calling convention attribute for both calling JIT from C and C calling back from JIT
-     SLJIT_W(number) : defining 64 bit constants on 64 bit architectures (compiler independent helper)
+     SLJIT_W(number) : defining 64 bit constants on 64 bit architectures (platform independent helper)
 */
 
 /*****************/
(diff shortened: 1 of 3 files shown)


sljit_src/sljitConfigInternal.h
```diff
@@ -451,10 +451,14 @@ typedef double sljit_f64;
 #if (defined SLJIT_CONFIG_UNSUPPORTED && SLJIT_CONFIG_UNSUPPORTED)
 #define SLJIT_W(w)	(w##l)
 #elif (defined SLJIT_64BIT_ARCHITECTURE && SLJIT_64BIT_ARCHITECTURE)
+#ifdef _WIN32
```
carenas, line 454: this could be _WIN64 but _WIN32 is more commonly used and both would work, hopefully for the foreseeable future

zherczeg: `#if (defined _WIN32 || defined _WIN64)` ?

carenas: only one is needed as it is documented[1], that _WIN32 is ALSO defined in 64-bit windows.

my point was that _WIN64 was the obvious one to use, since this code is 64-bit specific and hopefully Windows arm64 is also LLP64 (something that would be nice to confirm), but in retrospective, the fact I felt it was worth mentioning makes it less intuitive and might be work correcting

[1] https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019

PR #75, opened 2020-08-06
config: refactor fastcall setup

carenas:
this is the first commit (which is safe and ready to merge independently) of two, let me know otherwise on ways to improve it.

the second commit (c5d1355559d3482d130935c3374c3e5ba41bcc9c) does too much guessing of intention, and hence it is published mainly to gather feedback (even if it is simpler) as it will be better to understand why the exception was added when it doesn't seem to be needed.

the last version of macOS with x86-32 support will be EOL soon but wouldn't count on that considering they are still people using the last version of macOS with PowerPC that was EOL more than a decade ago.

### DIFF
--- a/sljit_src/sljitConfigInternal.h
+++ b/sljit_src/sljitConfigInternal.h
@@ -551,17 +551,19 @@ typedef double sljit_f64;
 
 #ifndef SLJIT_FUNC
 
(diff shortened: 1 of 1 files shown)


zherczeg:
What is the gain of this patch?

carenas:
> What is the gain of this patch?

avoids breaking the build in haiku x86, or anyone else that is using a gcc old enough not to have the fastcall attribute

PR #76, opened 2020-08-07
util: fix race for initialization of executable allocator lock in windows

carenas:
fixes the known race condition for the initialization of the executable allocator lock in windows (which also affects the unused global lock, which has been left intentionally untouched)

Fixes: PCRE BUG2377[1]

[1] https://bugs.exim.org/show_bug.cgi?id=2377

### DIFF
--- a/sljit_src/sljitUtils.c
+++ b/sljit_src/sljitUtils.c
@@ -28,46 +28,40 @@
 /*  Locks                                                                   */
 /* ------------------------------------------------------------------------ */
 
+/* Executable Allocator */
+
 #if (defined SLJIT_EXECUTABLE_ALLOCATOR && SLJIT_EXECUTABLE_ALLOCATOR) \
 	&& !(defined SLJIT_WX_EXECUTABLE_ALLOCATOR && SLJIT_WX_EXECUTABLE_ALLOCATOR)
-
 #if (defined SLJIT_SINGLE_THREADED && SLJIT_SINGLE_THREADED)
-
 #define SLJIT_ALLOCATOR_LOCK()
 #define SLJIT_ALLOCATOR_UNLOCK()
+#elif !(defined _WIN32)
+#include <pthread.h>
 
-#else /* SLJIT_EXECUTABLE_ALLOCATOR && !SLJIT_WX_EXECUTABLE_ALLOCATOR */
+static pthread_mutex_t allocator_lock = PTHREAD_MUTEX_INITIALIZER;
 
-#ifdef _WIN32
-
-static HANDLE allocator_mutex = 0;
+#define SLJIT_ALLOCATOR_LOCK() pthread_mutex_lock(&allocator_lock)
+#define SLJIT_ALLOCATOR_UNLOCK() pthread_mutex_unlock(&allocator_lock)
+#else /* windows */
+static HANDLE allocator_lock;
 
 static SLJIT_INLINE void allocator_grab_lock(void)
 {
-	/* Static mutexes should never fail... */
-	if (!allocator_mutex)
-		allocator_mutex = CreateMutex(NULL, TRUE, NULL);
-	else
-		WaitForSingleObject(allocator_mutex, INFINITE);
+	HANDLE lock;
+	if (SLJIT_UNLIKELY(!allocator_lock)) {
+		lock = CreateMutex(NULL, FALSE, NULL);
+		if (InterlockedCompareExchangePointer(&allocator_lock, lock, NULL))
+			CloseHandle(lock);
+	}
+	WaitForSingleObject(allocator_lock, INFINITE);
 }
 
 #define SLJIT_ALLOCATOR_LOCK() allocator_grab_lock()
-#define SLJIT_ALLOCATOR_UNLOCK() ReleaseMutex(allocator_mutex)
-
-#else /* !_WIN32 */
-
-#include <pthread.h>
-
-static pthread_mutex_t allocator_mutex = PTHREAD_MUTEX_INITIALIZER;
-
-#define SLJIT_ALLOCATOR_LOCK() pthread_mutex_lock(&allocator_mutex)
-#define SLJIT_ALLOCATOR_UNLOCK() pthread_mutex_unlock(&allocator_mutex)
-
-#endif /* _WIN32 */
-
-#endif /* SLJIT_SINGLE_THREADED */
-
+#define SLJIT_ALLOCATOR_UNLOCK() ReleaseMutex(allocator_lock)
+#endif /* thread implementation */
 #endif /* SLJIT_EXECUTABLE_ALLOCATOR && !SLJIT_WX_EXECUTABLE_ALLOCATOR */
+
+/* Global */
 
 #if (defined SLJIT_UTIL_GLOBAL_LOCK && SLJIT_UTIL_GLOBAL_LOCK)
 
--- a/sljit_src/sljitLir.c
+++ b/sljit_src/sljitLir.c
@@ -28,7 +28,6 @@
 
 #ifdef _WIN32
 
-/* For SLJIT_CACHE_FLUSH, which can expand to FlushInstructionCache. */
 #include <windows.h>
 
 #endif /* _WIN32 */


zherczeg:
Cannot we have both?

carenas:
> Cannot we have both?

Only if we are willing to maintain it, and it will be definitely a lot more complicated than the newer/safer version.

Additionally, having an initialization function usually complicates the logic from thread agnostic libraries that might include this code (ex: PCRE) further

sljit_src/sljitUtils.c
```diff
@@ -28,47 +28,48 @@
...
+#include <pthread.h>
 
-#else /* SLJIT_EXECUTABLE_ALLOCATOR && !SLJIT_WX_EXECUTABLE_ALLOCATOR */
+static pthread_mutex_t allocator_lock = PTHREAD_MUTEX_INITIALIZER;
 
-#ifdef _WIN32
+#define SLJIT_ALLOCATOR_LOCK() pthread_mutex_lock(&allocator_lock)
+#define SLJIT_ALLOCATOR_UNLOCK() pthread_mutex_unlock(&allocator_lock)
+#else /* windows */
+#if _WIN32_WINNT >= 0x0600
```
zherczeg, line 46: `#elif` can be used here

sljit_src/sljitUtils.c
```diff
@@ -28,47 +28,46 @@
...
 static SLJIT_INLINE void allocator_grab_lock(void)
 {
-	/* Static mutexes should never fail... */
-	if (!allocator_mutex)
-		allocator_mutex = CreateMutex(NULL, TRUE, NULL);
-	else
-		WaitForSingleObject(allocator_mutex, INFINITE);
+	HANDLE lock;
+	if (!allocator_lock) {
+		lock = CreateMutex(NULL, TRUE, NULL);
```
carenas, line 57: should use FALSE instead

zherczeg:
What happened with the Vista+ code? It looked good, since you can use locks without allocating a resource.

carenas:
> What happened with the Vista+ code?

After a testing scare I decided to remove it and make sure the code is consistent and more transparent by just having one implementation instead of two.

The main problem with it, is that as coded was selecting at compile time which implementation to use, so for example in ReactOS will use the hand made version because their SDK support is as old as the long EOL versions of windows I decided to ignore originally.

In Windows 10, will use the fallback if I used mingw-gcc (or I am crosscompiling from linux/macOS using mingw64) and unless I defined WIN32_WINNT=0x0600 to override their default, but the newer code if using Visual Studio (at least the version I have. but might change with old enough versions), clang will likely do the same if it came from Visual Studio but might not if it was installed directly and bound to a different SDK (like the one from mingw), so the issue is not even consistent for the same compiler.

the hand made code is less efficient and might have bugs (indeed there is an optimization intentionally missing for maintainability) but should be sufficiently bug free to work in all cases and performant enough (it will be used only a few times until the lock is initialized and jump over most of the time), and most importantly more maintainable since if it has a bug, everyone will be affected. 

> It looked good, since you can use locks without allocating a resource.

not sure how it is implemented but I don't think an extra pointer in userspace might make much of a difference in its overall resource use.

I presume though that it will be more optimized and tested than my hand made solution, which is why I suggested originally to instead use that exclusively.

maybe we can bring it back with a compile time flag like SLJIT_UTIL_LOCK_NON_EOL_VERSIONS so people that can drop support for Windows XP and Windows Server 2003 can enable it knowingly at compile time, or drop support of it once ReactOS catches up or we are willing to drop support (even Microsoft itself doesn't support anything older than Windows 8 nowadays, and only for extended support, which is not even that relevant).

sljit_src/sljitUtils.c
```diff
@@ -28,47 +28,41 @@
...
-#else /* SLJIT_EXECUTABLE_ALLOCATOR && !SLJIT_WX_EXECUTABLE_ALLOCATOR */
-
-#ifdef _WIN32
+static pthread_mutex_t allocator_lock = PTHREAD_MUTEX_INITIALIZER;
 
-static HANDLE allocator_mutex = 0;
+#define SLJIT_ALLOCATOR_LOCK() pthread_mutex_lock(&allocator_lock)
+#define SLJIT_ALLOCATOR_UNLOCK() pthread_mutex_unlock(&allocator_lock)
+#else /* windows */
+static HANDLE allocator_lock;
```
dangelog, line 46: Are you sure this doesn't need to be volatile? Otherwise the load on line 51 is not atomic nor ordered. (Assuming volatile in C on Windows has the same semantics as in C++). If non-MSVC is supported, then does line 51 need to become a fully fledged atomic load, like InterlockedCompareExchangePointer(allocator_lock, NULL, NULL)?

carenas: volatile in C doesn't imply atomic, the load/store is atomic because of the CPU.

I would expect the same in C++ but will be interested on some RTFM you might have, as well on that non-MSVC constrain that might require and extra Interlocked Exchange.  we definitely support gcc and clang in windows and should be compiler agnostic as much as possible.

regarding ordering, I will look again but this was written under the assumption x86 was being used and this might had change with the recent introduction of arm; is that something you had observed?, I definitely have no windows on ARM to test on.

dangelog: The relevant documentation is

https://docs.microsoft.com/en-us/cpp/cpp/volatile-cpp?view=vs-2019#end-of-iso-compliant

On MSVC, in C++, by default on non-ARM (= x86 and x64), volatile not only implies atomicity but also synchronization (writes have release semantics, reads have acquire semantics). I cannot find any documentation regarding the behavior for C -- is it the same?

To say it all, I don't know if on Windows 
1) non-MSVC compilers are supported by sljit in the first place;
2) non-MSVC compilers would honour the non-standard MSVC behavior here;
3) non-x86 is supported

but I'm pretty sure that _some_ synchronization is still needed. The normal theory is that, otherwise, the compiler and/or the CPU are allowed to reorder the store of the allocator_lock pointer before e.g. committing the memory pointed to by it; another thread might find the allocator_lock set, pointing to garbage. (Of course, this might not actually be the case on x86.) Therefore, the suggestion of using volatile and/or loading via Interlocked APIs (which ensures full ordering, according to https://docs.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchangepointer#remarks "This function generates a full memory barrier (or fence) to ensure that memory operations are completed in order.").

carenas: as explained before :

1. we aim to support ALL compilers and ALL architectures, have you seen a problem with any, or is this just theoretical?
2. using the non standard volatile is not even recommended by MSVC as per the provided documentation, so regardless if it is honoured or not by other compilers seems like a bad thing to add eitherway.
3. atomicity is warranted by the CPU and ordering might be affected in ARM (since this code predates "Windows on ARM"), but while I don't have a way to test it, I would expect that since the variable is static and global concerns with reordering of local variables doesn't apply and writing it is behind a memory barrier through the existing  InterlockedCompareExchangePointer() call (as you pointed out).

dangelog: 1. it's "theoretical" in the sense that the code is formally wrong. It may happen to work because of the combination of compiler+CPU won't do anything evil. It doesn't mean the code is correct. Sticking ourselves to Windows, its supported platforms, and the fact that the code is C98, then you need a C98 way to perform an acquire-load (read barrier). A way to do it, without relying on the MSVC-specific behavior, is `InterlockedCompareExchangePointer(allocator_lock, NULL, NULL)`.
2. It is absolutely not recommended. But you have nothing else in C++98/C99, which have no notion of "threads", so you rely on platform-specific behavior anyways for threading, atomicity and synchronization. The recommended, portable way is atomics (C++11/C11). 
3. "since the variable is static and global concerns with reordering of local variables doesn't apply" this simply doesn't exist. The fact that the variable is global doesn't change anything regarding its access. You need a write barrier (provided by the Interlocked API) and a corresponding read barrier, which is missing from the code. (As an example, the load on ARM has to be a LDAR or similar, not a LDR).

zherczeg: Note: `InterlockedCompareExchangePointer` is used below.

dangelog: To write (line 53), but not to read (line 51). That's my concern. Even if allocator_lock itself is set atomically (which, pedantically, it's platform-specific behavior since we're in pre-C11 land), the memory pointed to by it isn't acquired because there's no barrier through this load.

zherczeg: I may misunderstand something but that function is designed to change things. Is there a variant for reading atomically?

carenas: > 1. it's "theoretical" in the sense that the code is formally wrong. It may happen to work because of the combination of compiler+CPU won't do anything evil. It doesn't mean the code is correct. Sticking ourselves to Windows, its supported platforms, and the fact that the code is C98, then you need a C98 way to perform an acquire-load (read barrier). A way to do it, without relying on the MSVC-specific behavior, is `InterlockedCompareExchangePointer(allocator_lock, NULL, NULL)`.

true, but the point was that it wouldn't change in practical terms (we indeed have several of those races in the codebase, where there is a global static being updated racily but it doesn't matter in practice because it will always store the same value at the end.

in this case, you can either read NULL or a not NULL value (which will always be the same because there is a memory barrier when writing it that ensures so and because writing or reading a pointer is atomic since it matches the size of the register that will be used and is done with one single instruction), without having to incur on the additional cost of a read barrier (note this is not an spinlock, just a plain regular mutex).

> 3. "since the variable is static and global concerns with reordering of local variables doesn't apply" this simply doesn't exist. The fact that the variable is global doesn't change anything regarding its access. You need a write barrier (provided by the Interlocked API) and a corresponding read barrier, which is missing from the code. (As an example, the load on ARM has to be a LDAR or similar, not a LDR).

my comment was about reordering the reads in the stack (which is naturally different per thread), where you are correct that the ordering could be changed without a barrier and so it could use an uninitialized value, but since this variable is global, it is already initialized at compile time to 0, so even without the barrier you CAN'T get any other value unless it has been written (and that is by design), which is why it is expected we will have looser threads that will then go and discard their own bogus handles (see the description in the commit) in line 54

https://github.com/zherczeg/sljit/blob/502744f5ba74504bd39e87f734a41b6504cd24c1/sljit_src/sljitUtils.c#L54

it will be interesting to see how this code gets compiled regardless, so if you have access to windows on arm, I am happy to see some disassembling.  I was expecting to see some  LDREX or other similar instructions, and will be interesting to see the disassembling of the Vista+ functions[1] we could have used instead as well.

and interesting article I just happen to come by a few days ago about this in :

  https://devblogs.microsoft.com/dotnet/arm64-performance-in-net-5/

[1] https://github.com/zherczeg/sljit/commit/518d3eb0359f5562b4a253b05a00b783084e0e5e

dangelog: > true, but the point was that it wouldn't change in practical terms (we indeed have several of those races in the codebase, where there is a global static being updated racily but it doesn't matter in practice because it will always store the same value at the end.

N wrongs in the code don't make a right. We're going in circles: the code is formally incorrect; you're just admitting that "it works in practice". What's the big problem at making it formally correct by adding a barrier on the load?

> in this case, you can either read NULL or a not NULL value (which will always be the same because there is a memory barrier when writing it that ensures so and because writing or reading a pointer is atomic since it matches the size of the register that will be used and is done with one single instruction), without having to incur on the additional cost of a read barrier (note this is not an spinlock, just a plain regular mutex).

This has longly been disproved; from "Perils of double check locking" and other similar papers by Meyers, Boehm, Williams and so on. The usual counterproof is that an "evil" compiler and/or an "evil" architecture may load through the pointer by finding it non null, but the CPU core performing the load of the data pointed to didn't flush the cachelines of the data pointed to -- because noone told that core to do so, because there was no read barrier --, so it reads garbage.

zherczeg: I don`t mind adding this check to the code, unless its overhead is measurable. Please submit a patch.

The truth is nobody ever complained this before (and the old code was even worse). The "evil compiler / arch" requirement explains this. If the attacker can replace the victim's compiler or their cpu, they can do far more damage in much simpler ways. Normal systems honor zero initialized data, and clears this data before the code is executed, so it does not contain random data.

PR #78, opened 2020-08-07
util: retire global lock

carenas:
split into 2 phases to allow for a staged release for anyone that might care about ABI compatibility, but considering it is buggy and most likely not used it might as well be done in one fell swoop

### DIFF
--- a/sljit_src/sljitConfig.h
+++ b/sljit_src/sljitConfig.h
@@ -63,10 +63,10 @@ extern "C" {
 /*  Utilities                                                            */
 /* --------------------------------------------------------------------- */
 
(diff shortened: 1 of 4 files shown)


zherczeg:
This is the same code which is used by the allocator. What is the problem with it?

carenas:
> This is the same code which is used by the allocator.

It is no longer the same after fad5ead630ac8388d84729f9de3f715e895916b6

> What is the problem with it?

it is not very useful and hard to use properly (if you need a lock, it is better if you add one to your application instead of using this) and has bugs, so why not get rid of it?

zherczeg (approved):
LGTM

PR #80, opened 2020-08-13
config: enable auto detection by default

carenas:
instead of needing -DSLJIT_CONFIG_AUTO or some other definition, do auto detection unless a specific definition was passed or it was explicitly disabled by -DSLJIT_CONFIG_AUTO=0

in platforms (ex: armel or armhf) where there might be multiple possible targets to chose (ex: thumb2, v5 or v7) allows reconfiguring the build without having to edit the makefile, for example: 

```
$ make EXTRA_CPPFLAGS=-DSLJIT_CONFIG_ARM_THUMB2
```

additionally, this allows to generate code for a different (but compatible) ABI (ex: ARM EABI 5 application generating Thumb2 code in an arm64 CPU/OS)


sljit_src/sljitConfigInternal.h
```diff
@@ -126,7 +126,20 @@ extern "C" {
 /* Automatic CPU detection (requires compiler support). */
 /********************************************************/
 
-#if (defined SLJIT_CONFIG_AUTO && SLJIT_CONFIG_AUTO)
+#if !((defined SLJIT_CONFIG_X86_32 && SLJIT_CONFIG_X86_32) \
```
zherczeg, line 129: The long `#if` above could handle this by defining SLJIT_CONFIG_AUTO if it is not defined and no other cpu is explicitly selected.

carenas: it would throw some warnings if some value other than 0 or 1 is passed with SLJIT_CONFIG_AUTO, but looks nicer

sljit_src/sljitConfigInternal.h
```diff
@@ -117,11 +100,30 @@ extern "C" {
 	+ (defined SLJIT_CONFIG_MIPS_32 && SLJIT_CONFIG_MIPS_32) \
 	+ (defined SLJIT_CONFIG_MIPS_64 && SLJIT_CONFIG_MIPS_64) \
 	+ (defined SLJIT_CONFIG_SPARC_32 && SLJIT_CONFIG_SPARC_32) \
-	+ (defined SLJIT_CONFIG_AUTO && SLJIT_CONFIG_AUTO) \
```
zherczeg, line 120: Don't remove this line.

sljit_src/sljitConfigInternal.h
```diff
@@ -122,6 +105,27 @@ extern "C" {
...
+	|| (defined SLJIT_CONFIG_ARM_THUMB2 && SLJIT_CONFIG_ARM_THUMB2) \
+	|| (defined SLJIT_CONFIG_ARM_64 && SLJIT_CONFIG_ARM_64) \
+	|| (defined SLJIT_CONFIG_PPC_32 && SLJIT_CONFIG_PPC_32) \
+	|| (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64) \
+	|| (defined SLJIT_CONFIG_MIPS_32 && SLJIT_CONFIG_MIPS_32) \
+	|| (defined SLJIT_CONFIG_MIPS_64 && SLJIT_CONFIG_MIPS_64) \
+	|| (defined SLJIT_CONFIG_SPARC_32 && SLJIT_CONFIG_SPARC_32) \
+	|| (defined SLJIT_CONFIG_TILEGX && SLJIT_CONFIG_TILEGX) \
+	|| (defined SLJIT_CONFIG_UNSUPPORTED && SLJIT_CONFIG_UNSUPPORTED) \
+	|| (defined SLJIT_CONFIG_AUTO && SLJIT_CONFIG_AUTO) \
```
zherczeg, line 121: Can the last two conditions be merged?
`|| (!defined SLJIT_CONFIG_AUTO || !SLJIT_CONFIG_AUTO)`

sljit_src/sljitConfigInternal.h
```diff
@@ -122,6 +105,27 @@ extern "C" {
...
+	|| !defined SLJIT_CONFIG_AUTO)
+#error "An architecture must be selected"
+#else
+#undef SLJIT_CONFIG_AUTO
```
zherczeg, line 125: Do we undef SLJIT_CONFIG_AUTO whenever the error is not triggered above?

zherczeg (changes requested):
Maybe the steps could be the following:
```
#if !(ANY_ARCH_OR_AUTO_DEFINED_AND_SET)
#if AUTO_DEFINED_AND_DISABLED
#error "An arch must be selected"
#else
define auto
#endif
#endif
```

sljit_src/sljitConfigInternal.h
carenas: no; we undef it here only to avoid a redefinition error below if it was defined and not 1.

we could drop the undef (as it was done in the previous version) and just print the warning as well as an option, but this feels cleaner (eventhough having that value set to something other than 0 or 1 was unexpected)

sljit_src/sljitConfigInternal.h
carenas: note the whole condition is !() so the first one means SLJIT_CONFIG_AUTO is defined and 1 (actually, non zero), and the second one means SLJIT_CONFIG_AUTO is undefined.

you are correct that the whole "defined" check is irrelevant but it might be better to just replace it with ") || !SLJIT_CONFIG_AUTO"

in the same line we could get rid of all those other "defined" checks and rely on the C99 rule §6.10.1 ¶3 that states all undefined macros default to 0, or invert the logic and only care about any of them being defined.

carenas: a simpler (an not redundant for SLJIT_CONFIG_AUTO) conditional in last revision, but note that it misconfigures the code (or maybe configures it correctly but that breaks the build) for at least SLJIT_CONFIG_UNSUPPORTED (hence why in draft)

FWIW, it might seem that with the previous logic some of the SLJIT_CONFIG_X86_64 were still being used when SLJIT_CONFIGU_UNSUPPORTED was defined

zherczeg:
The last variant is still incorrect. You can define these with the value of 0, and that is the same thing as being not defined. I have proposed a variant above, that should not have this weakness.

sljit_src/sljitConfigInternal.h
```diff
@@ -122,6 +105,27 @@ extern "C" {
 #error "Multiple architectures are selected"
 #endif
 
+#if !(defined SLJIT_CONFIG_X86_32 && SLJIT_CONFIG_X86_32) \
+	|| !(defined SLJIT_CONFIG_X86_64 && SLJIT_CONFIG_X86_64) \
```
carenas, line 109: makes more sense with A && B .. && Z or !(A || B .. || Z) but funny enough this also works because of the rule above that prevents having more than one of them

sljit_src/sljitConfigInternal.h
```diff
@@ -122,6 +105,28 @@ extern "C" {
...
+	&& !(defined SLJIT_CONFIG_MIPS_32 && SLJIT_CONFIG_MIPS_32) \
+	&& !(defined SLJIT_CONFIG_MIPS_64 && SLJIT_CONFIG_MIPS_64) \
+	&& !(defined SLJIT_CONFIG_SPARC_32 && SLJIT_CONFIG_SPARC_32) \
+	&& !(defined SLJIT_CONFIG_TILEGX && SLJIT_CONFIG_TILEGX) \
+	&& !(defined SLJIT_CONFIG_UNSUPPORTED && SLJIT_CONFIG_UNSUPPORTED) \
+	&& !(defined SLJIT_CONFIG_AUTO && SLJIT_CONFIG_AUTO)
+#if defined SLJIT_CONFIG_AUTO && !SLJIT_CONFIG_AUTO
+#error "An architecture must be selected"
+#else /* SLJIT_CONFIG_AUTO */
+#undef SLJIT_CONFIG_AUTO
```
zherczeg, line 125: One more minor thing: at this point I think `SLJIT_CONFIG_AUTO` cannot be defined. That is the point of the previous checks.

zherczeg (approved):
LGTM

PR #83, opened 2020-08-16
Support allocator data for executable allocators

zherczeg:
Furthermore support pre and post configuration headers for the compiler.

### DIFF
--- a/test_src/sljitTest.c
+++ b/test_src/sljitTest.c
@@ -61,8 +61,6 @@ static sljit_s32 successful_tests = 0;
 static sljit_s32 verbose = 0;
 static sljit_s32 silent = 0;
 
-static void *test64_ptr_hack;
-
 #define FAILED(cond, text) \
 	if (SLJIT_UNLIKELY(cond)) { \
 		printf(text); \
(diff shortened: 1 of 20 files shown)


test_src/sljitTest.c
```diff
@@ -76,17 +74,19 @@ static void *test64_ptr_hack;
...
-		test64_ptr_hack = sljit_malloc_exec(size);
+	return SLJIT_BUILTIN_MALLOC_EXEC(size, exec_allocator_data);
+}
 
-	return test64_ptr_hack;
+/* For interface testing. */
+void sljit_test_free_code(void* code, void *exec_allocator_data)
+{
+	SLJIT_BUILTIN_FREE_EXEC(code, exec_allocator_data);
 }
```
carenas, line 90: should move below the !SLJIT_CONFIG_UNSUPPORTED below (won't fix the build though, but it will help)

test_src/sljitTest.c
```diff
@@ -6318,29 +6319,27 @@ static void test64(void)
 		printf("Run test64\n");
 
 	/* lock next allocation; see sljit_test_malloc_exec() */
-	test64_ptr_hack = (void *)64;
-	malloc_addr = (sljit_uw)SLJIT_MALLOC_EXEC(1024);
+	malloc_addr = (sljit_uw)SLJIT_MALLOC_EXEC(4096, NULL);
```
carenas, line 6322: the generated code used to fit in 130 or so bytes, using 1024 (as done originally) should allow for it to be allocated that exactly and probably remove the need for the max_offset?

zherczeg:
Thank you for the review. I forgot about SLJIT_CONFIG_UNSUPPORTED. I think I fixed all issues.

PR #88, opened 2020-08-24
util: add O_CLOEXEC to open_dev_zero and fix building without MAP_ANON in macOS

carenas:
open_dev_zero is inconsistently defined and used, so make sure that
all invocations use dev_zero internally in a thread safe way, and
while at it, make sure it doesn't leak the file handle on fork() if
support for O_CLOEXEC is available.

### DIFF
--- a/sljit_src/sljitUtils.c
+++ b/sljit_src/sljitUtils.c
@@ -65,7 +65,11 @@ static SLJIT_INLINE void allocator_grab_lock(void)
 /*  Stack                                                                   */
 /* ------------------------------------------------------------------------ */
 
-#if (defined SLJIT_UTIL_STACK && SLJIT_UTIL_STACK) || (defined SLJIT_EXECUTABLE_ALLOCATOR && SLJIT_EXECUTABLE_ALLOCATOR)
+#if ((defined SLJIT_UTIL_STACK && SLJIT_UTIL_STACK) \
+	&& !(defined SLJIT_UTIL_SIMPLE_STACK_ALLOCATION && SLJIT_UTIL_SIMPLE_STACK_ALLOCATION)) \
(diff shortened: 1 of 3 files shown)


sljit_src/sljitUtils.c
```diff
@@ -82,14 +86,22 @@ static SLJIT_INLINE void allocator_grab_lock(void)
 
 #include <fcntl.h>
 
+#ifdef O_CLOEXEC
+#define SLJIT_CLOEXEC	O_CLOEXEC
+#else
+#define SLJIT_CLOEXEC	0
+#endif
```
zherczeg, line 93: else/endif missing comment

zherczeg (changes requested):
Only a minor change is needed.

sljit_src/sljitUtils.c
```diff
@@ -82,14 +86,22 @@ static SLJIT_INLINE void allocator_grab_lock(void)
...
-static sljit_s32 dev_zero = -1;
+static int dev_zero = -1;
 
 #if (defined SLJIT_SINGLE_THREADED && SLJIT_SINGLE_THREADED)
 
-static SLJIT_INLINE sljit_s32 open_dev_zero(void)
+static SLJIT_INLINE int open_dev_zero(void)
 {
-	dev_zero = open("/dev/zero", O_RDWR);
+	if (SLJIT_UNLIKELY(dev_zero < 0))
```
carenas, line 102: this could be taken out, which then makes this change mainly about adding O_CLOEXEC

carenas:
> Only a minor change is needed.

while re-reading the code, realized though the original inconsistency was actually a performance optimization which when added back might make the changes around the nested if unnecessary.

should we keep the original code or use the slightly more compact new one?

zherczeg (approved):
LGTM

zherczeg:
I cannot land a draft patch

PR #90, opened 2020-08-29
Apple Bigsur support

carenas:
Has been tested extensibly (albeit not using real hardware/software but a POSIX look alike) and while is expected to work it might not be as efficient/secure/elegant as it could be, because it needs to return writeable pages for consistency with the other allocators, and also needs to update metadata in those executable pages and therefore has to make the pages writeable more often than ideal.

Since enabling MAP_JIT adds undocumented restrictions to the use of those pages, would be important to confirm that notarized applications with the right entitlement (com.apple.security.cs.allow-jit) to allow creating JIT pages, can still use those pages after fork() and if it doesn't raise the issue with Apple ASAP.

if fork() is no longer restricted it might be worth enabling MAP_JIT unconditionally also in Intel (at least for BigSur) but that will need further testing and therefore is punted from this PR.
 
Fixes: #51

### DIFF
--- a/sljit_src/sljitExecAllocator.c
+++ b/sljit_src/sljitExecAllocator.c
@@ -79,6 +79,7 @@
 */
 
 #ifdef _WIN32
+#define SLJIT_UPDATE_WX_FLAGS(from, to, enable_exec)
 
 static SLJIT_INLINE void* alloc_chunk(sljit_uw size)
 {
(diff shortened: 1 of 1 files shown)


sljit_src/sljitExecAllocator.c
```diff
@@ -91,78 +91,88 @@ static SLJIT_INLINE void free_chunk(void *chunk, sljit_uw size)
...
-
+	if (map_jit_flag < 0) {
 		map_jit_flag = 0;
 		uname(&name);
-
-		/* Kernel version for 10.14.0 (Mojave) */
-		if (atoi(name.release) >= 18) {
+		os_kernel_major = atoi(name.release);
+
+		/* this is likely to cause problems as seen in PCRE BUG2334 */
```
carenas: FWIW I haven't seen any of the reported problems while testing in 10.15.1 (Catalina), so it might be as well safe to use MAP_JIT by default.

obviously it will cause mmap to fail if the hardened runtime is in use AND the application doesn't have the com.apple.security.cs.allow-jit entitlement, but that is the expected behaviour.

MAP_JIT doesn't prevent creating more than 1 map or inheriting the map through fork() like it does in 10.13.6

sljit_src/sljitExecAllocator.c
```diff
@@ -189,7 +200,10 @@ static SLJIT_INLINE void free_chunk(void *chunk, sljit_uw size)
 	munmap(chunk, size);
 }
 
-#endif
+#endif /* windows */
+#ifndef SLJIT_UPDATE_WX_FLAGS
+#define SLJIT_UPDATE_WX_FLAGS(from, to, enable_exec)
+#endif /* !SLJIT_UPDATE_WX_FLAGS */
```
carenas, line 203: alternatively, this could be defined in the windows code to avoid the awkward duplicate ifndef

sljit_src/sljitExecAllocator.c
```diff
@@ -94,7 +94,6 @@ static SLJIT_INLINE void free_chunk(void *chunk, sljit_uw size)
 #else /* POSIX */
 
 #ifdef __APPLE__
-#include <AvailabilityMacros.h>
```
carenas: since there is only one macOS version that runs in arm, there is no need to look at the version, obsoleting this

sljit_src/sljitExecAllocator.c
```diff
@@ -115,19 +115,14 @@ static SLJIT_INLINE int get_map_jit_flag()
...
 		uname(&name);
-		os_kernel_major = atoi(name.release);
 
-		/* this is likely to cause problems as seen in PCRE BUG2334 */
-		if (os_kernel_major > 19)
-			map_jit_flag = MAP_JIT;
-		else if (os_kernel_major >= 18) {
-			/* Kernel version for 10.14.0 (Mojave) or later */
+		/* Kernel version for 10.14.0 (Mojave) or later */
+		if (atoi(name.release) >= 18) {
```
carenas: alternatively, if MAP_JIT is considered safe could keep the original logic but most likely only special casing Mojave for doing the test

sljit_src/sljitExecAllocator.c
```diff
@@ -142,25 +137,27 @@ static SLJIT_INLINE int get_map_jit_flag()
...
-#else /* !MAP_JIT */
-#define SLJIT_MAP_JIT	(0)
-#endif /* MAP_JIT */
-
-#if defined(MAC_OS_X_VERSION_MAX_ALLOWED) \
-	&& MAC_OS_X_VERSION_MAX_ALLOWED >= 110000
+#else /* !SLJIT_CONFIG_X86 */
+#if !(defined SLJIT_CONFIG_ARM && SLJIT_CONFIG_ARM)
+#error Unsupported architecture
+#endif /* SLJIT_CONFIG_ARM */
```
carenas, line 143: technically not needed and could be made more strict by checking for ARM_64, but will hopefully help when apple does its next CPU migration

zherczeg:
This patch looks ok, but it would be good if somebody with hw could try it.

Keno:
I tried this patch on hardware (applied to the pcre2 vendored version) and I am seeing the following crash:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x10be5c008)
  * frame #0: 0x000000011099b910 libpcre2-8.0.dylib`sljit_malloc_exec(size=64) at sljitExecAllocator.c:304:20
    frame #1: 0x000000011099b5d0 libpcre2-8.0.dylib`pcre2_jit_compile_8(code=0x000000010e8ebe80, options=1) at pcre2_jit_compile.c:14200:15

(lldb) f 0
frame #0: 0x000000011099b910 libpcre2-8.0.dylib`sljit_malloc_exec(size=64) at sljitExecAllocator.c:304:20
   301 		chunk_size -= sizeof(struct block_header);
   302 		total_size += chunk_size;
   303 	
-> 304 		header->prev_size = 0;
   305 		if (chunk_size > size + 64) {
   306 			/* Cut the allocated space into a free and a used block. */
   307 			allocated_size += size;
```

carenas:
> I tried this patch on hardware (applied to the pcre2 vendored version) and I am seeing the following crash:

if you want to backport it to an older version than pcre 10.36-RC1[1] then you will need to apply more patches than just this one so the crash makes sense

if by "vendored" you mean your own local copy of it, then you will have to first update it to the RC1; apologize for not being clear about it.

[1] https://lists.exim.org/lurker/message/20201106.173459.b8ecdf06.en.html

Keno:
This was the version of sljit in PCRE 10.36-RC1.

carenas:
there must be a problem then on the way it is detecting that it gets compiled for macOS BigSur (make sure it uses the right SDK and minimum OS target or you have other restrictions (maybe a notarized app)?

if the later will need to also apply the correct entitlements at codesign time (see the carenas/sljit fork for hints on how to do that within sljit and built through CI)

```
<dict>
<key>com.apple.security.cs.allow-jit</key>
<true/>
</dict>
```

sljit_src/sljitExecAllocator.c
```diff
@@ -91,78 +92,84 @@ static SLJIT_INLINE void free_chunk(void *chunk, sljit_uw size)
...
+#include <pthread.h>
+
+#define SLJIT_MAP_JIT	(MAP_JIT)
+#define SLJIT_UPDATE_WX_FLAGS(from, to, enable_exec) \
+                        apple_update_wx_flags(enable_exec)
+
+static SLJIT_INLINE void apple_update_wx_flags(sljit_s32 enable_exec)
+{
+	pthread_jit_write_protect_np(!enable_exec);
```
Keno, line 152: Are you sure that this inversion here is correct? https://developer.apple.com/documentation/apple_silicon/porting_just-in-time_compilers_to_apple_silicon seems to indicate that `true` passed to this function sets permissions to `R-X`.

Keno: I verified it works if I delete the inversion, so 
```suggestion
	pthread_jit_write_protect_np(enable_exec);
```

carenas:
the last version, including @Keno bugfix to the only line of code that was not exercised through the "emulator" should be safe to merge IMHO

zherczeg (approved):
LGTM

wtdcode:
> I tried this patch on hardware (applied to the pcre2 vendored version) and I am seeing the following crash:
> 
> ```
> * thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x10be5c008)
>   * frame #0: 0x000000011099b910 libpcre2-8.0.dylib`sljit_malloc_exec(size=64) at sljitExecAllocator.c:304:20
>     frame #1: 0x000000011099b5d0 libpcre2-8.0.dylib`pcre2_jit_compile_8(code=0x000000010e8ebe80, options=1) at pcre2_jit_compile.c:14200:15
> 
> (lldb) f 0
> frame #0: 0x000000011099b910 libpcre2-8.0.dylib`sljit_malloc_exec(size=64) at sljitExecAllocator.c:304:20
>    301 		chunk_size -= sizeof(struct block_header);
>    302 		total_size += chunk_size;
>    303 	
> -> 304 		header->prev_size = 0;
>    305 		if (chunk_size > size + 64) {
>    306 			/* Cut the allocated space into a free and a used block. */
>    307 			allocated_size += size;
> ```

Take my whole day to figure out why such crash happens...... Thanks for your fix.

Issue #89, opened 2020-08-24
Port PCRE2 JIT to Linux on IBMz (s390x)


edelsohn:
Enable PCRE2 JIT on Linux on IBMz (s390x-linux) and optimize to achieve equivalent speedup over non-JIT code as x86_64.  Goal is full functionality and passing testsuite with JIT enabled.

An unfinished port of PCRE2 JIT to s390x exists in the Linux-on-IBM-z Github account and can be used as a starting point.  IBM will contribute the code as necessary.

https://github.com/linux-on-ibm-z/pcre2/tree/s390x-experimental

A $5000 bounty from IBM available posted on Bountysource
https://www.bountysource.com/issues/92353837-port-pcre2-jit-to-linux-on-ibmz-s390x

Systems access is available through the LinuxONE Community Cloud at Marist
https://linuxone.cloud.marist.edu/#/register?flag=VM

This issue is tied to a feature request to the Exim community for PCRE2 support.
https://bugs.exim.org/show_bug.cgi?id=2635

edelsohn:
@zherczeg Do you know of someone in the PCRE2 JIT community who would like to work on this issue?

zherczeg:
Perhaps @carenas might be interested.
As for me I am working in a University which has many industrial collaborations, that could be an option, although that will be more pricey.

carenas:
the [development branch s390x](/carenas/sljit/s390x) shows (only in the cloud though, not qemu) some encouraging progress : 

$ bin/regex_test
Pass -v to enable verbose, -s to disable this hint.

REGEX tests: all tests are PASSED on s390x 64bit (big endian + unaligned)

pcre2 itself fails (even without JIT) and as zherczeg pointed out will likely need a significant amount of work to get close to that performance objective

edelsohn:
I can't access the branch created by carenas.  Is that the same as the branch from linux-on-ibm-z?

How much effort do you estimate to make PCRE2 functional?

How much effort to optimize it for the platform?

What amount of bounty would make it interesting, probably split into sub-milestones.

carenas:
> I can't access the branch created by carenas. Is that the same as the branch from linux-on-ibm-z?

my bad, fixed the link.  it is mostly the linux-on-ibm-z code but with a few fixes on top to fight the bitrot so it will build on top of the current tree for further development.

> How much effort

[...]

Issue #49, opened 2020-07-06
More than 3 arguments


fgmccabe:
How is one supposed to handle a function that has more than 3 arguments? 
One would normally spill these on the stack; but the callee needs to access them.
Could you point me to the appropriate place to look?
Thanks

zherczeg:
Currently sljit supports maximum of four arguments. They can be word, int32, single or double precision. Each ABI defines how these arguments are passed to the callee. Here are some examples:

- Some ABIs passes all data on stack, some uses registers for the first n arguments

- Some ABIs use the first n registers, others keep their argument position index.
Example: f(int a, double b, int c, double d)
ABI type1: a -> reg0, c -> reg1, b -> float_reg0, d -> float_reg1
ABI type 2: a -> reg0, b-> float_reg1, c -> reg2, d -> float_reg3

- Some ABIs squeze arguments into as little space as possible, others leave gaps.
Example: f(float a, double b, float c) // ecah passed on stack
ABI type 1: push a, push c, push b (required size 16 bytes)
ABI type 2, push a, skip 4 bytes, push b, push c, skip 4 bytes (required size 24 bytes)

If you want to find an ABI for the target, just search "ABI cput type" in google.
Linux, x86-32: https://01.org/sites/default/files/file_attach/intel386-psabi-1.0.pdf
Linux, x64: https://uclibc.org/docs/psABI-x86_64.pdf
Windows, x64: https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention?view=vs-2019
ARM-32 : https://developer.arm.com/architectures/system-architectures/software-standards/abi
ARM-64: https://c9x.me/compile/bib/abi-arm64.pdf
PowerPC: https://refspecs.linuxfoundation.org/ELF/ppc64/PPC-elf64abi.html
etc.

It is a lot of work to support them all. To tell you the truth supporting any kind of arguments is beyond the scope of a lightweight jit. But if you prefer using stack, you can create stack space for your arguments, and pass the starting address of the argument space to the called function.

PR #91, opened 2020-09-05
s390x: initial port

carenas:
implements (eventhough not in the cleanest or more performant way) all functionality that is required to pass all tests in both sljit and pcre2, but still lacking : 

- support for FPU and related functions (including vector operations)
- several unimplemented or partially implemented opcodes/features (obviously none needed by PCRE with the target CPU)
- no optimizations or profiling (including plenty of TODO that might need tackling before this can be released to users)

to enable will need to explicitly define SLJIT_CONFIG_S390X at compile time and it is known to work with z12 or later (using a Linux ONE remote machine[0], either directly or through CI) with a expected target of z13 or z14 for the needed vector support that is required for a high performance JIT in PCRE2.

Partially closes: #89

[0] https://linuxone.cloud.marist.edu/#/register?flag=VM

### DIFF
--- a/sljit_src/sljitConfigInternal.h
+++ b/sljit_src/sljitConfigInternal.h
@@ -100,6 +100,7 @@ extern "C" {
 	+ (defined SLJIT_CONFIG_MIPS_32 && SLJIT_CONFIG_MIPS_32) \
 	+ (defined SLJIT_CONFIG_MIPS_64 && SLJIT_CONFIG_MIPS_64) \
 	+ (defined SLJIT_CONFIG_SPARC_32 && SLJIT_CONFIG_SPARC_32) \
+	+ (defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X) \
 	+ (defined SLJIT_CONFIG_AUTO && SLJIT_CONFIG_AUTO) \
 	+ (defined SLJIT_CONFIG_UNSUPPORTED && SLJIT_CONFIG_UNSUPPORTED) >= 2
 #error "Multiple architectures are selected"
@@ -117,6 +118,7 @@ extern "C" {
 	&& !(defined SLJIT_CONFIG_MIPS_64 && SLJIT_CONFIG_MIPS_64) \
 	&& !(defined SLJIT_CONFIG_SPARC_32 && SLJIT_CONFIG_SPARC_32) \
 	&& !(defined SLJIT_CONFIG_TILEGX && SLJIT_CONFIG_TILEGX) \
+	&& !(defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X) \
 	&& !(defined SLJIT_CONFIG_UNSUPPORTED && SLJIT_CONFIG_UNSUPPORTED) \
 	&& !(defined SLJIT_CONFIG_AUTO && SLJIT_CONFIG_AUTO)
 #if defined SLJIT_CONFIG_AUTO && !SLJIT_CONFIG_AUTO
@@ -160,6 +162,8 @@ extern "C" {
 #define SLJIT_CONFIG_SPARC_32 1
 #elif defined(__tilegx__)
 #define SLJIT_CONFIG_TILEGX 1
+#elif defined(__s390x__)
+#define SLJIT_CONFIG_S390X 1
 #else
 /* Unsupported architecture */
 #define SLJIT_CONFIG_UNSUPPORTED 1
@@ -413,7 +417,8 @@ typedef long int sljit_sw;
 	&& !(defined SLJIT_CONFIG_ARM_64 && SLJIT_CONFIG_ARM_64) \
 	&& !(defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64) \
 	&& !(defined SLJIT_CONFIG_MIPS_64 && SLJIT_CONFIG_MIPS_64) \
-	&& !(defined SLJIT_CONFIG_TILEGX && SLJIT_CONFIG_TILEGX)
+	&& !(defined SLJIT_CONFIG_TILEGX && SLJIT_CONFIG_TILEGX) \
+	&& !(defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X)
 #define SLJIT_32BIT_ARCHITECTURE 1
 #define SLJIT_WORD_SHIFT 2
 typedef unsigned int sljit_uw;
@@ -510,7 +515,8 @@ typedef double sljit_f64;
 
 #endif /* !SLJIT_MIPS_REV */
 
-#elif (defined SLJIT_CONFIG_SPARC_32 && SLJIT_CONFIG_SPARC_32)
+#elif (defined SLJIT_CONFIG_SPARC_32 && SLJIT_CONFIG_SPARC_32) \
+	|| (defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X)
 
 #define SLJIT_BIG_ENDIAN 1
 
@@ -537,7 +543,8 @@ typedef double sljit_f64;
 	|| (defined SLJIT_CONFIG_ARM_THUMB2 && SLJIT_CONFIG_ARM_THUMB2) \
 	|| (defined SLJIT_CONFIG_ARM_64 && SLJIT_CONFIG_ARM_64) \
 	|| (defined SLJIT_CONFIG_PPC_32 && SLJIT_CONFIG_PPC_32) \
-	|| (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64)
+	|| (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64) \
+	|| (defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X)
 #define SLJIT_UNALIGNED 1
 #endif
 
@@ -714,6 +721,12 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_sw sljit_exec_offset(void* ptr);
 #define SLJIT_NUMBER_OF_REGISTERS 10
 #define SLJIT_NUMBER_OF_SAVED_REGISTERS 5
 #define SLJIT_LOCALS_OFFSET_BASE 0
+
+#elif (defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X)
+
+#define SLJIT_NUMBER_OF_REGISTERS 12
+#define SLJIT_NUMBER_OF_SAVED_REGISTERS 8
+#define SLJIT_LOCALS_OFFSET_BASE 160
 
 #elif (defined SLJIT_CONFIG_UNSUPPORTED && SLJIT_CONFIG_UNSUPPORTED)
 
--- a/sljit_src/sljitConfigInternal.h
+++ b/sljit_src/sljitConfigInternal.h
@@ -162,8 +162,6 @@ extern "C" {
 #define SLJIT_CONFIG_SPARC_32 1
 #elif defined(__tilegx__)
 #define SLJIT_CONFIG_TILEGX 1
-#elif defined(__s390x__)
-#define SLJIT_CONFIG_S390X 1
 #else
 /* Unsupported architecture */
 #define SLJIT_CONFIG_UNSUPPORTED 1
@@ -724,9 +722,30 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_sw sljit_exec_offset(void* ptr);
 
 #elif (defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X)
 
+/*
+ * https://refspecs.linuxbase.org/ELF/zSeries/lzsabi0_zSeries.html#STACKFRAME
+ *
+ * 160
+ *  .. FR6
+ *  .. FR4
+ *  .. FR2
+ * 128 FR0
+ * 120 R15 (used for SP)
+ * 112 R14
+ * 104 R13
+ *  96 R12
+ *  ..
+ *  48 R6
+ *  ..
+ *  16 R2
+ *   8 RESERVED
+ *   0 SP
+ */
+#define SLJIT_S390X_DEFAULT_STACK_FRAME_SIZE 160
+
 #define SLJIT_NUMBER_OF_REGISTERS 12
 #define SLJIT_NUMBER_OF_SAVED_REGISTERS 8
-#define SLJIT_LOCALS_OFFSET_BASE 160
+#define SLJIT_LOCALS_OFFSET_BASE SLJIT_S390X_DEFAULT_STACK_FRAME_SIZE
 
 #elif (defined SLJIT_CONFIG_UNSUPPORTED && SLJIT_CONFIG_UNSUPPORTED)
 
--- a/sljit_src/sljitLir.c
+++ b/sljit_src/sljitLir.c
@@ -1384,6 +1384,8 @@ static SLJIT_INLINE CHECK_RETURN_TYPE check_sljit_emit_op_custom(struct sljit_co
 #elif (defined SLJIT_CONFIG_ARM_THUMB2 && SLJIT_CONFIG_ARM_THUMB2)
 	CHECK_ARGUMENT((size == 2 && (((sljit_sw)instruction) & 0x1) == 0)
 		|| (size == 4 && (((sljit_sw)instruction) & 0x3) == 0));
+#elif (defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X)
+	CHECK_ARGUMENT(size == 2 || size == 4 || size == 6);
 #else
 	CHECK_ARGUMENT(size == 4 && (((sljit_sw)instruction) & 0x3) == 0);
 #endif
@@ -2119,6 +2121,8 @@ static SLJIT_INLINE sljit_s32 sljit_emit_cmov_generic(struct sljit_compiler *com
 #	include "sljitNativeSPARC_common.c"
 #elif (defined SLJIT_CONFIG_TILEGX && SLJIT_CONFIG_TILEGX)
 #	include "sljitNativeTILEGX_64.c"
+#elif (defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X)
+#	include "sljitNativeS390X.c"
 #endif
 
 #if !(defined SLJIT_CONFIG_MIPS && SLJIT_CONFIG_MIPS)
--- a/sljit_src/sljitLir.c
+++ b/sljit_src/sljitLir.c
@@ -633,7 +633,10 @@ static SLJIT_INLINE sljit_s32 get_arg_count(sljit_s32 arg_types)
 	return arg_count;
 }
 
-#if !(defined SLJIT_CONFIG_X86 && SLJIT_CONFIG_X86)
+
+/* Only used in RISC architectures where the instruction size is constant */
+#if !(defined SLJIT_CONFIG_X86 && SLJIT_CONFIG_X86) \
+	&& !(defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X)
 
 static SLJIT_INLINE sljit_uw compute_next_addr(struct sljit_label *label, struct sljit_jump *jump,
 	struct sljit_const *const_, struct sljit_put_label *put_label)
@@ -655,7 +658,7 @@ static SLJIT_INLINE sljit_uw compute_next_addr(struct sljit_label *label, struct
 	return result;
 }
 
-#endif /* !SLJIT_CONFIG_X86 */
+#endif /* !SLJIT_CONFIG_X86 && !SLJIT_CONFIG_S390X */
 
 static SLJIT_INLINE void set_emit_enter(struct sljit_compiler *compiler,
 	sljit_s32 options, sljit_s32 args, sljit_s32 scratches, sljit_s32 saveds,
--- a/sljit_src/sljitNativeS390X.c
+++ b/sljit_src/sljitNativeS390X.c
@@ -1,0 +1,2519 @@
+/*
+ *    Stack-less Just-In-Time compiler
+ *
+ *    Copyright Zoltan Herczeg (hzmester@freemail.hu). All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are
+ * permitted provided that the following conditions are met:
+ *
+ *   1. Redistributions of source code must retain the above copyright notice, this list of
+ *      conditions and the following disclaimer.
+ *
+ *   2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *      of conditions and the following disclaimer in the documentation and/or other materials
+ *      provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDER(S) AND CONTRIBUTORS ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER(S) OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR
+ * BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <sys/auxv.h>
+
+#ifdef __ARCH__
+#define ENABLE_STATIC_FACILITY_DETECTION 1
+#else
+#define ENABLE_STATIC_FACILITY_DETECTION 0
+#endif
+#define ENABLE_DYNAMIC_FACILITY_DETECTION 1
+
+SLJIT_API_FUNC_ATTRIBUTE const char* sljit_get_platform_name(void)
+{
+	return "s390x" SLJIT_CPUINFO;
+}
+
+/* Instructions. */
+typedef sljit_uw sljit_ins;
+
+/* Instruction tags (most significant halfword). */
+const sljit_ins sljit_ins_const = (sljit_ins)(1) << 48;
+
+/* General Purpose Registers [0-15]. */
+typedef sljit_uw sljit_gpr;
+const sljit_gpr r0 = 0; /* 0 in address calculations; reserved */
+const sljit_gpr r1 = 1; /* reserved */
+const sljit_gpr r2 = 2; /* 1st argument */
+const sljit_gpr r3 = 3; /* 2nd argument */
+const sljit_gpr r4 = 4; /* 3rd argument */
+const sljit_gpr r5 = 5; /* 4th argument */
+const sljit_gpr r6 = 6; /* 5th argument; 1st saved register */
+const sljit_gpr r7 = 7;
+const sljit_gpr r8 = 8;
+const sljit_gpr r9 = 9;
+const sljit_gpr r10 = 10;
+const sljit_gpr r11 = 11;
+const sljit_gpr r12 = 12;
+const sljit_gpr r13 = 13;
+const sljit_gpr r14 = 14; /* return address and flag register */
+const sljit_gpr r15 = 15; /* stack pointer */
+
+const sljit_gpr tmp0 = 0; /* r0 */
+const sljit_gpr tmp1 = 1; /* r1 */
+
+/* Link registers. The normal link register is r14, but since
+   we use that for flags we need to use r0 instead to do fast
+   calls so that flags are preserved. */
+const sljit_gpr link_r = 14;     /* r14 */
+const sljit_gpr fast_link_r = 0; /* r0 */
+
+/* Flag register layout:
+
+   0               32  33  34      36      64
+   +---------------+---+---+-------+-------+
+   |      ZERO     | 0 | 0 |  C C  |///////|
+   +---------------+---+---+-------+-------+
+*/
+const sljit_gpr flag_r = 14; /* r14 */
+
+static const sljit_gpr reg_map[SLJIT_NUMBER_OF_REGISTERS + 1] = {
+	2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 15
+};
+
+struct sljit_s390x_const {
+	struct sljit_const const_; /* must be first */
+	sljit_sw init_value;       /* required to build literal pool */
+};
+
+/* Convert SLJIT register to hardware register. */
+static sljit_gpr gpr(sljit_s32 r)
+{
+	SLJIT_ASSERT(r != SLJIT_UNUSED);
+	r -= SLJIT_R0; /* normalize */
+	SLJIT_ASSERT(r < (sizeof(reg_map) / sizeof(reg_map[0])));
+	return reg_map[r];
+}
+
+/* Size of instruction in bytes. Tags must already be cleared. */
+static sljit_uw sizeof_ins(sljit_ins ins)
+{
+	if (ins == 0) return 2; /* keep faulting instructions */
+	if ((ins&0x00000000ffff) == ins) return 2;
+	if ((ins&0x0000ffffffff) == ins) return 4;
+	if ((ins&0xffffffffffff) == ins) return 6;
+	abort();
+}
+
+static sljit_s32 push_inst(struct sljit_compiler *compiler, sljit_ins ins)
+{
+	sljit_ins *ibuf = (sljit_ins *)ensure_buf(compiler, sizeof(sljit_ins));
+	FAIL_IF(!ibuf);
+	*ibuf = ins;
+	compiler->size++;
+	return SLJIT_SUCCESS;
+}
+
+static sljit_s32 encode_inst(void **ptr, sljit_ins ins)
+{
+	sljit_u16 *ibuf = (sljit_u16 *)*ptr;
+	sljit_uw size = sizeof_ins(ins);
+	SLJIT_ASSERT((size&6) == size);
+	switch (size) {
+	case 6: *ibuf++ = (sljit_u16)(ins >> 32);
+	case 4: *ibuf++ = (sljit_u16)(ins >> 16);
+	case 2: *ibuf++ = (sljit_u16)(ins);
+	}
+	*ptr = (void*)ibuf;
+	return SLJIT_SUCCESS;
+}
+
+/* Map the given type to a 4-bit condition code mask. */
+static sljit_uw get_cc(sljit_s32 type) {
+	const sljit_uw eq = 1 << 3; /* equal {,to zero} */
+	const sljit_uw lt = 1 << 2; /* less than {,zero} */
+	const sljit_uw gt = 1 << 1; /* greater than {,zero} */
+	const sljit_uw ov = 1 << 0; /* {overflow,NaN} */
+	const sljit_uw mask = 0xf;
+
+	switch (type) {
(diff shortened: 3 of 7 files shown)


zherczeg:
Let me know when the code is ready for review. Considering the missing parts, this patch definitely not closes #89.

carenas:
> Let me know when the code is ready for review. Considering the missing parts, this patch definitely not closes #89.

it is ready for review for step 1, and while I agree it doesn't fully close #89, I was hoping to close at least the "redefined" objective on it so I can get the work for step 2 going

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2613 @@
...
+		return (__ARCH__ >= 11 /* z13 */);
+	case MISCELLANEOUS_INSTRUCTION_EXTENSIONS_2_FACILITY:
+	case VECTOR_ENHANCEMENTS_1_FACILITY:
+		return (__ARCH__ >= 12 /* z14 */);
+	}
+#endif
+	return 0;
+}
+
+static SLJIT_INLINE long get_hwcap()
```
zherczeg, line 233: What this function does?

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2613 @@
...
+/* Report whether the given facility is available. This function always
+   performs a runtime check. */
+static int have_facility_dynamic(facility_bit x)
+{
+#if ENABLE_DYNAMIC_FACILITY_DETECTION
+	static struct {
+		sljit_uw bits[4];
+	} cpu_features;
+	size_t size = sizeof(cpu_features);
+	const sljit_uw word_index = x / 64;
```
zherczeg, line 260: `>> 6`

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2613 @@
...
+	const sljit_uw bit_index = ((1UL << 63) >> (x & 63));
+
+	SLJIT_ASSERT(sizeof(cpu_features) >= ((x+7)/8));
```
zherczeg, line 263: Unusual syntax.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2613 @@
...
+	if (SLJIT_UNLIKELY(!have_stfle())) {
+		return 0;
+	}
```
zherczeg, line 266: No need a block here, but keep the newline.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2613 @@
...
+		have = have_facility_dynamic(bit) ? 1 : 0; \
+	} \
+	return SLJIT_LIKELY(have); \
+}
+HAVE_FACILITY(have_eimm,    EXTENDED_IMMEDIATE_FACILITY)
+HAVE_FACILITY(have_ldisp,   FAST_LONG_DISPLACEMENT_FACILITY)
+HAVE_FACILITY(have_genext,  GENERAL_INSTRUCTION_EXTENSION_FACILITY)
+HAVE_FACILITY(have_lscond1, LOAD_STORE_ON_CONDITION_1_FACILITY)
+HAVE_FACILITY(have_lscond2, LOAD_STORE_ON_CONDITION_2_FACILITY)
+HAVE_FACILITY(have_misc2,   MISCELLANEOUS_INSTRUCTION_EXTENSIONS_2_FACILITY)
```
zherczeg, line 301: Some extra newlines would make this looks nicer.

zherczeg (changes requested):
This patch is quite big, will take time to review (I will do it in multiple steps). I noticed a large amount of syntax errors, I would recommend to fix them in the meantime. Please also squash the commits.

.travis.yml
```diff
@@ -0,0 +1,38 @@
+language: c
```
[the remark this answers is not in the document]
carenas: while it is not needed; since there is no way for other developers to test their changes in s390x unless they use the IBM cloud it will be an important addition (as discussed in #89 ).  i'll remove it from the initial port PR and keep it on my own tree, but then it will be only effective after rebasing on top of your master and might be too late to prevent breaking changes being introduced.

it will required that you "install" travis into your github account though but there is no cost to do so if using their opensource cloud.

additionally it could help filter out PR that could otherwise break in architectures that you are not normally using like macOS and windows or some non x86 linux if extended further and with further tweaks, which I can assist with.

sljit_src/sljitConfigInternal.h
```diff
@@ -715,6 +720,12 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_sw sljit_exec_offset(void* ptr);
 #define SLJIT_NUMBER_OF_SAVED_REGISTERS 5
 #define SLJIT_LOCALS_OFFSET_BASE 0
 
+#elif (defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X)
+
+#define SLJIT_NUMBER_OF_REGISTERS 12
```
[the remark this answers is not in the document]
carenas: I was guessing  that is used for making sure the 2 "reserved" registers are not touched (indeed SLJIT_R0 == r2)

sljit_src/sljitLir.c
```diff
@@ -633,10 +633,12 @@ static SLJIT_INLINE sljit_s32 get_arg_count(sljit_s32 arg_types)
 	return arg_count;
 }
 
-#if !(defined SLJIT_CONFIG_X86 && SLJIT_CONFIG_X86)
+#if !(defined SLJIT_CONFIG_X86 && SLJIT_CONFIG_X86) \
+	&& !(defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X)
 
-static SLJIT_INLINE sljit_uw compute_next_addr(struct sljit_label *label, struct sljit_jump *jump,
-	struct sljit_const *const_, struct sljit_put_label *put_label)
+static SLJIT_INLINE sljit_uw compute_next_addr(struct sljit_label *label,
```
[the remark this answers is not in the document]
carenas: I was adding code near it, so I though a little reformatting to avoid really long lines wouldn't hurt

sljit_src/sljitLir.c
```diff
@@ -655,7 +657,7 @@ static SLJIT_INLINE sljit_uw compute_next_addr(struct sljit_label *label, struct
 	return result;
 }
 
-#endif /* !SLJIT_CONFIG_X86 */
+#endif /* !CISC */
```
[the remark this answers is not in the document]
carenas: would you prefer !RISC or CISC?; apologizes for getting it backwards, I was hoping it will be clearer then to whoever find this function later than it only make sense with RISC since CISC will also need the op to calculate the next address.

sljit_src/sljitNativeS390X.c
carenas: returns a 64bit bitmap with "capabilities" that can be used to determine what the CPU supports (ex: VECTOR operations)

the original code has several of those, that seem to have been done statically at compile time and also at runtime, probably to optimize them but also to provide wide enough support.

since we will be targeting z14, it is mostly noise (and indeed the coverage for the !have_genext() needs work) but it is better to keep it for now since it will be useful in step 3 and beyond (even if supporting QEMU is likely not to be possible)

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2613 @@
...
+	return reg_map[r];
+}
+
+/* Size of instruction in bytes. Tags must already be cleared. */
+static sljit_uw sizeof_ins(sljit_ins ins)
+{
+	if (ins == 0) return 2; /* keep faulting instructions */
+	if ((ins & 0x00000000ffff) == ins) return 2;
+	if ((ins & 0x0000ffffffff) == ins) return 4;
+	if ((ins & 0xffffffffffff) == ins) return 6;
```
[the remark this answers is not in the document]
carenas: no complains in s390x (neither gcc or clang), which also support __int128, so guess it still has a long way to go.

I added suffix to some cases that could benefit from it, but left the rest for later as I wanted to avoid unnecessary churn in the first change (which is big enough already)

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2613 @@
...
+	r -= SLJIT_R0; /* normalize */
+	SLJIT_ASSERT(r < (sljit_s32)(sizeof(reg_map) / sizeof(reg_map[0])));
+	return reg_map[r];
+}
+
+/* Size of instruction in bytes. Tags must already be cleared. */
+static sljit_uw sizeof_ins(sljit_ins ins)
+{
+	if (ins == 0) return 2; /* keep faulting instructions */
+	if ((ins & 0x00000000ffff) == ins) return 2;
```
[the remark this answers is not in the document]
carenas: It might be worth adopting though, as it just looks like a smarter switch and it is simpler than the alternative

sljit_src/sljitNativeS390X.c
carenas: nicer is subjective, but added a newline to make the macro standup more.

since this are macro generated functions, having them weird looking might be a feature IMHO

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2613 @@
...
+	2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 15
+};
+
+struct sljit_s390x_const {
+	struct sljit_const const_; /* must be first */
+	sljit_sw init_value;       /* required to build literal pool */
+};
+
+/* Convert SLJIT register to hardware register. */
+static sljit_gpr gpr(sljit_s32 r)
```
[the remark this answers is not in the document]
carenas: normally you wouldn't need it, but on this implementation it does normalization and it is therefore not only necessary but used where you would use reg_map normally.

I should had pointed out as another oddity, but while strange seems to work, which is why i punted changing it until now (it is also tricky to fix, and was hoping it didn't need fixing)

sljit_src/sljitConfigInternal.h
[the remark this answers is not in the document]
mundaym: Thanks for taking a look at this @carenas and @zherczeg, it's good to see it being worked on again. I am happy to answer questions. Feel free to tag me.

> I was guessing that is used for making sure the 2 "reserved" registers are not touched

Yes, I set the number of registers to 12 because some of the registers are 'reserved' to hold temporary values during codegen.

> same goes for the FPU avoidance

I think we just ignored it in our initial port because (at least at the time) PCRE2JIT didn't use the FPU and we never finished it.

> the use of the constant pool

The constant pool is probably overengineered. You could emit a LLILF-IIHF pair for each constant load. There are definitely some gotchas when modifying instructions like that though. In particular the effect on the performance of the instruction cache can be bad (they usually aren't heavily optimized for self modifying code - I think this is true on x86 too though).

sljit_src/sljitNativeS390X.c
mundaym: Just some background: the HWCAP bit vector gives you information about what hardware features the Linux kernel supports. This is important since STFLE might tell you that vector instructions are available on the machine but if the kernel does not advertise support for them then they can't be used (the kernel needs to save/restore vector registers during context switches amongst other things).

mundaym: > since we will be targeting z14, it is mostly noise (and indeed the coverage for the !have_genext() needs work IMHO)

Yeah, if targeting z14 and above then I'd just drop facility checks like `genext` (maybe do them at startup to give a nice error message on older machines). When initially starting the port I added the checks but as it became clear that it was not going to be easy to support older machines I started to leave them out.

zherczeg (changes requested):
I will continue the review later.

sljit_src/sljitConfigInternal.h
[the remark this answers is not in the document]
zherczeg: Hi @mundaym!
Thank you for joining. I forgot about the stack pointer, but usually 2 temporary registers are enough to implement al instructions. I hope at some point we will only use r0 (since it cannot be used for addressing) and r14 (link register).

On the fly code modification is not frequent, but they often worth it, so their cost should be negligible in the long run.

 I used constant pool on ARMv5, but I decided to never use it again. It make things way too complex.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2647 @@
...
+	if ((ins & 0xffffffffffffL) == ins)
+		 return 6;
+
+	SLJIT_UNREACHABLE();
+	return (sljit_uw)-1;
+}
+
+static sljit_s32 push_inst(struct sljit_compiler *compiler, sljit_ins ins)
+{
+	sljit_ins *ibuf = (sljit_ins *)ensure_buf(compiler, sizeof(sljit_ins));
```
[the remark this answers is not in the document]
carenas: 8 bytes are allocated for the instruction buffer, which then is written with the instructions, that then are copied (only the 2, 4 or 6 bytes) to the "code" ptr in generate_code(); I know it is wasteful but it simplifies this part not having to worry about alignment or endianess

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2647 @@
...
+		if ((sljit_uw)v == (v & 0x00000000ffffffffU))
+			return push_inst(compiler, llilf(target, (sljit_u32)v));
+		else if ((sljit_uw)v == (v & 0xffffffff00000000U))
+			return push_inst(compiler, llihf(target, (sljit_u32)(v>>32)));
+
+		FAIL_IF(push_inst(compiler, llilf(target, (sljit_u32)v)));
+		return push_inst(compiler, iihf(target, (sljit_u32)(v>>32)));
+	}
+	/* TODO(mundaym): instruction sequences that don't use extended immediates */
+	abort();
```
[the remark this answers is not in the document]
carenas: those abort() were left explicitly (together with a matching TODO) as the target for this codebase are still developers (also why is disabled for auto detection).

once step2 is completed, it will be mandatory because the code will only be enabled if targeting z14 (or most likely z13, since the Ubuntu instances from Travis can't build for anything newer than that), but leaving it there allows us to develop and test and implementation (once a developer removes the safeguards)

if the community fails to materialize, or there is no demand for it (which will require people finding this after they try to force it and crash) then we can remove it and make the "officially unsupported" claim and we should make that choice before this code is enabled for end users (with auto detection on) but I think it is premature now.

also I am no mainframe expert and while it is likely eimm should be available in all cases (since it comes with z10 or higher) there might be cases were it is not enabled (maybe needs a license, or is an option that can be turned off at the hypervisor) so crashing can be useful (with the constrains explained above) and removing it makes us lose visibility of places where the codebase might need improvement.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2647 @@
...
+static sljit_s32 push_load_imm_inst(struct sljit_compiler *compiler, sljit_gpr target, sljit_sw v)
+{
+	/* 4 byte instructions */
+	if (is_s16(v))
+		return push_inst(compiler, lghi(target, (sljit_s16)v));
+
+	if ((sljit_uw)v == (v & 0x000000000000ffffU))
+		return push_inst(compiler, llill(target, (sljit_u16)v));
+	else if ((sljit_uw)v == (v & 0x00000000ffff0000U))
+		return push_inst(compiler, llilh(target, (sljit_u16)(v>>16)));
```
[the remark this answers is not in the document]
carenas: Apologies, I obviously misunderstood how relevant these syntax issues were at this phase, but could you spell out what is the issue?

I did remove the {} from the original code (unneeded when only one instruction) and even added a suffix to the longs.  had found both "} else {" and "else {" in the codebase, so assumed the later for the next iteration, but that shouldn't be the problem here either.

added spaces around ">>" too, which was missing from the previous edit that added spaces around the "&", but it was because I was thinking (obviously incorrectly in hindsight) that it was better to keep working code than make sure it matches perfectly to our style (at least for this phase, and not out of laziness, but to avoid introducing bugs that will distract us from the objective)

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2647 @@
...
+	SLJIT_ASSERT(is_s20(d));
+	return dh | dl;
+}
+
+#define SLJIT_S390X_INSTRUCTION(op, ...) \
+static SLJIT_INLINE sljit_ins op(__VA_ARGS__)
+
+/* RR form instructions. */
+#define SLJIT_S390X_RR(name, pattern) \
+SLJIT_S390X_INSTRUCTION(name, sljit_gpr dst, sljit_gpr src) \
```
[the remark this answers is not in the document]
carenas: they are defined as static function (i fixed that on my first patchset) through the (non C89 compatible variadic macro) in line 333

https://github.com/carenas/sljit/blob/s390x/sljit_src/sljitNativeS390X.c#L334-L335

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2647 @@
...
+	const sljit_uw eq = 1 << 3; /* equal {,to zero} */
+	const sljit_uw lt = 1 << 2; /* less than {,zero} */
+	const sljit_uw gt = 1 << 1; /* greater than {,zero} */
+	const sljit_uw ov = 1 << 0; /* {overflow,NaN} */
+	const sljit_uw mask = 0xf;
+
+	switch (type) {
+	case SLJIT_EQUAL:
+	case SLJIT_EQUAL_F64:
+		return mask & eq;
```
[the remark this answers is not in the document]
carenas: mask is needed for when type is 0, so the result can be warranted (at least in SLJIT_NOT_EQUAL and SLJIT_NOT_OVERFLOW) to be 0xf and not (sljiw_uw)-1

the rest were probably added for consistency, but having the mask first in the & might not be standard

carenas:
Hopefully no more style issues; code still has plenty of TODOs and abort() but that is expected as it only meant to be a starting base for further development, with a focus of supporting enough JIT to allow PCRE to run and a target of developers (not final users)

 it does have enough support to run cleanly both SLJITs and PCREs tests in recent enough hardware though (z15 based VMs/containers from Linux ONE) using both gcc and clang, and RedHat 8 (compiled for z14) and two recent versions of Ubuntu (compiled for z13) though

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2781 @@
...
+
+/* ADD IMMEDIATE */
+SLJIT_S390X_RILA(afi,   0xc20900000000, sljit_s32)
+SLJIT_S390X_RILA(agfi,  0xc20800000000, sljit_s32)
+
+/* ADD IMMEDIATE HIGH */
+SLJIT_S390X_RILA(aih,   0xcc0800000000, sljit_s32) /* TODO(mundaym): high-word facility? */
+
+/* ADD LOGICAL IMMEDIATE */
+SLJIT_S390X_RILA(alfi,  0xc20b00000000, sljit_u32)
```
zherczeg, line 536: Is there warnings for these long constants?

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2781 @@
...
+#undef SLJIT_S390X_INSTRUCTION
+
+/* load condition code as needed to match type */
+static sljit_s32 push_load_cc(struct sljit_compiler *compiler, sljit_s32 type)
+{
+	type &= ~SLJIT_I32_OP;
+	switch (type) {
+	case SLJIT_ZERO:
+	case SLJIT_NOT_ZERO:
+		FAIL_IF(push_inst(compiler, cih(flag_r, 0)));
```
zherczeg, line 937: These can be shorten to `return push_inst(compiler, cih(flag_r, 0))`

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2781 @@
...
+
+/* RR form instructions. */
+#define SLJIT_S390X_RR(name, pattern) \
+SLJIT_S390X_INSTRUCTION(name, sljit_gpr dst, sljit_gpr src) \
+{ \
+	return (pattern) | ((dst & 0xf) << 4) | (src & 0xf); \
+}
+
+/* ADD */
+SLJIT_S390X_RR(ar,   0x1a00)
```
zherczeg, line 355: These definitions are strange to me, but lets keep it for now.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2781 @@
...
+		index = tmp;
+		off = 0; /* clear offset */
+	}
+	addr->base = base;
+	addr->index = index;
+	addr->offset = off;
+	return SLJIT_SUCCESS;
+}
+
+static sljit_s32 load_word(struct sljit_compiler *compiler, sljit_gpr dst,
```
zherczeg, line 1070: These function definitons are not consistent (see above). I would prefer less lines, similar to this definition.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2781 @@
...
+	sljit_ins ins;
+
+	SLJIT_ASSERT(src & SLJIT_MEM);
+	if (have_ldisp() || !is_32bit)
+		FAIL_IF(make_addr_bxy(compiler, &addr, src, srcw, tmp));
+	else
+		FAIL_IF(make_addr_bx(compiler, &addr, src, srcw, tmp));
+
+	if (is_32bit)
+		ins = is_u12(addr.offset) ?
```
zherczeg, line 1084: It would be good to invent something for these. The style is strange and it looks like these have the same format, so they could select the correct variant internally.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2781 @@
...
+	jump = compiler->jumps;
+	put_label = compiler->put_labels;
+
+	/* calculate the size of the code */
+	for (buf = compiler->buf; buf != NULL; buf = buf->next) {
+		sljit_uw len = buf->used_size / sizeof(sljit_ins);
+		sljit_ins *ibuf = (sljit_ins *)buf->memory;
+		for (i = 0; i < len; ++i, ++j) {
+			sljit_ins ins = ibuf[i];
+			if (ins & sljit_ins_const) {
```
zherczeg, line 1149: If we have 64 bit, where 16 bit is never used, we could use more flags to handle the cases below in a similar way. Unless we want to reduce the instructon space in the future.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2781 @@
...
+				jump = jump->next;
+			}
+			if (put_label && put_label->addr == j) {
+				pool_size += sizeof(*pool);
+				put_label = put_label->next;
+			}
+			ins_size += sizeof_ins(ins);
+		}
+		/* emit trailing label */
+		if (label && label->size == j) {
```
zherczeg, line 1176: Does this needs to done in every iteration instead of the end?

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2781 @@
...
+		label = label->next;
+	}
+
+	/* reset jumps */
+	jump = compiler->jumps;
+	put_label = compiler->put_labels;
+
+	/* emit the code */
+	j = 0;
+	for (buf = compiler->buf; buf != NULL; buf = buf->next) {
```
zherczeg, line 1214: The other implementations does not need two loops. Cannot we compute the size in a similar way to others? Furthermore it seems only one constant pool is created. What happens if the offsets are too long? Can be a security issue.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2781 @@
...
+	sljit_s32 args = get_arg_count(arg_types);
+	sljit_sw frame_size;
+
+	CHECK_ERROR();
+	CHECK(check_sljit_emit_enter(compiler, options, arg_types, scratches, saveds, fscratches, fsaveds, local_size));
+	set_emit_enter(compiler, options, arg_types, scratches, saveds, fscratches, fsaveds, local_size);
+
+	/* saved registers go in callee allocated save area */
+	compiler->local_size = (local_size + 0xf) & ~0xf;
+	frame_size = compiler->local_size + 160;
```
zherczeg, line 1351: There is a define constant for 160.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2783 @@
...
+			FAIL_IF(push_inst(compiler, la(r15, 0, tmp1, r15)));
+		}
+	}
+
+	if (args >= 1)
+		FAIL_IF(push_inst(compiler, lgr(gpr(SLJIT_S0), gpr(SLJIT_R0))));
+	if (args >= 2)
+		FAIL_IF(push_inst(compiler, lgr(gpr(SLJIT_S1), gpr(SLJIT_R1))));
+	if (args >= 3)
+		FAIL_IF(push_inst(compiler, lgr(gpr(SLJIT_S2), gpr(SLJIT_R2))));
```
zherczeg, line 1370: Is the fourth argument mapped to s3?

zherczeg (changes requested):
We are at halfway now!

sljit_src/sljitNativeS390X.c
carenas: no (and I was surprised as well), the C standard allows for literals to be assigned an integer type (depending on the architecture) up to unsigned long and at least in the gcc/clang that comes with RHEL8/s390x it automatically gets it to the right type without the need for suffixes or triggering warnings.

```
$ cat t.c
#include <stdio.h>

#define I	0xffffffff
#define L	0xffffffffffffffff

int main(void) {
  printf("%zu %zu\n", sizeof(I), sizeof(L));
  return 0;
}
$ gcc -Wall -Wextra -o t t.c
$ clang -Weverything -o t t.c
$ ./t
4 8
```

sljit_src/sljitNativeS390X.c
zherczeg: I have an idea. We could use a SELECT/WHEN or something similar macro (you choose the name you like):
`WHEN(is_u12(addr.offset), l, ly, addr)`

It seems the check is always is_u12, so this can be an option as well:
`WHEN_U12(addr, l, ly)`

sljit_src/sljitConfigInternal.h
```diff
@@ -715,6 +720,12 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_sw sljit_exec_offset(void* ptr);
...
+#define SLJIT_NUMBER_OF_SAVED_REGISTERS 8
+#define SLJIT_LOCALS_OFFSET_BASE 160
```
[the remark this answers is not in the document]
mundaym: The platform ABI specifies a 160 byte area allocated in the callers stack frame for callee-save registers. So when calling C functions up to 160 bytes of the callers stack frame might be clobbered:

```
    :     local     :
    :   variables   :
    |               |
    +---------------+ <- caller stack pointer + 160
    |               |
    |  callee-save  |
    |   registers   |
    |               |
    +---------------+ <- caller stack pointer
    |               |
    :  callee stack :
    :     frame     :
```

Having such a large callee-save area can be a bit wasteful but changing it would obviously cause compatability issues and in practice it works ok.

There is more detail about the (non-vector) calling convention here: http://legacy.redhat.com/pub/redhat/linux/7.1/es/os/s390x/doc/lzsabi0.pdf.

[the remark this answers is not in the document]
carenas: it is only needed if the function is not a leaf function (hence why I was expecting to get to it on step 3 with the other optimizations, as more of our functions should be leaf functions IMHO)

since sljit only supports up to 4 arguments (with the current implementation only managing 3) we are not going to need to go to the stack to get any parameters, but as @mundaym pointed out then our generated code won't be fully ABI compatible either (the part that worries me more, is the mention of allocating a frame with space for the back chain + some 8 bytes of reserved space which is both marked as optional and as required in pages 24-26 of the legacy ABI document that was shared)

sljit_src/sljitNativeS390X.c
carenas: Long term might be worth refactoring all these instructions to use a similar macro based approach as done in other architectures, but I had seen both approaches in different code generators and then I would rather focus on the missing parts for now.

sljit_src/sljitNativeS390X.c
carenas: the use of instruction tags instead of the linked lists was an oddity of the original code that I punted changing originally to avoid unnecessary refactoring, but as you pointed out might be worth considering as well in all other generators, as well as for all types in this one.

since we need max 6 bytes for all instructions, using 8 in general makes sense (unlike x86 where we could have up to 15); everything else is RISC so having one single instruction size typedef works like a charm.

then again I think that is probably not worth refactoring right now either IMHO, in retrospective I probably should had documented more verbosely those things like the original author did by adding TODOs, but I was of the mentality of removing them instead of adding more ;)

sljit_src/sljitNativeS390X.c
carenas: the two loops was another of those things that looked like worth fixing in step 3 (or later) instead of now, and I have to admit that by doing it in 2 loops is easier to understand for new developers IMHO.

the constant pool is usually created before the function code and there is one per function; you are correct this implementation is different but since they are 32bit wide shouldn't be a problem unless we have a JIT function bigger than 4GB

sljit_src/sljitNativeS390X.c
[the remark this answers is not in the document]
carenas: added as TODO for later as it will require undef of this macro on each redefinition and we might have a different solution long term (ex: non function macros)

note though that this code will only be built with gcc/clang and both seem to support this feature on this platform, so strict c89 compliance might not be a requirement IMHO (most other issues pointed out from pedantic were cleaned in the first batch)

sljit_src/sljitNativeS390X.c
carenas: yes, through the reg_map; but not sure if more is needed as I had to admit I just found out about this fourth argument from your answer to #49 

references to SLJIT_S3 and SLJIT_S4 in the testsuite seem to work fine so I am hoping it is, unless there is a deficiency there

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2817 @@
...
+				ins = llc(reg, mem.offset, mem.index, mem.base);
+				break;
+			case SLJIT_MOV32_S8:
+				ins = lb(reg, mem.offset, mem.index, mem.base);
+				break;
+			case SLJIT_MOV32_U16:
+				ins = llh(reg, mem.offset, mem.index, mem.base);
+				break;
+			case SLJIT_MOV32_S16:
+				ins = is_u12(mem.offset) ?
```
zherczeg, line 1647: I would like to use the WHEN macro everywhere, not for only two cases above. I saw a lot of similar constructs below.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2817 @@
...
+		ins |= (sljit_ins)(src[2]) << 8;
+		ins |= (sljit_ins)(src[3]);
+		break;
+	case 6:
+		ins |= (sljit_ins)(src[0]) << 40;
+		ins |= (sljit_ins)(src[1]) << 32;
+		ins |= (sljit_ins)(src[2]) << 24;
+		ins |= (sljit_ins)(src[3]) << 16;
+		ins |= (sljit_ins)(src[4]) << 8;
+		ins |= (sljit_ins)(src[5]);
```
zherczeg, line 2478: What about using an `memcpy((sljit_u8*)&inst + sizeof(sljit_ins) - size, instruction, size);` instead of the switch statement? This should work because the system is big endian, and the inst is initialized to 0.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2817 @@
...
+/* TODO(carenas): r12 might conflict in PIC code, reserve? */
+/* TODO(carenas): r13 is usually ponted to "pool" per ABI, using a tmp
+ *                like we do know might be faster though, reserve?
+ */
+
+/* TODO(carenas): should be named TMP_REG[1-2] for consistency */
+#define tmp0	r0
+#define tmp1	r1
+
+/* TODO(carenas): flags should move, there is also a[2-15] available */
```
zherczeg, line 90: Yes, this is extremely important. Most cpus predict inidirect branch to link register as "return", and they use a special stack to predict the target address of these. For other indirect jumps, they will not use this stack which has a high performance costs.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2817 @@
...
+		src_r = FAST_IS_REG(src) ? gpr(src) : tmp1;
+		if (src & SLJIT_MEM)
+			FAIL_IF(load_word(compiler, tmp1, src, srcw, tmp1, 0));
+
+		return push_inst(compiler, br(src_r));
+	case SLJIT_SKIP_FRAMES_BEFORE_FAST_RETURN:
+	case SLJIT_PREFETCH_L1:
+	case SLJIT_PREFETCH_L2:
+	case SLJIT_PREFETCH_L3:
+	case SLJIT_PREFETCH_ONCE:
```
zherczeg, line 2432: Are there no prefect instructions in s390x or they have just been not implemented yet? If the latter, please add a todo comment.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2817 @@
...
+		/* TODO(mundaym): distinct operand facility where needed */
+		if (src1_r != dst_r && src1_r != tmp0) {
+			FAIL_IF(push_inst(compiler, (op & SLJIT_I32_OP) ?
+				lr(tmp0, src1_r) : lgr(tmp0, src1_r)));
+			src1_r = tmp0;
+		}
+		sljit_ins ins = 0;
+		switch (GET_OPCODE(op) | (op & SLJIT_I32_OP)) {
+		/* 64-bit ops */
+		case SLJIT_ADD: ins = signed_flags ?
```
zherczeg, line 2328: Incorrect syntax. There are other similar cases below. Maybe a similar macro to WHEN could make these nice.

sljit_src/sljitNativeS390X.c
```diff
@@ -0,0 +1,2817 @@
...
+		else
+			FAIL_IF(make_addr_bxy(compiler, &mem, src2, src2w, tmp1));
+
+		int can_u12 = is_u12(mem.offset) ? 1 : 0;
+		sljit_ins ins = 0;
+		switch (GET_OPCODE(op) | (op & SLJIT_I32_OP)) {
+		/* 64-bit ops */
+		#define EVAL(op) op(dst_r, mem.offset, mem.index, mem.base)
+		case SLJIT_ADD:
+			ins = signed_flags ? EVAL(ag) : EVAL(alg);
```
zherczeg, line 2251: Could this be `WHEN(signed_flags, ag, alg, mem)` ?

zherczeg (changes requested):
I saw other issues, but I will try to do those in follow-up patches.

sljit_src/sljitNativeS390X.c
mundaym: I recommend deleting the comment about `a[2-15]` being available. Access registers aren't really suitable for flags since there aren't any instructions to set the condition code or branch based on their value. Moving values between access registers and general purpose registers is also likely to be a multi-cycle operation (the scheduling information in LLVM is a good source of estimates for this sort of thing: https://github.com/llvm/llvm-project/blob/819c1651f723b51490eb98b3fc11cbd7dafc7831/llvm/lib/Target/SystemZ/SystemZScheduleZ14.td#L699).

sljit_src/sljitNativeS390X.c
carenas: not sure if nicer, so suggest different names or scope if not

sljit_src/sljitNativeS390X.c
carenas: yes, the comment wasn't about moving the flags there but on probably using them if we run out of registers for other things.

I have an old [branch](/carenas/sljit/tree/s390x-old-broken) that was using r13 for it, but I couldn't get it not to crash (after a very long night and too many wild chases) and couldn't figure out why either since the generated code looked sane (considering the different registers used, of course) and abandoned.

my plan was to do that and use an "a" registers as last resort (but for regular use only) since we were running quickly out of registers once we took seriously all those "reserved" by the ABI, but remember that I didn't document the extra registers while changing that line ;)

sljit_src/sljitNativeS390X.c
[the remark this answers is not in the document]
edelsohn: I'm a little confused about this Travis comment because Sleef is testing in Travis and getting z15.  There only are z15 systems.  Maybe this is because the Travis yml configuration is requesting xenial instead of bionic.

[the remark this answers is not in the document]
edelsohn: It's a little confusing to have technical questions scattered through a pull request review.  Maybe the technical questions could be in the feature request issue itself to keep them localized?  I definitely can forward questions or ask mainframe colleagues to join to answer questions. And when the community starts to look at performance issues, the IBM Linux on Z team definitely can assist with performance analysis and advice.

zherczeg (approved):
There are still many improvements needed, but lets use this as first version. I will try to get an account for IBM cloud to work on them. I suggest to do the code quality improvements first before we add new features.

PR #97, opened 2020-11-03
Use the standard code path of sljit_emit_cmov on mips r6.

zherczeg:


ffontaine:
This PR fix the build failure on mips r6.

zherczeg:
Thank you.

PR #98, opened 2020-12-06
Fix initialization of the allocator_lock on Win32

dangelog:
Fix two problems in one go:

- The read of `allocator_lock` is not atomic, and races against the write
  a few lines below. The Interlocked API docs clearly state [1]:

  > "This function is atomic with respect to calls to other interlocked
  > functions."

  A plain read is *not* atomic (see below).

- There is no read barrier to acquire the memory pointed by `allocator_lock`;
  therefore, the write barrier inserted by the Interlocked API a few lines
  below does not synchronize with a non-NULL read. (A non-NULL read may
  not find the memory pointed by `allocator_lock` committed yet.)

There are two possible solutions, a MSVC-specific one and a generic one.
MSVC-specific would be to mark `allocator_lock` as volatile. volatile
has almost nothing to do with atomicity and synchronization, except on
MSVC on non-ARM, where it is actually atomic and synchronizes [2].

A generic one is to use the Interlocked API for the read part as well,
so that we're not tied to MSVC and non-ARM.

This commit implements this last one.

[1] https://docs.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchangepointer
[2] https://docs.microsoft.com/en-us/cpp/cpp/volatile-cpp?view=msvc-160&viewFallbackFrom=vs-2019#end-of-iso-compliant

### DIFF
--- a/sljit_src/sljitUtils.c
+++ b/sljit_src/sljitUtils.c
@@ -48,7 +48,7 @@ static HANDLE allocator_lock;
 static SLJIT_INLINE void allocator_grab_lock(void)
 {
 	HANDLE lock;
-	if (SLJIT_UNLIKELY(!allocator_lock)) {
+	if (SLJIT_UNLIKELY(!InterlockedCompareExchangePointer(&allocator_lock, NULL, NULL))) {
 		lock = CreateMutex(NULL, FALSE, NULL);
 		if (InterlockedCompareExchangePointer(&allocator_lock, lock, NULL))
 			CloseHandle(lock);


dangelog:
If you want to play with the generated code: https://gcc.godbolt.org/z/cPsn1e

zherczeg:
I am ok with this change.
@carenas what do you think?

carenas (approved):
Great Idea; while it will have a performance impact, this code shouldn't be in the hot path and will only be called multiple times in applications that are abusing it already.
it is also likely to make this code more robust, specially in windows arm that is not really well tested/supported by the current version.

Note that there is an implicit conversion to PVOID * that increases the alignment requirements for "allocator_lock" but hopefully that is handled transparently by all compilers used but it will probably also worth making it explicit to be 8byte aligned as it would seem MSVC still defaults to 4 per their documentation[1]

[1] https://docs.microsoft.com/en-us/cpp/cpp/align-cpp?view=msvc-160

dangelog:
Hi,

thanks for the feedback. Regarding:

> Note that there is an implicit conversion to PVOID * that increases the alignment requirements for "allocator_lock" but hopefully that is handled transparently by all compilers used but it will probably also worth making it explicit to be 8byte aligned as it would seem MSVC still defaults to 4 per their documentation[1]

Isn't the conversion there already for the existing call to InterlockedX? Also note that on x64 pointers (and `allocator_lock` is a HANDLE, hence a PVOID) are naturally aligned to 8 bytes https://docs.microsoft.com/en-us/cpp/build/x64-software-conventions?view=msvc-160#scalar-types

dangelog:
In fact, adding or removing alignas(8) from https://gcc.godbolt.org/z/fq3hs1 makes no difference...

dangelog:
In fact, adding or removing alignas(8) from https://gcc.godbolt.org/z/fq3hs1 makes no difference...

carenas:
> In fact, adding or removing alignas(8) from https://gcc.godbolt.org/z/fq3hs1 makes no difference...

alignas requires C++11 and we support C89, but you are correct that even adding the proprietary MSVC __declspec(align(8)) might not be needed, or if needed (and I might had misread the documentation) is an orthogonal bug.

note that per documentation it should be doing 4 byte alignment in 32-bit code, but haven't tested it other than guessing it will if the segment is aligned and because it allocates DD for the pointer per godbolt.

dangelog:
Do you mean that on MSVC there's the chance that a global isn't aligned to its natural alignment? Oh dear.

dangelog:
Well, feel free to push this; adding the declspec can always be done.

zherczeg:
Is the conclusion that this patch can be landed?

carenas:
correct

PR #102, opened 2021-02-15
freebsd: add support for PROT_MAX (since 13)

carenas:
starting with FreeBSD 13 an additional restriction on the way page
permissions are allowed to shift was added by indicating at mmap
time the maximum expected value using PROT_MAX.

indicate that the allocated page is also expected to have execute
permissions to avoid future silent failures at mprotect time.


zherczeg (approved):
LGTM

PR #103, opened 2021-02-15
util: allow get_page_alignment to work in pre POSIX systems

carenas:
likely to be useful in really old systems that predate POSIX and that
have at least SUSv2 support (not HP-UX)

avoid breaking in at least FreeBSD 13 when building with _POSIX_SOURCE


zherczeg (approved):
LGTM

Issue #99, opened 2021-02-09
Attempts to mprotect() with MAP_JIT failing on Apple Silicon as of macOS 11.2


logancollins:
We are using the PCRE2 regular expression library (version 10.36, and by extension, SLJIT), and have noticed a change in behavior after macOS 11.2 came out. Now, any attempt to allocate memory with SLJIT fails. It's okay for us, as PCRE2 can fall back to not using JIT for an expression. However, I wanted to dig in and figure out _why_ it was suddenly failing.

It turns out that, with `sljitExecAllocator.c`, the `mmap()` with `MAP_JIT` call is succeeding, but the subsequent call to `mprotect()` (on line 185) fails with a generic "permission denied" error from the system. After spending a day not really being able to tell "why", we filed a feedback issue with Apple DTS.

Apple DTS provided the following response:

> I think the documentation could state this more clearly, but I think the
> basic answer here is that this configuration:
> 
>   const int prot = PROT_READ | PROT_WRITE | PROT_EXEC;
> 
> ...is inherently invalid on Apple Silicon.  That is, it is inherently
> invalid to mark a page "rwx".  The documentation doesn't QUITE say that,
> but you can see the implications inside this article:
> 
> <https://developer.apple.com/documentation/apple_silicon/porting_just-in-time_compilers_to_apple_silicon?language=objc>
> 
> "When memory protection is enabled, a thread cannot write to a memory
> region and execute instructions in that region at the same time. Apple
> silicon enables memory protection for all apps, regardless of whether
> they adopt the Hardened Runtime."
> 
> "Each thread has its own set of access permissions for a given memory
> region. When you call mmap to create the memory region, the system
> initially configures it as readable and executable (R-X) for all
> threads. Calling pthread_jit_write_protect_np with the value false
> offers a secure and efficient way to simultaneously remove the
> executable permission and add the writable permission (RW-) for the
> current thread."
> 
> "Important
> Because pthread_jit_write_protect_np changes only the current thread’s
> permissions, avoid accessing the same memory region from multiple
> threads. Giving multiple threads access to the same memory region opens
> up a potential attack vector, in which one thread has write access and
> another has executable access to the same region."
> 
> If you read between the lines there, particularly with the last
> "important" block, the underlying issue here is that you simply CAN'T*
> have a "rwx" page on Apple Silicon.  I'm not sure what your mprotect
> call was doing pre-11.2, but I don't think it was behaving correctly.
> 
[...]

PR #105, opened 2021-02-21
macos: avoid mprotect checks since 11.2

carenas:
starting with macOS 11.2, the mprotect calls for RWX pages will fail at
least in Apple Silicon.

avoid the additional check that was originally added to workaround WX
implementations that lie about the permission of pages that were
allocated (ex: HardenedBSD) unless it has been specifically requested
(through a compile flag).

for macOS this allows the implementation provided for Apple Silicon in
hardware through pthread_jit_write_protect_np() for the versions that
had that support.

fixes: #99

### DIFF
--- a/sljit_src/sljitExecAllocator.c
+++ b/sljit_src/sljitExecAllocator.c
@@ -182,10 +182,12 @@ static SLJIT_INLINE void* alloc_chunk(sljit_uw size)
 	if (retval == MAP_FAILED)
 		return NULL;
 
+#ifdef SLIJT_WX_OS_NEEDSCHEK
 	if (mprotect(retval, size, prot) < 0) {
 		munmap(retval, size);
 		return NULL;
 	}
+#endif
 
 	SLJIT_UPDATE_WX_FLAGS(retval, (uint8_t *)retval + size, 0);
 


sljit_src/sljitExecAllocator.c
```diff
@@ -182,10 +182,12 @@ static SLJIT_INLINE void* alloc_chunk(sljit_uw size)
 	if (retval == MAP_FAILED)
 		return NULL;
 
+#ifdef SLIJT_WX_OS_NEEDSCHEK
```
carenas, line 185: HardenedBSD might be the only OS I am aware of that might have still an implementation of PAX that lies on their mmap() and need this, other than macOS (which also lies, but had already working code to flip permissions as needed since #90)

zherczeg:
This patch is in draft state for half a year, do you intend to work on this?

carenas:
> This patch is in draft state for half a year, do you intend to work on this?

yes, but have to admit I am still unsure of the right way to define it since it is common to several OS and their oddities

sljit_src/sljitExecAllocator.c
nikic: Just to suggest a possible alternative, the `PROT_READ | PROT_WRITE | PROT_EXEC` bit could be a macro that is set to `PROT_READ | PROT_EXEC` for M1, because those are the actually permissions for the initial mmap. Then the mprotect check should work fine, right?

(This is just based on my understanding of how this works, I have no system to test.)

> there are no OS especific checks where this could be set yet, neither I am sure if the flag could have a better name so pushing this as a draft to discuss further before merging.

Alternatively could make this an `#ifndef SLJIT_WX_DONT_CHECK` that is defined in the M1 branch.

PR #111, opened 2021-06-14
apple: restrict calling pthread_jit_write_protect_np to macOS >= 11

carenas:
as reported in BUG2769[1], it could result in a miscompiled binary
if using a Big Sur SDK but targeting and older version without the
call.

since macOS in Apple Silicon has a hard requirement for this API in
JIT pages, abort with an error for versions older than Big Sur.

while at it, add quotes around another #error call for consistency.

[1] https://bugs.exim.org/show_bug.cgi?id=2769


zherczeg (approved):
LGTM

PR #113, opened 2022-02-13
misc: build/documentation fixes from recent changes

carenas:
A hotpotch of changes that were required to build in recent macOS/linux and other related documentation/updates

PR #114, opened 2022-02-14
misc: fix -Wextra related errors

carenas:
Since d345a64 (misc: build/documentation fixes from recent changes
(#113), 2022-02-13), and as initially fixed in 92fcb0f (Fix fallthrough
errors., 2022-02-13); the build will be broken in modern versions of gcc
that will warn with -Wimplicit-fallthrough (included as part of -Wextra).

Make sure all architectures have the corresponding tags to avoid this
as required.

Additionally, avoid -Wunused-parameter errors from s390x from parameters
that weren't needed or are still pending implementation.

### DIFF
--- a/sljit_src/sljitNativeS390X.c
+++ b/sljit_src/sljitNativeS390X.c
@@ -194,6 +194,7 @@ static SLJIT_INLINE sljit_u8 get_cc(struct sljit_compiler *compiler, sljit_s32 t
 				return (cc0 | cc3);
 			return (cc0 | cc2);
 		}
+		/* fallthrough */
 
 	case SLJIT_EQUAL_F64:
 		return cc0;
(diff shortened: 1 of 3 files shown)


sljit_src/sljitNativeS390X.c
```diff
@@ -1243,8 +1247,9 @@ struct ins_forms {
 	sljit_ins op_g;
 };
 
-static sljit_s32 emit_commutative(struct sljit_compiler *compiler, const struct ins_forms *forms,
-	sljit_s32 dst, sljit_sw dstw,
+static sljit_s32 emit_commutative(struct sljit_compiler *compiler,
+	const struct ins_forms *forms,
```
zherczeg, line 1251: I would keep these in the previous line as it was before.

zherczeg (approved):
LGTM

PR #115, opened 2022-02-19
arm: Remove duplicated variable declaration

kiyolee:
For Windows, the variable `imm` is redeclared again inside conditional compile.


zherczeg:
This should be removed in 03469c8e1cf53a5fa4f0c096a2fe69244a8a9730
It is strange there is no conflicts here.

PR #117, opened 2022-03-07
sparc: avoid NOP implementation for clear_cache with gcc

carenas:
both issues introduced with 0863c54 (Support Android ARM64
cacheflush. Patch by Tavian Barnes., 2016-04-06) and might affect also PPC.

per gcc's bug(link in code), the builtin in clang (eventhough it might be only 64-bit) might be ok to use but don't have a way to test that and the fallback works fine otherwise.


zherczeg (approved):
LGTM

PR #118, opened 2022-03-13
power: avoid bogus clear_cache builtin in gcc >= 10

carenas:
Since 0863c54 (Support Android ARM64 cacheflush. Patch by Tavian Barnes.,
2016-04-06), the fallback to avoid using gcc's builtin has been masked if
the compiler also supported __has_builtin.

Shown breaking tests 11 and 12 in Debian 12 ppc64le

### DIFF
--- a/sljit_src/sljitConfigInternal.h
+++ b/sljit_src/sljitConfigInternal.h
@@ -324,18 +324,19 @@ extern "C" {
 /****************************/
 
 #if (!defined SLJIT_CACHE_FLUSH && defined __has_builtin)
-#if __has_builtin(__builtin___clear_cache)
+#if __has_builtin(__builtin___clear_cache) && !defined(__clang__)
 
 /*
  * https://gcc.gnu.org/bugzilla//show_bug.cgi?id=91248
- * clear_cache builtin in sparc's gcc is broken
+ * https://gcc.gnu.org/bugzilla//show_bug.cgi?id=93811
+ * gcc's clear_cache builtin for power and sparc are broken
  */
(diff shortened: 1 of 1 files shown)


sljit_src/sljitConfigInternal.h
```diff
@@ -324,18 +324,19 @@ extern "C" {
 /****************************/
 
 #if (!defined SLJIT_CACHE_FLUSH && defined __has_builtin)
-#if __has_builtin(__builtin___clear_cache)
+#if __has_builtin(__builtin___clear_cache) && !defined(__clang__)
```
zherczeg, line 327: I don't get the `!defined(__clang__)` check here. Is that function always wrong on clang?

carenas: gcc >= 10 will have both, but not clang.

clang's implementation of the builtin works fine (tested also in sparc, through NetBSD)

zherczeg: Ok, I misunderstood the check. The "will" means it has not been available yet? I mean the comment says `gcc >= 10` but  I don't see `__GNUC__ ` check.

carenas: __GNUC__ is also defined by clang and icc (at least the latest version that is based on clang and probably also in the older version), hence why checking for __has_builtin is a better indication.

you are correct though, that there might be  compiler that is not based in clang (which has a working builtin) and that also supports __has_builtin without being gcc; but I am willing to take that risk and bet their builtin also works (unlike gcc's)

zherczeg (approved):
LGTM

PR #119, opened 2022-03-28
s390x: be explicit about cache coherence needs

carenas:
Practically not needed, as the current code will otherwise fallthrough
to the gcc check below that sets the builtin (which does nothing) for
gcc < 10 (like Ubuntu 20.04 or RHEL 8) or never hit this for gcc >= 10.


zherczeg (approved):
LGTM

PR #121, opened 2022-04-06
macos: fix recent regression breaking build in macos/arm64e

carenas:
Since 75a5556 (power: avoid bogus clear_cache builtin in gcc >= 10 (#118),
2022-03-18), builds in macOS/arm64 will fail because of a missing function
definition for sys_icache_invalidate.

The reason why that code is suddenly excercised, is because before the
excemption added in that commit, clang will report its builtin and use
that instead.

Using the builtin in arm is problematic, because it could fail and the
builtin doesn't report failures[1] and so it is better to use the system
specific code in this case, but to do so the header needs to be added and
the parameters cast to the correct types.

As a sideeffect, the use of an APPLE specific API unbreaks the build in
iOS, so add a comment to take that into consideration when/if using the
builtin with clang is preferred later.

To avoid further regressions, reenable the use of the builtin once the
sparc and power fallbacks had been configured, and make sure that for the
case we are reaching into the android specific call, a header is also
present.

[1] https://gcc.gnu.org/bugzilla//show_bug.cgi?id=91053

### DIFF
--- a/sljit_src/sljitConfigInternal.h
+++ b/sljit_src/sljitConfigInternal.h
@@ -323,6 +323,16 @@ extern "C" {
 /* Instruction cache flush. */
 /****************************/
 
+/*
+ * TODO:
+ *
+ * clang >= 15 could be safe to enable below
+ * older versions are known to abort in some targets
+ * https://github.com/PhilipHazel/pcre2/issues/92
+ *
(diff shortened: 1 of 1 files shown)


zherczeg (approved):
LGTM

PR #123, opened 2022-04-10
build: add barebones cmake for Windows

carenas:
While installing GNU make is usually not that difficult, this addresses a specific pain point of Windows developers without IMHO making it more difficult to support the build in other systems.

As a sideeffect, the renaming of the Makefile also helps BSD users that might accidentally invoke BSD make instead of `gmake` and be greeted with a syntax error.


zherczeg:
Does this affect linux users?

carenas:
If by "this" you mean the default stack size being too small to run sljit_test, then no.  AFAIK only Windows has 1MB stacks.

Using a 4MB stack instead of the default would also work in macOS and Linux, but I intentionally decided to add the logic only to the "cmake" file so it will most likely only be used in Windows.

zherczeg:
I mean can you still build with "make" on Linux? The makefile is renamed.

carenas:
> I mean can you still build with "make" on Linux? The makefile is renamed.

yes, GNU make will use the new name automatically.

carenas:
> linux with musl libc (well known example is alpine linux) defaults to 128kb thread stacks. openbsd to 64kb even.

AFAIK OpenBSD has been using 512KB thread stack sizes for a few years already, but note that the restriction I was referring to wasn't per thread but per process, where mostly everyone (with 64-bit userspaces) provides at least 8MB by default.

zherczeg:
I can land this patch if it is ready. However, I have no plans to work on many build systems unless the change is trivial.

zherczeg (approved):
LGTM

PR #124, opened 2022-04-12
wxexec: drop suport for NetBSD < 8

carenas:
Fixing support for the EOL (mid 2020) 7.2 version (or even older) could be added with minimal changes (as shown in the first [iteration](/carenas/sljit/tree/oldnetbsd-support), but long term it is better to drop support and clean the codebase (as shown in the current patchset) even if the change is more intrusive.

### DIFF
--- a/sljit_src/sljitWXExecAllocator.c
+++ b/sljit_src/sljitWXExecAllocator.c
@@ -59,38 +59,15 @@
 #include <sys/mman.h>
 
 #ifdef __NetBSD__
-#if defined(PROT_MPROTECT)
-#define check_se_protected(ptr, size) (0)
 #define SLJIT_PROT_WX PROT_MPROTECT(PROT_EXEC)
-#else /* !PROT_MPROTECT */
-#ifdef _NETBSD_SOURCE
-#include <sys/param.h>
-#else /* !_NETBSD_SOURCE */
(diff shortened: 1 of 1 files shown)


carenas:
Alternatively, and as discussed in https://github.com/zherczeg/sljit/pull/67#pullrequestreview-459027540, we could instead accept it was broken and nobody needed it and remove all support for systems without PROT_MPROTECT.

FWIW the "SElinux" allocator is also broken on those systems, and I have to admit I wasn't planning on fixing it.

zherczeg:
Unfortunately it is hard to track me what is happening. It seems there are a lot of similar patches which affects the same OSes. Is there a valid way to support all of them?

carenas:
> Is there a valid way to support all of them?

The first iteration of this pull request had [code](https://github.com/carenas/sljit/commit/6cc6424e1e9f730cd8e69ef7a04cae6c56c23a6b) to do that, but as noted on the later comments, I think it is better instead to drop support as was done for the "selinux" allocator in #58.

It was broken since it was introduced and since nobody complained it is unlikely to be needed as was [argued](https://github.com/zherczeg/sljit/pull/67#pullrequestreview-459027540) by @krytarowski almost two years ago.

zherczeg (approved):
LGTM

PR #125, opened 2022-04-18
windows: allow building with MinGW64 gcc/clang

carenas:
Fix building with MinGW in Windows using either gcc or clang x86_64 target compilers.

Tested not to introduce regressions to MSVC builds with either Visual Studio or clang (CLangCL toolset) compilers.


zherczeg (approved):
LGTM

PR #126, opened 2022-04-19
windows: fix building in arm64 with clang

carenas:
Just like in the case of building for iOS, the __clear_cache()
implementation is missing (eventhough it might be for different
reasons)

Promote the documented system provided call and improve
the warning in the code to avoid future regressions like the
one introduced in 0863c5499c8ebd3c274b46442e1a0a9633ad42e2


zherczeg (approved):
LGTM

PR #127, opened 2022-04-22
macos: allow building with an os target older than 11.0

carenas:
Make the Apple Silicon code smarter, so it can allow binaries built for not existent OS versions, to easy the pain of "fat binary" builders that need to target older versions than Big Sur.

Since those OS versions are not available, it can be safe to assume that the code can't fail at runtime either, but it will need adjusting if support is added for non macOS or other toolchains than clang (with compiler_rt) as provided by Xcode.

Note that since the code generated will have a runtime dependency in the toolchain used to build it, it is inherently non portable and should be ideally linked statically with binaries using the same toolchain.  A warning to make that more obvious to users of this "feature" might be worth adding, but has been punted for this version as it was known to cause confusion instead in previous reports.


zherczeg (approved):
LGTM

PR #135, opened 2022-11-05
FreeBSD 13.1 and above w/x pages fix.

devnexen:
If kern.elf32/64.allow_wx sysctl settings is disabled then sljit cannot create pages with both w/x bits. However, procctl api allows to enable it per process.

### DIFF
--- a/sljit_src/sljitExecAllocator.c
+++ b/sljit_src/sljitExecAllocator.c
@@ -168,6 +168,25 @@ static SLJIT_INLINE void apple_update_wx_flags(sljit_s32 enable_exec)
 #define SLJIT_MAP_JIT	(0)
 #endif /* !SLJIT_MAP_JIT */
 
+#ifdef __FreeBSD__
+#include <sys/procctl.h>
+#include <limits.h>
+
+static SLJIT_INLINE int allow_wx_map(void)
+{
+#ifdef PROC_WXMAP_CTL
+	static int r = INT_MIN;
+	if (r == INT_MIN) {
+		int sljit_wx_enable = PROC_WX_MAPPINGS_PERMIT;
+		r = procctl(P_PID, 0, PROC_WXMAP_CTL, &sljit_wx_enable);
+	}
+	return r == 0 ? 0: -1;
+#else
+	return -1;
+#endif
+}
+#endif /* __FreeBSD__ */
+
 static SLJIT_INLINE void* alloc_chunk(sljit_uw size)
 {
 	void *retval;
@@ -189,8 +208,19 @@ static SLJIT_INLINE void* alloc_chunk(sljit_uw size)
 #endif /* MAP_ANON */
 
 	retval = mmap(NULL, size, prot, flags, fd, 0);
-	if (retval == MAP_FAILED)
+	if (retval == MAP_FAILED) {
+#ifdef __FreeBSD__
+		if (allow_wx_map() == 0) {
+			retval = mmap(NULL, size, prot, flags, fd, 0);
+			if (retval == MAP_FAILED)
+				return NULL;
+		} else {
+			return NULL;
+		}
+#else
 		return NULL;
+#endif
+	}
 
 #ifdef __FreeBSD__
         /* HardenedBSD's mmap lies, so check permissions again */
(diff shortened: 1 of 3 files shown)


sljit_src/sljitExecAllocator.c
```diff
@@ -168,6 +168,28 @@ static SLJIT_INLINE void apple_update_wx_flags(sljit_s32 enable_exec)
 #define SLJIT_MAP_JIT	(0)
 #endif /* !SLJIT_MAP_JIT */
 
+#ifdef __FreeBSD__
+#include <sys/param.h>
+#if __FreeBSD_version >= 1301000
```
carenas, line 173: using PROC_WXMAP_CTL directly would be preferably and avoid hardcoding a specific OS version

sljit_src/sljitExecAllocator.c
```diff
@@ -168,6 +168,28 @@ static SLJIT_INLINE void apple_update_wx_flags(sljit_s32 enable_exec)
...
+#include <sys/procctl.h>
+static SLJIT_INLINE int allow_wx_prot(void)
+{
+	static int r = INT_MIN;
+	if (r == INT_MIN) {
+		int sljit_wx_enable = PROC_WX_MAPPINGS_PERMIT;
+		r = procctl(P_PID, getpid(), PROC_WXMAP_CTL, &sljit_wx_enable);
```
carenas, line 180: use 0 instead of getpid()

carenas (changes requested):
IMHO, making these code FreeBSD specific might be better (there is already such an additional check because of the way HardenedBSD implements PAX)

it should be also probably reversed, since it is more likely that the WX restriction is not enabled, and even when it is, I would expect an admin using the ELF flag first to add an exception than relying in this magic.

zherczeg:
@devnexen do you plan to fix the requests above?

carenas:
the code is in better shape; it is still unclear what the motivation is to do the check ahead of time instead of a fallback.

isn't more reasonable (as it is done for the apple systems, see the logic for setting MAP_JIT), to ONLY do the extra work to enable WX maps if it will be denied?

As it is, and assuming that most systems won't have `sysctl kern.elf{32,64}=1`, the extra work is wasted, and even ignores the policy decision that someone might had set by adding an exception with `elfctl -e +wxneeded $BINARY`

devnexen:
The extra work is done once at worse.

sljit_src/sljitExecAllocator.c
```diff
@@ -168,17 +168,45 @@ static SLJIT_INLINE void apple_update_wx_flags(sljit_s32 enable_exec)
...
+	if (r == INT_MIN) {
+		int sljit_wx_enable = PROC_WX_MAPPINGS_PERMIT;
+		r = procctl(P_PID, 0, PROC_WXMAP_CTL, &sljit_wx_enable);
+	}
+	if (r == 0) {
+		return mmap(NULL, size, prot, flags, fd, 0);
+	}
+
+#endif
+	return NULL;
```
carenas, line 200: mixing map addresses and NULL is a bad idea, this function expands on mmap output and therefore shouldn't use NULL to flag an invalid map but MAP_FAILED

sljit_src/sljitExecAllocator.c
```diff
@@ -188,8 +216,7 @@ static SLJIT_INLINE void* alloc_chunk(sljit_uw size)
 	fd = dev_zero;
 #endif /* MAP_ANON */
 
-	retval = mmap(NULL, size, prot, flags, fd, 0);
-	if (retval == MAP_FAILED)
+	if (!(retval = sljit_map_wx(size, flags, fd)))
```
carenas, line 219: there is no need to refactor these, and indeed by doing so a bug was introduced

carenas (changes requested):
A proposed [fixup](devnexen/sljit#1) could be squashed on top of this for next iteration

sljit_src/sljitExecAllocator.c
```diff
@@ -168,17 +168,45 @@ static SLJIT_INLINE void apple_update_wx_flags(sljit_s32 enable_exec)
...
+#endif
+
+	ptr = mmap(NULL, size, prot, flags, fd, 0);
+	if (ptr != MAP_FAILED)
+		return ptr;
+#ifdef PROC_WXMAP_CTL
+	static int r = INT_MIN;
+	if (r == INT_MIN) {
+		int sljit_wx_enable = PROC_WX_MAPPINGS_PERMIT;
+		r = procctl(P_PID, 0, PROC_WXMAP_CTL, &sljit_wx_enable);
```
carenas, line 193: checking for the right kind of error from this function would be better, since AS-IS it would just trigger EINVAL and disable JIT in systems that don't understand PROX_WXMAP_CTL (a 13.1 binary running in older FreeBSD versions)

zherczeg:
Why the patch is called "Improve cmov on s390x "

PR #137, opened 2022-11-09
Support auto selection of an allocator

zherczeg:
@carenas what do you think of this patch?


minipli-oss:
Heh, nice. But it would still be nice if one could select which allocators will be included in a given build to address pcre2's concerns about the protexec allocator being "experimental".

Another nice enhancement would be if one (or sljit automatically) could pre-select an allocator to use, because while this change makes JIT available under grsec, it'll still create log entries for the failing attempts using the exec and wxexec allocators:

```
[188840.865516] grsec: denied RWX mmap of <anonymous mapping> by /home/minipli/src/sljit/bin/sljit_test[sljit_test:107091] uid/euid:1000/1000 gid/egid:1000/1000, parent /usr/bin/bash[bash:106778] uid/euid:1000/1000 gid/egid:1000/1000
[188840.865531] grsec: denied RWX mprotect of <anonymous mapping> by /home/minipli/src/sljit/bin/sljit_test[sljit_test:107091] uid/euid:1000/1000 gid/egid:1000/1000, parent /usr/bin/bash[bash:106778] uid/euid:1000/1000 gid/egid:1000/1000
```

Maybe make use of `sljit_get_runtime_support()` of #136 to initialize `selected_allocator` to `PROT_ALLOCATOR_SELECTED` if the function returns 0?

carenas:
IMHO protexec should almost never be used, after all it will segfault if the application forks.  wxexec might be a more suitable generic fallback even if I would argue than adding an exception in the policy so RWX maps are used instead should be the preferred option.

agree that building on top of #136 is a good idea, all checks that trigger those logs would be then done silently and would be consolidated.

zherczeg:
I have realized this is a bad idea. Allocators are not something which are dynamically selected, they should be tied to the target platform. Although more generic allocatoes are better, but ultimately the platform needs to decide which one to use, not the opposite direction.

PR #139, opened 2022-11-27
build failure at least with clang in 64bit arm

carenas:
Likely some missing code during recent refactorings and most likely to affect macOS in Apple Silicon


zherczeg (approved):
LGTM

PR #140, opened 2022-11-29
riscv: avoid SIGILL in 64bit while jumping

carenas:
breaking test15 consistently when building against both available riscv nodes in the GCC compiler farm[1]

[1] https://cfarm.tetaneutral.net/machines/list/#


zherczeg (approved):
LGTM

zherczeg:
Nice catch! Are these machines freely available for testing code?

carenas:
> Are these machines freely available for testing code?

need to request for an account first, but they are meant for opensource development :

  https://cfarm.tetaneutral.net/users/new/

PR #143, opened 2022-11-30
config: remove SLJIT_EXEC_OFFSET from list of required macros for an external allocator

carenas:
Since it is only required as part of "protexec", it shouldn't need to be provided also by external allocators.

Expand the fallback definition, so that the only two macros that are required are the ones that point to the external functions to allocate and free executable memory.

Keep using one if it is already provided for backward compatibility.

### DIFF
--- a/sljit_src/sljitConfig.h
+++ b/sljit_src/sljitConfig.h
@@ -96,7 +96,9 @@ extern "C" {
 
 /* Executable code allocation:
    If SLJIT_EXECUTABLE_ALLOCATOR is not defined, the application should
-   define SLJIT_MALLOC_EXEC, SLJIT_FREE_EXEC, and SLJIT_EXEC_OFFSET. */
+   define SLJIT_MALLOC_EXEC and SLJIT_FREE_EXEC.
+   Optionally, depending on the implementation used for the allocator,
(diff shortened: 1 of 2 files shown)


sljit_src/sljitConfig.h
```diff
@@ -96,7 +96,7 @@ extern "C" {
 
 /* Executable code allocation:
    If SLJIT_EXECUTABLE_ALLOCATOR is not defined, the application should
-   define SLJIT_MALLOC_EXEC, SLJIT_FREE_EXEC, and SLJIT_EXEC_OFFSET. */
+   define SLJIT_MALLOC_EXEC and SLJIT_FREE_EXEC. */
```
zherczeg, line 99: We should describe that there is an optional `SLJIT_EXEC_OFFSET` macro, since it might be needed in some cases.

carenas: the same applies to SLJIT_UPDATE_WX_FLAGS, but since it was not mentioned and had already a default fallback assumed these only made the formal interface cleaner.

zherczeg (approved):
LGTM

We could think about more consistent names, such as everything starts with SLJIT_EXEC_ or similar prefix.

PR #144, opened 2022-12-09
arm: add support for detecting NEON (WIP)

carenas:
Probably worth some refactoring to make it more generic and apply as well to s390x, or more extensible considering there are now multiple vector instruction sets per architecture with the most recent ones being variable width.

Is there any value on haviing implementation speciific names?, wouldn't it better to use SLJIT_HW_SIMD?

the arm code for SIMD in PCRE2 has a lot of issues, but one of the advantages it had was that it was less verbose and offloaded some of the hand coding by using the compiler intrinsics (but should be instead using sljit intrinsics), why not add them?


zherczeg:
These days vector seems more popular than simd. And should be generic of course. Sljit does not emit any vector instructions at the moment.

sljit_src/sljitNativeARM_32.c
```diff
@@ -41,6 +41,21 @@ SLJIT_API_FUNC_ATTRIBUTE const char* sljit_get_platform_name(void)
 #endif
 }
 
+#if (defined SLJIT_DETECT_NEON && SLJIT_DETECT_NEON)
+#define CPU_FEATURE_NEON	(1UL << 12)
+
+#include <sys/auxv.h>
```
zherczeg, line 47: Is it available in al compilers?

carenas: it is provided by the OS and available in most ELF systems.

this code is linux specific though and will need more care before being ready, this was mainly posted for testing and discussion

sljit_src/sljitConfigInternal.h
```diff
@@ -580,10 +580,12 @@ typedef double sljit_f64;
 
 #endif /* !SLJIT_FPU_UNALIGNED */
 
-#if (defined SLJIT_CONFIG_X86_32 && SLJIT_CONFIG_X86_32)
-/* Auto detect SSE2 support using CPUID.
-   On 64 bit x86 cpus, sse2 must be present. */
-#define SLJIT_DETECT_SSE2 1
+#if (defined SLJIT_CONFIG_X86_32 && SLJIT_CONFIG_X86_32) || \
+	(defined SLJIT_CONFIG_ARM_V7 && SLJIT_CONFIG_ARM_V7)
+/* Auto detect availability of SSE2 (using CPUID) or NEON.
```
carenas, line 585: It might be worth doing the checks here to avoid adding the "detect" support for cases that are known not to require it (ex: OpenBSD or macOS)

sljit_src/sljitNativeARM_32.c
```diff
@@ -41,6 +41,69 @@ SLJIT_API_FUNC_ATTRIBUTE const char* sljit_get_platform_name(void)
...
 
+#define CPU_FEATURE_NEON	(1UL << 12)
+static unsigned long cpu_feature_list;
+
+#if (defined SLJIT_DETECT_SIMD && SLJIT_DETECT_SIMD)
+#if defined(HAVE_GETAUXVAL) || defined(HAVE_ELF_AUX_INFO)
+#include <sys/auxv.h>
+#endif
+
+#ifdef __NetBSD__
```
carenas, line 52: might also apply to FreeBSD, specially when !HAVE_ELF_AUX_INFO but hadn't been able to test that yet

sljit_src/sljitNativeARM_32.c
```diff
@@ -41,6 +41,69 @@ SLJIT_API_FUNC_ATTRIBUTE const char* sljit_get_platform_name(void)
...
+#include <sys/sysctl.h>
+#endif
+
+static void get_cpu_features(void)
+{
+	if (cpu_feature_list)
+		return;
+
+#if defined(__ARM_ARCH) && __ARM_ARCH == 8
+	/* TODO: confirm if optional with armv9 */
```
carenas, line 63: if true (because NEON is being "replaced" by SVE2, then this might also apply to the 64 bit version.
FWIW some of these code might need to move regardless to a _common file to also apply to T2

test_src/sljitTest.c
```diff
@@ -10743,7 +10751,16 @@ int sljit_test(int argc, char* argv[])
...
 		printf(COLOR_RED "%d" COLOR_DEFAULT " (" COLOR_RED "%d%%" COLOR_DEFAULT ") tests are " COLOR_RED "FAILED" COLOR_DEFAULT " ", TEST_COUNT - successful_tests, (TEST_COUNT - successful_tests) * 100 / TEST_COUNT);
-	printf("on " COLOR_ARCH "%s" COLOR_DEFAULT "%s\n", sljit_get_platform_name(), sljit_has_cpu_feature(SLJIT_HAS_FPU) ? " (with fpu)" : " (without fpu)");
+
+	fpu = sljit_has_cpu_feature(SLJIT_HAS_FPU);
+	if (simd && fpu)
+		strcpy(features, " (with: fpu, simd)");
+	else if (fpu)
+		strcpy(features, " (with fpu)");
+	else
+		strcpy(features, " (without fpu)");
```
carenas, line 10761: this might misrepreset the case where (!fpu && simd), but that is by design as that combination (while possible) is not something that is available in the market yet.

SLJIT_IS_FPU_AVAILABLE and its usecases is also missing and worth discussing, but IMHO might be able to wait until SIMD support is added to sljit proper.

PR #145, opened 2022-12-12
config: improve autodetection for ARM

carenas:
Avoid misdetection of armv8 (or newer) systems by using a more standard and forward compatible macro of __ARM_ARCH.

Not sure exactly which release it was introduced with, but the gcc commit that did is from May 2012 and clang seems even older.

tested with gcc 5.5 (from NetBSD 8.2) and gcc 4.8.2 and clang 3.6 (from Ubuntu 14.04)

### DIFF
--- a/sljit_src/sljitConfigInternal.h
+++ b/sljit_src/sljitConfigInternal.h
@@ -141,16 +141,14 @@ extern "C" {
 #define SLJIT_CONFIG_X86_32 1
 #elif defined(__x86_64__)
 #define SLJIT_CONFIG_X86_64 1
-#elif defined(__arm__) || defined(__ARM__)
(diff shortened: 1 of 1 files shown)


carenas:
The new macro is produced by both clang and gcc (tested in Debian 11 with versions  11 and 10, respectivelly) and defined by the ACLE[1] standard,  while the older one is NOT defined when the target is armv8 (which is 10 years old) or higher and was incomplete (at least missing M, EM and S types).

Will add them back for backward compatibility, expand my testing and improve the commit message

[1] https://developer.arm.com/documentation/101028/0005/Feature-test-macros

zherczeg (approved):
LGTM

PR #147, opened 2022-12-18
riscv: allow disabling the FPU on demand or if soft-float ABI is requested

carenas:
Mainly deal with the fact that the FPU was always enabled, even though is optional, and therefore it would fail if running the tests in a soft-float system.

Most OS target lp64d at a minimum though and most commercially available CPUs seem to include an FPU and the extensions to use it on its ISA, so no "distribution" is affected.

### DIFF
--- a/sljit_src/sljitNativeRISCV_common.c
+++ b/sljit_src/sljitNativeRISCV_common.c
@@ -531,6 +531,13 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 {
 	switch (feature_type) {
 	case SLJIT_HAS_FPU:
+#ifdef SLJIT_IS_FPU_AVAILABLE
+		return SLJIT_IS_FPU_AVAILABLE;
+#elif defined(__riscv_float_abi_soft)
+		return 0;
+#else
+		return 1;
+#endif
 	case SLJIT_HAS_ZERO_REGISTER:
 		return 1;
 	default:


zherczeg:
Is there a way to detect it? Some kind of cpuid instruction?

zherczeg:
Btw no-fpu is not the same as soft-float, we just don't use fpu in this case. Soft float on arm for example means you pass values in core registers even though you can use fpu instructions to process them.

carenas:
> Is there a way to detect it? Some kind of cpuid instruction?

It is a compile time setting that is controlled by the -march parameter.  Each letter indicates an optional feature that is part of the ISA the CPU supports (ex: rv64i only hast the integer feature, rv64imafd or rv64g include also the 32-bit (f) and 64-bit (d) instructions with the width of the register defined with __riscv_flen.

you could also have (as indicated in the commit message) an rv32if march with rv32f abi which would then use 32bit wide float registers (__riscv_flen=32, __riscv_float_abi_single) but luckily that doesn't seem to exist yet, or at least not supported by the opensource OS/compiler/libc and instead rv{32,64}g+ ISA (with both sets of instructions and 64-bit wide registers) is instead the norm.

carenas:
> Soft float on arm for example means you pass values in core registers even though you can use fpu instructions to process them.

this is also the case with RISCV (given the combinations that seem to be supported), a good explanation of it in:

  https://www.sifive.com/blog/all-aboard-part-1-compiler-args

zherczeg:
So the commit description is incorrect, the patch simply detects fpu, abi is not affected. I am unhappy we need the arm32 like soft/hard abi thing on riscv as well.

zherczeg:
Is this patch ready?

PR #148, opened 2022-12-20
FreeBSD 13.1 and above w/x pages fix.

devnexen:
If kern.elf3264.allow_wx sysctl settings is disabled then sljit cannot create w/x pages.


carenas (approved):
naming and dependencies could be improved (as shown by proposed code that was discarded) upon, though

carenas:
an updated non conflicting alternative in the following [branch](https://github.com/carenas/sljit/tree/mem_freebsd)

zherczeg:
Is this still a valid patch?

PR #149, opened 2022-12-21
test: update test15 for 4 arguments

carenas:


zherczeg:
What is the motivation of this patch? When I added the 4th arg support, I have added other tests for them. Is this better this way?

carenas:
The use of a scratch register that was consecutive to the ones used to give parameters to the function called got me confused.

Still, adding a test with 4 parameters, and one that uses a different register and shows they are not overlapping is useful IMHO.

The typo fix is at least worth something.

zherczeg:
The patch was good, and the current version is even better. I was just curious about the motivation. I can land it if you are ready.

zherczeg:
Forgot to land this. Sorry

PR #150, opened 2022-12-23
Move allocators to separate files

zherczeg:
@carenas please check this code. The aim is reducing the ifdef hell in the allocators by moving them to separate files. Very close allocators can be merged together, but large differences should not.


zherczeg:
Tested on Linux and Windows, but there are likely issues on others.

PR #151, opened 2022-12-26
Move allocators to separate files

carenas:
Consolidated changes to refactor the allocator files into mostly independent files per supported platform


zherczeg:
Are you ok with the move patch? Do you want to merge it with this patch before landing (I can close my pr and let you do the whole thing)?

carenas:
I don't think the original patch is ready for merging, even after squashing this change on top, which is needed to fix building in macOS (only test it with Intel) and avoid crashes in 32bit NetBSD with ProtExec (should likely be also broken in Linux and other 32bit *NIX but didn't test it yet).

zherczeg:
It is very likely there will be issues, so it will take time to fix everything. Hopefully we will get feedback from users as well. I never thought the allocators will be this complex, and we need to do something with the growing complexity at some point.

zherczeg:
The jit te4sts works on 32 bit linux with prot exec allocator. Btw CHUNK_SIZE is configurable now, so Prot allocators could use bigger files than 64KB, something like 1MB should be better.

zherczeg:
Why do we want to combine netbsd and linux? The point is separating OS-es for easier maintainability.
What are the plans for this work? Shall we land them as independent patches, or you could combine the two patches in this pr?

carenas:
I tried to get a [branch](https://github.com/carenas/sljit/tree/int) rebased and integrated with all changes, but it is not working.

Will keep looking at it and retarget in this PR then when it is ready, feel free to close your original PR

zherczeg:
Ok. You can make a single patch from it (combining the original and your work). I hope you can do it soon.

zherczeg (approved):
LGTM

PR #153, opened 2023-01-04
config: really improve autodetection for arm

carenas:
Reformat previously added line for readability, add armv9 and most importantly make sure all parenthesis are paired and there are no typos.

### DIFF
--- a/sljit_src/sljitConfigInternal.h
+++ b/sljit_src/sljitConfigInternal.h
@@ -134,7 +134,6 @@ extern "C" {
 /********************************************************/
 
 #if (defined SLJIT_CONFIG_AUTO && SLJIT_CONFIG_AUTO)
-
 #ifndef _WIN32
 
 #if defined(__i386__) || defined(__i386)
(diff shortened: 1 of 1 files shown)


sljit_src/sljitConfigInternal.h
```diff
@@ -145,7 +144,10 @@ extern "C" {
 #define SLJIT_CONFIG_ARM_64 1
 #elif defined(__thumb2__)
 #define SLJIT_CONFIG_ARM_THUMB2 1
-#elif (defined(__ARM_ARCH) && __ARM_ARCH >= 7) || ((defined(__ARM_ARCH_7__) || defined(__ARM_ARCH_7A__) || defined(__ARM_ARCH_7R__) || defined(__ARM_ARCH_7S__)) || (defined(__ARM_ARCH_8A__) || defined_(__ARM_ARCH_8R__))
+#elif (defined(__ARM_ARCH) && __ARM_ARCH >= 7) || \
```
zherczeg, line 147: I am not sure, but I think you mentioned that this macro is always defined. Do we need the rest, or it is just there because of the old code?

carenas: The new macro should be always defined for compilers that care about ACE (which includes gcc and clang), but sadly not everyone does (ex: tinyc).

Keeping the old macro for backward compatibility (and when properly maintained) helps with that.

zherczeg: Ok. It is just a macro.

zherczeg (approved):
LGTM

PR #154, opened 2023-01-08
Fix Windows build

kiyolee:
Windows does not have `<sys/mman.h>`.


zherczeg:
This file should not have any includes. cc @carenas

carenas:
my bad; those were meant to go in the _posix file and indeed I added them there later when splitting the FreeBSD code which hasn't yet landed

PR #158, opened 2023-04-03
improve atomic instructions support

carenas:
Use an ISO C standard type and avoid failing tests for the unsupported ARMv5 platform.

Missing support for other architectures that affects building punted.


zherczeg:
Sorry for the extra work, I had the intention of improving the original patch

PR #159, opened 2023-04-17
Use unsigned to represent memory flags size

JetXujing:
Left shift of WORD_SIZE and INT_SIZE by 30 places cannot be represented in type 'int'. Using 'unsigned int' instead.

### DIFF
--- a/sljit_src/sljitNativeARM_64.c
+++ b/sljit_src/sljitNativeARM_64.c
@@ -937,14 +937,14 @@ set_flags:
 #define STORE		0x10
 #define SIGNED		0x20
 
-#define BYTE_SIZE	0x0
-#define HALF_SIZE	0x1
-#define INT_SIZE	0x2
-#define WORD_SIZE	0x3
+#define BYTE_SIZE	0x0u
+#define HALF_SIZE	0x1u
+#define INT_SIZE	0x2u
+#define WORD_SIZE	0x3u
 
 #define MEM_SIZE_SHIFT(flags) ((sljit_ins)(flags) & 0x3)
 
-static sljit_s32 emit_op_mem(struct sljit_compiler *compiler, sljit_s32 flags, sljit_s32 reg,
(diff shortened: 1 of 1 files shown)


JetXujing:
Yes, I agree with you and I have made the relevant changes. In addition, the input parameters of some functions and variable types are adjusted to avoid implicit conversion.

sljit_src/sljitNativeARM_64.c
```diff
@@ -1702,7 +1704,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_op_custom(struct sljit_compiler *c
 /*  Floating point operators                                             */
 /* --------------------------------------------------------------------- */
 
-static sljit_s32 emit_fop_mem(struct sljit_compiler *compiler, sljit_s32 flags, sljit_s32 reg, sljit_s32 arg, sljit_sw argw)
+static sljit_s32 emit_fop_mem(struct sljit_compiler *compiler, sljit_ins flags, sljit_s32 reg, sljit_s32 arg, sljit_sw argw)
```
zherczeg, line 1707: Is flags directly used in instruction creation? If not this change does not makes sense.

JetXujing: INT_SIZE or WORD_SIZE is used as the parameter call point, for example:
https://github.com/zherczeg/sljit/blob/7f28861274052a5d3975835d593fa1e56ecb7e6e/sljit_src/sljitNativeARM_64.c#L1765
https://github.com/zherczeg/sljit/blob/7f28861274052a5d3975835d593fa1e56ecb7e6e/sljit_src/sljitNativeARM_64.c#L1800
Therefore, the 'unsigned' type needs to be declared.

carenas: my suggestion was incorrect, this was addressed already in 335ee51f, by adding a cast directly in the macro that did the originally affected shift.

PR #161, opened 2023-04-18
arm64: avoid undefined behaviour with signed shifts

carenas:
As shown when building in Linux/macOS with `-fsanitize=undefined`:

```
$ bin/sljit_test 

sljit_src/sljitNativeARM_64.c:1964:13: runtime error: left shift of negative value -9223372036854775808
sljit_src/sljitNativeARM_64.c:1936:13: runtime error: left shift of negative value -2147483648
```


zherczeg:
Where is the signed shift?

carenas:
triggered by test93 as shown in the updated comment (and once test92 is commented out, of course to avoid the other bug)

zherczeg (approved):
LGTM

Issue #160, opened 2023-04-17
arm64: test92 infloops at least with M1 (apple's CPU)


carenas:
The issue is caused on the first test, but all others are likely affected as well, as the logic is similar.  What kind of atomic operation was being modeled in them?

From what I see the problem might be with the test itself, which invalidates the atomic read for SLJIT_R1 (AKA x1) by writing it back to an address just 8 bytes away (hence likely hitting the same cache line) as shown by the following disassembly of the generated code:

```
   0xffffb7fd0da0:	ldxr	x1, [x19]
   0xffffb7fd0da4:	str	x1, [x19,#8]
   0xffffb7fd0da8:	mov	x0, x1
   0xffffb7fd0dac:	mov	x1, #0xffffffffffffd980    	// #-9856
   0xffffb7fd0db4:	stxr	w9, x1, [x19]
   0xffffb7fd0db8:	cmp	w9, #0x0
   0xffffb7fd0dbc:	b.ne	0xffffb7fd0da0
```

If the store is done outside of the ldxr/stxr segment or at least 24 bytes away then it succeeds.

zherczeg:
After the reading, the data is written back for checking. The address is different from the region locked for atomic update, so it should not be a problem. In the arm64 doc there is a section for these instructions:

B2.7.3 Load-Acquire, Store-Release

It is hard to understand, but I don't see any restrictions for memory writes. Am I missing something? At least they should warn about this side effect, or is it a cpu bug?

carenas:
While a CPU bug is a possibility, I assume that it would be unlikely considering how common those are.

The most likely reason is AFAIK that the behaviour is permitted by the specification (in B2.9.3) as spelled with:

```
A Load-Exclusive instruction marks a small block of memory for exclusive access. The size of the marked block is IMPLEMENTATION DEFINED
```

While I assume it would be possible that someone would code a critical section that updates a close enough memory address that it will fail as in the test, I consider that sort of situation unlikely, which is why I was wondering what sort of atomic access was being modeled by the test.

It is worth also mentioning that the recommendation says:

```
In some implementations the CTR identifies the Exclusives reservation granule, see CTR_EL0. Otherwise, software must assume that the maximum Exclusives reservation granule, 512 words, is implemented.
```

zherczeg:
What does "exclusive access" exactly means? It is accessed only by the current thread, looks exclusive enough. The test does the basic atomic access, reads the memory, and writes another value if nobody updated the target value.

[...]

PR #162, opened 2023-04-23
arm64: refactor test92 to avoid possibly conflicting writes

carenas:
Avoid the infinite loop with M1, while trying to keep the test logic as similar as possible (eventhough some tests didn't seem needed and there is more cleanup pending)

Some tests were left doing some potentially unnecessary operations that were in the original, in an attempt to keep using the same registers for the atomic calls, but IMHO the register usage could do with some tightening and consistency.

Fixes: #160


test_src/sljitTest.c
```diff
@@ -11529,245 +11538,205 @@ static void test92(void)
...
-	sljit_s32 i;
+	struct sljit_jump *jump, *skip;
+	sljit_sw buf[12];
 
 	if (verbose)
 		printf("Run test92\n");
 
 	FAILED(!compiler, "cannot create compiler\n");
 
-	for (i = 0; i < 32; i++)
```
zherczeg, line 11541: The current testing works on big endian (e.g. s390x), I would not remove a lot of logic.

carenas: The new one with the macros does as well, this initialization loop is no longer needed because the `\xff` bytes are added by the macros instead and the end result is clearer to read.

zherczeg: If 32 tests is reduced to 12 then something is missing. I don't really understand the 1 letter long macros.

test_src/sljitTest.c
```diff
@@ -11529,245 +11538,205 @@ static void test92(void)
...
+	buf[11] = HV(sljit_u16, 1, 14876);
 
 	sljit_emit_enter(compiler, 0, SLJIT_ARGS1(VOID, P), 5, 5, 0, 0, 2 * sizeof(sljit_sw));
 
 	label = sljit_emit_label(compiler);
 	sljit_emit_atomic_load(compiler, SLJIT_MOV, SLJIT_R1, SLJIT_S0);
-	/* buf[1] */
-	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(SLJIT_S0), sizeof(sljit_sw), SLJIT_R1, 0);
-	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_R1, 0);
+	jump = sljit_emit_cmp(compiler, SLJIT_NOT_EQUAL, SLJIT_R1, 0, SLJIT_MEM1(SLJIT_S0), 0);
```
zherczeg, line 11566: I would keep the original logic of storing the result of loads.

test_src/sljitTest.c
carenas: Some "cases" where consolidated (eg, one checked the value, another the padding; I made them one single case preferably tested together with the same `FAILED`.

Others were removed because there are no longer writes (other than the required one by the LL/SC); the check to make sure the read succeeds is now done inside the test itself so there is no need for a `buf[]` for it.

An advantage of doing it inside the test is that it also verifies our assumption that reads to the "reserved area" are OK and would hopefully make it more visible if that assumption breaks with other implementations (ex: we might have a possible infloop with power)

test_src/sljitTest.c
carenas: Wouldn't this require then memory barriers to be added?, moving the store outside of the critical loop seems to work, but without a memory barrier it might get reordered and would cause failures (although not as obvious, as the infinite loop)

When refactoring, I thought about using a second array for the stores, but with a possible granule of 512 words, it will need ugly tricks to keep it safe anyway.

zherczeg: On ARM the doc says in "Load-Acquire, Store-Release" section:
- A Load-Acquire places no additional ordering constraints on any loads or stores appearing before the
Load-Acquire.
- The Store-Release places no additional ordering constraints on any loads or stores appearing after the
Store-Release instruction.

If barriers needs to be added on some cpus, the instruction should implicitly add them.

carenas: Note that (even before LSE) in aarch64 there are 2 pairs of instructions for its LL/SC.

Our implementation uses the weak (and better performing) pair of `ldxr`/`stxr` but their semantics don't correspond to what is used on other processors and that seem to match better the strong (and slower) pair (`ldaxr`/`stlxr`).

Not that anyone should, but if someone would implement a futex or mutex with sljit, it might work fine in x86/s390x and break with ARM (and probably the other RISC CPUs if implemented using LL/SC).

I have to admit that without a clear understanding on how this API is to be used, I am not sure if it is a problem but it might be something worth considering by maybe adding a "weak" parameter that would be a NOOP in strong ordered architectures and allow selecting the right pair in weak ordered ones like ARM.  

If this is to be modeled like C11/C++11 atomics then maybe a full set of memory_order options might be needed as well.

PR #163, opened 2023-04-24
x86_64: correctly detect mode32 for atomic_store

carenas:
test 92 case 1 not failing because of an unlucky choice of test values.

### DIFF
--- a/sljit_src/sljitNativeX86_common.c
+++ b/sljit_src/sljitNativeX86_common.c
@@ -3494,8 +3494,9 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_store(struct sljit_compiler
 	SLJIT_ASSERT(FAST_IS_REG(src_reg) || src_reg == SLJIT_MEM1(SLJIT_SP));
 	SLJIT_ASSERT(FAST_IS_REG(temp_reg) || temp_reg == SLJIT_MEM1(SLJIT_SP));
 
+	op = GET_OPCODE(op);
 #if (defined SLJIT_CONFIG_X86_32 && SLJIT_CONFIG_X86_32)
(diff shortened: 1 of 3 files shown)


sljit_src/sljitNativeX86_common.c
```diff
@@ -3549,6 +3549,15 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_store(struct sljit_compiler
 	INC_SIZE(1);
 	*inst = GROUP_LOCK;
 
+#if (defined SLJIT_CONFIG_X86_64 && SLJIT_CONFIG_X86_64)
+	if (!compiler->mode32) {
+		inst = (sljit_u8*)ensure_buf(compiler, 1 + 1);
+		FAIL_IF(!inst);
+		INC_SIZE(1);
+		*inst = REX_W;
```
zherczeg, line 3557: This prefix should be never added manually.

You can add a `op = GET_OPCODE(op);` after the SLJIT_ASSERT in the beginning of the function, and replace all `GET_OPCODE(op)` to `op`.

zherczeg (changes requested):
Nice catch!

sljit_src/sljitNativeX86_common.c
zherczeg: Why this is resolved? This code should be deleted.

zherczeg: To give background:
`compiler->mode32 = op != SLJIT_MOV && op != SLJIT_MOV_P;` does this, but I forgot that `op` has a flag, so it should be GET_OPCODE() guarded. Then, I realized we can do this early.

carenas: My bad, didn't understood your comment about the prefix, and indeed the root cause of the bug, thanks or your explanation

zherczeg (approved):
LGTM

PR #164, opened 2023-04-25
arm64: alternative implementation of atomics using CAS

carenas:
Works only on CPUs that implement LSE, and which is available since armv8.1-a (--march=armv8-a+lse), like the M1.

First patch implement the feature, last one is just leftover code that might be useful while debugging and therefore optional

### DIFF
--- a/sljit_src/sljitNativeARM_64.c
+++ b/sljit_src/sljitNativeARM_64.c
@@ -80,6 +80,9 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define BLR 0xd63f0000
 #define BR 0xd61f0000
 #define BRK 0xd4200000
+#define CAS 0xc8a07c00
+#define CASB 0x08a07c00
+#define CASH 0x48a07c00
 #define CBZ 0xb4000000
 #define CCMPI 0xfa400800
 #define CLZ 0xdac01000
@@ -152,6 +155,13 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define UBFM 0xd3400000
 #define UDIV 0x9ac00800
 #define UMULH 0x9bc03c00
+
+#define CSET (CSINC | RM(TMP_ZERO) | RN(TMP_ZERO))
+#define LDR (STRI | (1 << 22))
+#define LDRB (STRBI | (1 << 22))
+#define LDRH (LDRB | (1 << 30))
+#define LDRSW ((LDRI ^ (1 << 30)) ^ (0x3 << 22))
+#define MOV (ORR | RN(TMP_ZERO))
 
 static sljit_s32 push_inst(struct sljit_compiler *compiler, sljit_ins ins)
 {
@@ -2487,11 +2497,30 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_load(struct sljit_compiler 
 	sljit_s32 dst_reg,
 	sljit_s32 mem_reg)
 {
-	sljit_ins ins = 0;
+	sljit_ins ins;
 
 	CHECK_ERROR();
 	CHECK(check_sljit_emit_atomic_load(compiler, op, dst_reg, mem_reg));
 
+#ifdef __ARM_FEATURE_ATOMICS
+	switch (GET_OPCODE(op)) {
+	case SLJIT_MOV32:
+		ins = LDR ^ (1 << 30);
+		break;
+	case SLJIT_MOV_U32:
+		ins = LDRSW;
+		break;
+	case SLJIT_MOV_U16:
+		ins = LDRH;
+		break;
+	case SLJIT_MOV_U8:
+		ins = LDRB;
+		break;
+	default:
+		ins = LDR;
+		break;
+	}
+#else /* !__ARM_FEATURE_ATOMICS */
 	switch (GET_OPCODE(op)) {
 	case SLJIT_MOV32:
 	case SLJIT_MOV_U32:
@@ -2507,7 +2536,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_load(struct sljit_compiler 
 		ins = LDXR;
 		break;
 	}
-
+#endif /* ARM_FEATURE_ATOMICS */
 	return push_inst(compiler, ins | RN(mem_reg) | RT(dst_reg));
 }
 
@@ -2517,9 +2546,53 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_store(struct sljit_compiler
 	sljit_s32 temp_reg)
 {
 	sljit_ins ins;
+	sljit_s32 tmp = temp_reg;
+	sljit_ins cmp = 0;
+	sljit_ins inv_bits = W_OP;
 
 	CHECK_ERROR();
 	CHECK(check_sljit_emit_atomic_store(compiler, op, src_reg, mem_reg, temp_reg));
+
+#ifdef __ARM_FEATURE_ATOMICS
+	if (op & SLJIT_SET_ATOMIC_STORED)
+		cmp = (SUBS ^ W_OP) | RD(TMP_ZERO);
+
+	switch (GET_OPCODE(op)) {
+	case SLJIT_MOV32:
+	case SLJIT_MOV_U32:
+		ins = CAS ^ (1 << 30);
+		break;
+	case SLJIT_MOV_U16:
+		ins = CASH;
+		break;
+	case SLJIT_MOV_U8:
+		ins = CASB;
+		break;
+	default:
+		ins = CAS;
+		inv_bits = 0;
+		if (cmp)
+			cmp ^= W_OP;
+		break;
+	}
+
+	if (cmp) {
+		FAIL_IF(push_inst(compiler, MOV ^ inv_bits | RM(temp_reg) | RD(TMP_REG1)));
+		tmp = TMP_REG1;
+	}
+	FAIL_IF(push_inst(compiler, ins | RM(tmp) | RN(mem_reg) | RD(src_reg)));
+	if (!cmp)
+		return SLJIT_SUCCESS;
+
+	FAIL_IF(push_inst(compiler, cmp | RM(tmp) | RN(temp_reg)));
+	FAIL_IF(push_inst(compiler, CSET ^ inv_bits | RD(tmp)));
+	return push_inst(compiler, cmp | RM(tmp) | RN(TMP_ZERO));
+#else /* !__ARM_FEATURE_ATOMICS */
+	SLJIT_UNUSED_ARG(tmp);
+	SLJIT_UNUSED_ARG(inv_bits);
+
+	if (op & SLJIT_SET_ATOMIC_STORED)
+		cmp = (SUBI ^ W_OP) | (1 << 29);
 
 	switch (GET_OPCODE(op)) {
 	case SLJIT_MOV32:
@@ -2538,7 +2611,8 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_store(struct sljit_compiler
 	}
 
 	FAIL_IF(push_inst(compiler, ins | RM(TMP_REG1) | RN(mem_reg) | RT(src_reg)));
-	return push_inst(compiler, (SUBI ^ W_OP) | (1 << 29) | RD(TMP_ZERO) | RN(TMP_REG1));
+	return cmp ? push_inst(compiler, cmp | RD(TMP_ZERO) | RN(TMP_REG1)) : SLJIT_SUCCESS;
+#endif /* __ARM_FEATURE_ATOMICS */
 }
 
 SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_get_local_base(struct sljit_compiler *compiler, sljit_s32 dst, sljit_sw dstw, sljit_sw offset)
--- a/test_src/sljitTest.c
+++ b/test_src/sljitTest.c
@@ -11530,7 +11530,7 @@ static void test92(void)
 	struct sljit_compiler *compiler = sljit_create_compiler(NULL, NULL);
 	struct sljit_label *label;
 	struct sljit_jump *jump;
-	sljit_sw buf[34];
+	sljit_sw buf[38];
 	sljit_s32 i;
 
 	if (verbose)
@@ -11538,11 +11538,11 @@ static void test92(void)
 
 	FAILED(!compiler, "cannot create compiler\n");
 
-	for (i = 0; i < 34; i++)
+	for (i = 0; i < 36; i++)
 		buf[i] = WCONST(0x5555555555555555, 0x55555555);
 
 	buf[0] = 4678;
-	*(sljit_u8*)(buf + 2) = 178;
+	*(sljit_u8*)(buf + 2) = 78;
 	*(sljit_u8*)(buf + 5) = 211;
 	*(sljit_u16*)(buf + 9) = 17897;
 	*(sljit_u16*)(buf + 12) = 57812;
@@ -11552,6 +11552,10 @@ static void test92(void)
 	((sljit_u8*)(buf + 26))[1] = 105;
 	((sljit_u8*)(buf + 28))[2] = 13;
 	((sljit_u16*)(buf + 31))[1] = 14876;
+#if (defined SLJIT_64BIT_ARCHITECTURE && SLJIT_64BIT_ARCHITECTURE)
+	((sljit_s32*)(buf + 33))[1] = -1;
+#endif /* SLJIT_64BIT_ARCHITECTURE */
+	buf[36] = 0xffaa55;
 
 	sljit_emit_enter(compiler, 0, SLJIT_ARGS1(VOID, P), 5, 5, 0, 0, 2 * sizeof(sljit_sw));
 
@@ -11591,7 +11595,7 @@ static void test92(void)
 	sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R0, 0, SLJIT_S0, 0, SLJIT_IMM, 5 * sizeof(sljit_sw));
 	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_S1, 0, SLJIT_IMM, 97);
 	/* buf[5] */
-	sljit_emit_atomic_store(compiler, SLJIT_MOV_U8 | SLJIT_SET_ATOMIC_STORED, SLJIT_S1, SLJIT_R0, SLJIT_S2);
+	sljit_emit_atomic_store(compiler, SLJIT_MOV32_U8 | SLJIT_SET_ATOMIC_STORED, SLJIT_S1, SLJIT_R0, SLJIT_S2);
 	sljit_set_label(sljit_emit_jump(compiler, SLJIT_ATOMIC_NOT_STORED), label);
 	/* buf[6] */
 	sljit_emit_op1(compiler, SLJIT_MOV32, SLJIT_MEM1(SLJIT_S0), 6 * sizeof(sljit_sw), SLJIT_R1, 0);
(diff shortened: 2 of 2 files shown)


zherczeg:
"atomic_store doesn't incorrectly set the atomic value if it changes while in the critical zone" - based on the discussion in #160 we should simply say, that any store between the load/store makes the operation undefined.

"the wrong value is provided as "expected" through temp_reg" - this already makes the operation undefined

There should be a test for reading data between the load/store. There should be a test for op_flags as well (an existing test can be extended though).

The atomic load/store must satisfy many rules, maybe creating a list of them would look better.

Do you want to work on this or shall I do it?

carenas:
> "atomic_store doesn't incorrectly set the atomic value if it changes while in the critical zone" - based on the discussion in #160 we should simply say, that any store between the load/store makes the operation undefined.

OK, but it should be important to note that might apply to both read and write operations, and not only to the memory address, but also its surrounding area.

Indeed, really old ARM cpu's which luckily we don't support, would invalidate locks if any write was done anywhere in memory[1]

> "the wrong value is provided as "expected" through temp_reg" - this already makes the operation undefined

That needs clarification in the documentation as well, note that the fact that ARM, and any other
implementation that uses LL/SC would just happily ignore that the value doesn't match results in the store incorrectly succeeding as shown by the test, and which is why it is a bug against the specification that says that if the value doesn't match then the memory is left unchanged.

I would also argue that calling it "temp_reg" is deceiving, and should be instead "state_reg" or something that better identifies it's actual use.

> There should be a test for reading data between the load/store. There should be a test for op_flags as well (an existing test can be extended though).

I would instead argue there should be no reads, so that there are no more surprises when all other CPUs are implemented.

> Do you want to work on this or shall I do it?

Considering I might seem to be still confused on how this is to be used, it might be better if you do it.

[1] https://en.wikipedia.org/wiki/Load-link/store-conditional

zherczeg:
> OK, but it should be important to note that might apply to both read and write operations, and not only to the memory address, but also its surrounding area.

I would not limit it to an area. All memory writes, regardless of the address, makes the operation undefined. I cannot think any practical uses, where writing cannot be postponed after the store.

> I would also argue that calling it "temp_reg" is deceiving, and should be instead "state_reg" or something that better identifies it's actual use.

The comment says: temp_reg must contain the value loaded into dst_reg by the sljit_emit_atomic_load operation (the store operation does not preserve the value of temp_reg)

I think "temp" emphasize that its value is not preserved, but I am open to any ideas. I don't like the "state" term, since it does not contain any status.

> I would instead argue there should be no reads, so that there are no more surprises when all other CPUs are implemented.

This is certainly possible, but it would be strange, if reads would block a memory store.

test_src/sljitTest.c
```diff
@@ -11707,6 +11708,29 @@ static void test92(void)
 	sljit_emit_atomic_store(compiler, SLJIT_MOV_U16 | SLJIT_SET_ATOMIC_STORED, SLJIT_R2, SLJIT_R1, SLJIT_R0);
 	sljit_set_label(sljit_emit_jump(compiler, SLJIT_ATOMIC_NOT_STORED), label);
 
+#if !defined(SLJIT_CONFIG_ARM) || defined(__ARM_FEATURE_ATOMICS)
+	sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R1, 0, SLJIT_S0, 0, SLJIT_IMM, 32 * sizeof(sljit_sw));
```
zherczeg, line 11712: ARM now has compare and swap?  Sounds great! Compare and swap also used by x86 and s390x, so these tests should cover them. I would like to add a comment that these tests compare and swap operation specifically.

sljit_src/sljitNativeARM_64.c
```diff
@@ -2521,6 +2537,28 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_store(struct sljit_compiler
...
+		break;
+	default:
+		ins = CASL;
+		break;
+	}
+	FAIL_IF(push_inst(compiler, atomic_load_ins(op) | RN(mem_reg) | RT(TMP_REG2)));
+	FAIL_IF(push_inst(compiler, SUBS | RM(TMP_REG2) | RN(temp_reg) | RD(TMP_REG1)));
+	FAIL_IF(push_inst(compiler, B_CC | (2 << 5) | 0x1));
+	FAIL_IF(push_inst(compiler, ins | RM(temp_reg) | RN(mem_reg) | RD(src_reg)));
+	FAIL_IF(push_inst(compiler, CLREX));
```
zherczeg, line 2560: This looks like a lot of instruction, Usually the point of these instruction is simplicity, why a single CAS instruction is not enough like on other cpus?

carenas: It is still a RISC cpu, and having 3 arguments was probably alien enough that they forgot to return a flag to indicate if their CAS was successful or not.

if we are keeping the "undefined behaviour" if the temp register doesn't match then it could be done in a much simpler way, albeit IMHO not as robust:

  https://github.com/carenas/sljit/commit/b1e13c8064b45e7808833ceee35ed35f8e0a7885

If instead of doing only load/store, we are to provide atomic operations (like atomic add) then it will be easier to map them to the CPU instructions.

zherczeg: Sljit has checks (asserts) to try to help when the api is misused, but in general, no runtime checks are generated, since they reduce the performance. If you want, you could try invent a way to check these load/store pairs at compile time, probably could help a lot.

sljit_src/sljitNativeARM_64.c
```diff
@@ -2508,19 +2510,58 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_load(struct sljit_compiler
...
+		break;
+	case SLJIT_MOV_U16:
+		ins = CASLH;
+		break;
+	default:
+		ins = CASL;
+		cmp = SUBS;
+		break;
+	}
+	FAIL_IF(push_inst(compiler, atomic_load_ins(op) | RN(mem_reg) | RT(TMP_REG2)));
```
zherczeg, line 2558: This value is in temp_reg. Could be moved to TMP_REGx and do a comparison later.

carenas: No, that is a fresh atomic load of the memory, which then will let us know if the CAS operation failed by comparing it with the value from temp_reg.

There is a small reservation being done between this and the next instruction, to protect the atomic from being changed by another thread.

zherczeg: It looks like the AL variants do the acquire and release as well:

https://developer.arm.com/documentation/ddi0602/2022-06/Base-Instructions/CASB--CASAB--CASALB--CASLB--Compare-and-Swap-byte-in-memory-

That is the normal compare and swap, so perhaps we could use it.

test_src/sljitTest.c
```diff
@@ -11555,7 +11556,6 @@ static void test92(void)
 
 	sljit_emit_enter(compiler, 0, SLJIT_ARGS1(VOID, P), 5, 5, 0, 0, 2 * sizeof(sljit_sw));
 
-	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(SLJIT_S0), sizeof(sljit_sw), SLJIT_IMM, -9856);
```
zherczeg, line 11558: Why is this removed? It checks that you can write data before sljit_emit_atomic_load

test_src/sljitTest.c
```diff
@@ -11570,9 +11570,11 @@ static void test92(void)
 	sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R1, 0, SLJIT_S0, 0, SLJIT_IMM, 2 * sizeof(sljit_sw));
 	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(SLJIT_S0), 3 * sizeof(sljit_sw), SLJIT_IMM, 203);
 	label = sljit_emit_label(compiler);
+	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R2, 0, SLJIT_IMM, -1);
 	sljit_emit_atomic_load(compiler, SLJIT_MOV_U8, SLJIT_R2, SLJIT_R1);
 	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_S2, 0, SLJIT_R2, 0);
-	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_MEM1(SLJIT_S0), 3 * sizeof(sljit_sw));
+	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_IMM, -1);
```
zherczeg, line 11576: This is dead code.

test_src/sljitTest.c
```diff
@@ -11602,6 +11604,7 @@ static void test92(void)
 
 	label = sljit_emit_label(compiler);
 	sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R1, 0, SLJIT_S0, 0, SLJIT_IMM, 9 * sizeof(sljit_sw));
+	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_S2, 0, SLJIT_IMM, WCONST(0xAAAAAAAAAAAAAAAAu, 0xAAAAAAAAu));
```
zherczeg, line 11607: What does this test here?

test_src/sljitTest.c
```diff
@@ -11727,6 +11731,11 @@ static void test92(void)
 	/* buf[33] */
 	sljit_emit_op_flags(compiler, SLJIT_MOV, SLJIT_MEM1(SLJIT_S0), 33 * sizeof(sljit_sw), SLJIT_ATOMIC_STORED);
 
+	sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R1, 0, SLJIT_S0, 0, SLJIT_IMM, 34 * sizeof(sljit_sw));
+	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R2, 0, SLJIT_MEM1(SLJIT_S0), 32 * sizeof(sljit_sw));
+	sljit_emit_atomic_load(compiler, SLJIT_MOV, SLJIT_R0, SLJIT_R1);
+	sljit_emit_atomic_store(compiler, SLJIT_MOV, SLJIT_R2, SLJIT_R1, SLJIT_R0);
```
zherczeg, line 11737: This can be unsuccessful, and that is valid.

test_src/sljitTest.c
```diff
@@ -11785,7 +11794,7 @@ static void test92(void)
 	FAILED(((sljit_u16*)(buf + 31))[1] != 51403, "test92 case 44 failed\n");
 	FAILED(buf[32] != 14876, "test92 case 45 failed\n");
 	FAILED(buf[33] != 1, "test92 case 46 failed\n");
-
```
zherczeg, line 11788: Why this line is removed?

test_src/sljitTest.c
carenas: sljit_emit_atomic_load() can't prevent any write operations before it is even executed

test_src/sljitTest.c
carenas: That the load doesn't spill over its specified register width, or even if it does, that the rest of the transaction is safe from it.

Indeed, might be worth adding to the spec what is the relevant width specially for U8 and U16.

test_src/sljitTest.c
carenas: I mainly added it for helping with future implementations, and to make sure that the "no flag passed" codepath is covered.  While it could fail, it would be very unlikely as part of sljit_test and might be removed once all other implementations are completed.

test_src/sljitTest.c
carenas: also, it looked weird when reviewing the generated code, as the same memory address was being overwritten a few instructions later

sljit_src/sljitNativeARM_64.c
carenas: Not sure if I understand; the code is already using "acquire/release" semantics in the small critical section.

do you mean to replace `lxdr` with `casa`?, the current version seems to be working fine at least on the M1 (which is the only ARM CPU I have with LSE); it would allow to reintroduce the "fail if temp_reg doesn't match", but thought you were not interested on that "feature".

I have to admit I am still not clear on the use cases for this, so not sure if my tests would give me enough confidence to decide between the two options.

carenas: > the code is already using "acquire/release" semantics in the small critical section.

except it doesn't work the way I expected, will update once I get one that does and hopefully will be even less code.

zherczeg: Something like:
```
TMP_REG1 = tmp_reg
CASALB TMP_REG1, src, [mem]
CMP TMP_REG1, tmp_reg
```
So CASB has 4 varaints, the AL variant do both part of the lock.

test_src/sljitTest.c
zherczeg: That is the whole point! The memory can be initialized one instruction earlier, and it must still work.

test_src/sljitTest.c
```diff
@@ -11568,8 +11569,8 @@ static void test92(void)
 	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(SLJIT_S0), sizeof(sljit_sw), SLJIT_R2, 0);
 
 	sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R1, 0, SLJIT_S0, 0, SLJIT_IMM, 2 * sizeof(sljit_sw));
-	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(SLJIT_S0), 3 * sizeof(sljit_sw), SLJIT_IMM, 203);
```
zherczeg, line 11571: Same as above, the point is writing just before the load, and the load must work.

test_src/sljitTest.c
zherczeg: The load overwrites the whole register. The upper bits are zero. You have already made a SLJIT_R2=-1 test above for this, so maybe you can also use -1 here.

test_src/sljitTest.c
```diff
@@ -11631,10 +11633,11 @@ static void test92(void)
 
 	sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_S2, 0, SLJIT_S0, 0, SLJIT_IMM, 15 * sizeof(sljit_sw));
 	label = sljit_emit_label(compiler);
+	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R2, 0, SLJIT_IMM, WCONST(0xAAAAAAAAAAAAAAAAu, 0xAAAAAAAAu));
 	sljit_emit_atomic_load(compiler, SLJIT_MOV_U32, SLJIT_R2, SLJIT_S2);
 	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_R2, 0);
 	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_S3, 0, SLJIT_R2, 0);
-	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R1, 0, SLJIT_IMM, 987654321);
+	sljit_emit_op1(compiler, SLJIT_MOV_U32, SLJIT_R1, 0, SLJIT_IMM, 987654321);
```
zherczeg, line 11640: Testing this should be elsewhere.

test_src/sljitTest.c
```diff
@@ -11647,15 +11650,15 @@ static void test92(void)
 	sljit_emit_atomic_load(compiler, SLJIT_MOV32, SLJIT_R0, SLJIT_R2);
 	sljit_emit_op1(compiler, SLJIT_MOV32, SLJIT_S1, 0, SLJIT_R0, 0);
 	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_S2, 0, SLJIT_R0, 0);
-	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_IMM, -573621);
+	sljit_emit_op1(compiler, SLJIT_MOV32, SLJIT_R0, 0, SLJIT_IMM, -573621);
```
zherczeg, line 11653: This way you don't test that the upper 32 bits on 64 bit are preserved.

sljit_src/sljitNativeARM_64.c
carenas: sadly CASA doesn't work like that; the fetched value is lost so the only way to know if it succeeds or not, is by fetching it again for the CMP as shown in the new version of the code.

I asm using CAS instead as it is "atomic" anyway but left the additional ops defined for swapping them if needed.

test_src/sljitTest.c
carenas: I made that test explicit with case 27, and to protect against false positives when the filling was 0xff, but you are correct it is no longer needed; will remove

test_src/sljitTest.c
zherczeg: Flags may be set even if you don't request to set them. For atomics, not setting flag is quite pointless, so I was lazy and set it all the time. But this test is ok as long as the result is not checked.

test_src/sljitTest.c
carenas: I got confused by my own change (though it was yours), agree it is useless, and indeed I think it (the oe using -1) might had been introduced by mistake from my own debugging.

Would it make sense to keep some of these enabled under some DEBUG mode?, will remove them otherwise.

sljit_src/sljitNativeARM_64.c
```diff
@@ -2517,10 +2536,62 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_store(struct sljit_compiler
...
+		FAIL_IF(push_inst(compiler, (ANDI ^ W_OP) | RN(src) | RD(TMP_REG1) | (0x1e << 9)));
+		src = TMP_REG1;
+		break;
+	case BYTE_SIZE:
+		FAIL_IF(push_inst(compiler, (ANDI ^ W_OP) | RN(src) | RD(TMP_REG1) | (0xe << 9)));
+		src = TMP_REG1;
+		break;
+	}
+
+	FAIL_IF(emit_op_mem(compiler, mem_flags, TMP_REG2, SLJIT_MEM1(mem_reg), 0, TMP_REG2));
```
carenas, line 2587: this works in practice, but it would be ideal if replaced with `push_inst()` specially to avoid having to rely on `emit_op_mem()` not obfuscating the value from TMP_REG2.

alternatively `temp_reg` could be used instead

sljit_src/sljitNativeARM_64.c
```diff
@@ -2509,18 +2532,68 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_load(struct sljit_compiler
...
 
+#ifdef __ARM_FEATURE_ATOMICS
+	if (op & SLJIT_SET_ATOMIC_STORED)
+		cmp = (SUBS ^ W_OP) | RD(TMP_ZERO);
+
+	switch (GET_OPCODE(op)) {
+	case SLJIT_MOV32:
+	case SLJIT_MOV_U32:
+		ins = CAS ^ (1 << 30);
+		l = LDRSW;
```
carenas, line 2559: Like all other operations, this will cause "Undefined Behaviour" if not correctly aligned, but would need special treatment to avoid that if the higher half or a worded size address is targeted.

test_src/sljitTest.c
zherczeg: I have an idea here. You could do a regular memory load after the atomic store, and if the value does not match, go back. No thread will overwrite the value, so it should be good enough for testing.

sljit_src/sljitNativeARM_64.c
```diff
@@ -2517,10 +2565,54 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_store(struct sljit_compiler
...
+		return SLJIT_SUCCESS;
+
+	if (mask) {
+		FAIL_IF(push_inst(compiler, (ANDI ^ W_OP) | RN(src) | RD(TMP_REG1) | logical_imm(mask, 16)));
+		src = TMP_REG1;
+	}
+
+	FAIL_IF(push_inst(compiler, atomic_load_ins(op) | RN(mem_reg) | RD(TMP_REG2)));
+	return push_inst(compiler, cmp | RM(TMP_REG2) | RN(src));
+#else
```
zherczeg, line 2610: comment missing: /* !__ARM_FEATURE_ATOMICS */

sljit_src/sljitNativeARM_64.c
```diff
@@ -2537,8 +2629,9 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_store(struct sljit_compiler
 		break;
 	}
 
-	FAIL_IF(push_inst(compiler, ins | RM(TMP_REG1) | RN(mem_reg) | RT(src_reg)));
-	return push_inst(compiler, (SUBI ^ W_OP) | (1 << 29) | RD(TMP_ZERO) | RN(TMP_REG1));
+	FAIL_IF(push_inst(compiler, ins | RM(TMP_REG1) | RN(mem_reg) | RT(src)));
+	return cmp ? push_inst(compiler, cmp | RD(TMP_ZERO) | RN(TMP_REG1)) : SLJIT_SUCCESS;
+#endif /* Armv8.1 LSE */
```
zherczeg, line 2634: /* __ARM_FEATURE_ATOMICS */

sljit_src/sljitNativeARM_64.c
zherczeg: The original code only used two instructions, so what is the advantage of this variant?
An idea: keep the load the way it is, and just use the correct casxx, that would also reduce the needed instructions.

zherczeg (changes requested):

test_src/sljitTest.c
```diff
@@ -11617,6 +11620,9 @@ static void test92(void)
 	sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R1, 0, SLJIT_S0, 0, SLJIT_IMM, 12 * sizeof(sljit_sw));
 	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(SLJIT_S0), 13 * sizeof(sljit_sw), SLJIT_IMM, 41306);
 	label = sljit_emit_label(compiler);
+#ifdef DEVELOPER
```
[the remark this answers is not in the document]
carenas: The difference I see, is that this kind of noise is more likely to help catch bugs on the implementation while it is being developed or the test logic itself.

Having them labeled, also reduces the cognitive logic required to understand the test and what is being tested IMHO

sljit_src/sljitNativeARM_64.c
carenas: The advantages of this code (compared with LL/SC) are:

* The "critical section" where operations could cause a deadlock (because of bugs, but also "false sharing") is eliminated.
* Performs better under high contention.

The additional instructions are needed to test the result of CAS and can't be eliminated by extending the critical area.  Remember that a small critical area that used a pair of `ldxr`/`casl` was discarded because it didn't work.

carenas: > sadly CASA doesn't work like that

and I was wrong; the fetched value doesn't get lost and is indeed accessible from `temp_reg` as shown by the latest version of the code

sljit_src/sljitNativeARM_64.c
```diff
@@ -80,8 +80,21 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
...
+#define CASAB 0x08e07c00
+#define CASAH 0x48e07c00
+#define CASAL 0xc8e0fc00
+#define CASALB 0x08e0fc00
+#define CASALH 0x48e0fc00
+#define CASB 0x08a07c00
+#define CASH 0x48a07c00
+#define CASL 0xc8a0fc00
+#define CASLB 0x08a0fc00
+#define CASLH 0x48a0fc00
```
zherczeg, line 94: Is it worth to add instruction which we don;t use?

zherczeg:
I think the patch is quite nice now!

You can add `/**/` comments freely anywhere to describe things, for me that is better than DEVELOPER.

I realized one type of test is missing: the double load. 
- Create a static global variable, should be far from the stack allocated buffer
- Load the global value with atomic load
- Load a local value value with atomic load
- Store a local value value with atomic store
This should work. A missing store should have no side effects.

sljit_src/sljitNativeARM_64.c
carenas: FWIW I used them while developing, so had been not only encoded but verified to be correct and could save some time when we have a need to use them later.

test_src/sljitTest.c
carenas: Doing instead a byte that is always atomic and can't be perturbed as there is no conflicting access.

carenas:
At least with ARM's LL/SC, a second load invalidates the reservation of the first, which IMHO would be also in line with the constrain to only support one atomic per thread that was documented.

zherczeg (approved):
LGTM. Thank you for doing all the changes

PR #166, opened 2023-05-02
tidy up #164

carenas:
Use MOV when appropriate, make test92 case 50 big endian safe and cleaner and documentation update

### DIFF
--- a/sljit_src/sljitLir.h
+++ b/sljit_src/sljitLir.h
@@ -1774,6 +1774,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fmem_update(struct sljit_compiler 
      - the memory operation (op) and the base address (stored in mem_reg)
        passed to the load/store operations must be the same (the mem_reg
        can be a different register, only its value must be the same)
+     - an store must always follow a load for the same transaction.
 
    op must be between SLJIT_MOV and SLJIT_MOV_P, excluding all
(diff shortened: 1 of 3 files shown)


sljit_src/sljitLir.h
```diff
@@ -1774,6 +1774,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fmem_update(struct sljit_compiler
      - the memory operation (op) and the base address (stored in mem_reg)
        passed to the load/store operations must be the same (the mem_reg
        can be a different register, only its value must be the same)
+     - an store must always follow a load for the same transaction.
```
zherczeg, line 1777: I hope we can avoid such requirement. Sometimes you want to do conditional updates, and want to do nothing if the condition failed.

Is there a cpu where this is required?

carenas: > Sometimes you want to do conditional updates, and want to do nothing if the condition failed.

Could you elaborate?

> Is there a cpu where this is required?

Presume all CPUS that will be implemented using LL/SC; definitely ARM64 will always fail the `stxr` call, unless the memory address that it refers to was first added to the monitor by its previous `lxdr`.

zherczeg: I am sorry. I misread the text.

zherczeg (approved):
LGTM

Issue #168, opened 2023-05-12
Incorrect or unclear documentation on function signatures


radgeRayden:
Learning the library and reading the following snippet: https://github.com/zherczeg/sljit/blob/6a8dc94e06e85a8380da2f0eef999f875576dcb8/sljit_src/sljitLir.h#L293-L297

As I understand it the pointer argument should be actually stored in `S0`, as it doesn't have the `_R` postfix. I also found it pretty confusing that the arguments aren't in order (sorted by register?), which took me a few readings to realize. Maybe a note would be useful.

On the later examples it appears that the rule is that if you have a register `S0`, the next integer argument will have a `1` index, even if it's a scratch register instead of a saved one. This isn't explicitly mentioned anywhere on the text, as far as I can tell. To be clear I didn't read the whole documentation yet, but I double checked the mentioned section.

zherczeg:
This describes the argument passing for `calls`:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitLir.h#L263

This describes the argument receiving by `enter`:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitLir.h#L268

These are similar, but for `enter`, you have an extra option to to move some arguments to saved registers. If you do that, the scratch register normally assigned to the argument is undefined.

I admit parameter passing is complicated, but also very practical.

> This isn't explicitly mentioned anywhere on the text, as far as I can tell.

https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitLir.h#L270

radgeRayden:
> https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitLir.h#L270

The confusion is on whether the saved registers increment the index counting for the scratch registers. I admit I might just not be getting some details, so if you think the documentation is clear enough I will keep reading and playing with it. Thanks.

zherczeg:
I am open to suggestions. The definition of "clear enough" is different for all people, and more text never hurts. Just because technically something is there, it is not necessary easy to understand. I am not a native speaker, so explaining things is not that easy for me.

radgeRayden:
> The index of the scratch register is the count of the previous integer arguments starting from SLJIT_R0.

Which is a pretty concise, but indeed correct description.
The mention of `SLJIT_R0` here made me think the count was referring to scratch registers only.
[...]

PR #169, opened 2023-05-16
documentation: refactor registers and function signatures

carenas:
Slightly expanded to avoid confusing references and describe better the way that the use of scratch registers affect `sljit_emit_enter()`

Fixes: #168

### DIFF
--- a/sljit_src/sljitLir.h
+++ b/sljit_src/sljitLir.h
@@ -109,7 +109,7 @@ extern "C" {
 #define SLJIT_ERR_EX_ALLOC_FAILED	3
 /* Unsupported instruction form. */
 #define SLJIT_ERR_UNSUPPORTED		4
-/* An ivalid argument is passed to any SLJIT function. */
+/* An invalid argument is passed to any SLJIT function. */
 #define SLJIT_ERR_BAD_ARGUMENT		5
 
 /* --------------------------------------------------------------------- */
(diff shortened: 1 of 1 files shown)


sljit_src/sljitLir.h
```diff
@@ -258,20 +258,32 @@ extern "C" {
...
+
+   Furthermore, the index that will be used for the scratch register depends
+   on how many previous integer registers (scratch or saved) were allocated,
+   starting with SLJIT_R0. See below for some examples.
+
+   The floating point arguments always use scratch registers (but not the
+   _R suffix like in the integer registers) and must be placed into SLJIT_FR0,
+   SLJIT_FR1, just like in the "call" functions.
 
    Note: if a function is called by sljit_emit_call/sljit_emit_icall and
```
carenas, line 288: I don't understand this "Note", specially the references to the index affecting the floating point register index somehow and that would seem to conflict with the fact that floating point registers should be always using `SLJIT_FR0 ..`

sljit_src/sljitLir.h
```diff
@@ -127,40 +127,40 @@ extern "C" {
...
   The purpose of this definition is convenience: saved registers can
-  be used as extra scratch registers. For example four registers can
-  be specified as scratch registers and the fifth one as saved register
-  on the CPU above and any user code which requires four scratch
-  registers can run unmodified. The SLJIT compiler automatically saves
-  the content of the two extra scratch register on the stack. Scratch
-  registers can also be preserved by saving their value on the stack
-  but this needs to be done manually.
+  be used as extra scratch registers. For example, in the previous
+  architecture four registers can be specified as scratch registers
```
carenas, line 148: s/architecture four/sample CPU, four/

the rest of the paragraph also needs more commas

sljit_src/sljitLir.h
zherczeg: It wants to explain that floating point arguments cannot be moved to saved registers, so the _R prefix is meaningless, and not used.

Thank you for improving the text.

sljit_src/sljitLir.h
zherczeg: Ok. You can also rephrase them to use more sentences.

sljit_src/sljitLir.h
carenas: I was referring to the note in https://github.com/zherczeg/sljit/blame/47ed75d0bcbe77f3e82b89e6e27f6b535815b821/sljit_src/sljitLir.h#LL276-L279C47

zherczeg: The note explains that if an argument is not moved to a saved register, the register assigned to the argument "keeps" its value. I.e., if you put an argument to R2 for the call, it is still in R2 after emit_enter, unless it is moved to a saved register (R2 is undefined in the latter case, just like the other registers).

zherczeg:
If this is the final version, I can land it

zherczeg (approved):
LGTM

PR #170, opened 2023-05-29
add base support for LoongArch 64bit

Xiao-Tao:
LoongArch is a new RISC ISA , which includes a 32-bit version and a 64-bit version, Here are some documents about LoongArch:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
more:
https://loongson.github.io/LoongArch-Documentation/README-EN.html

We tested according to the data and test methods provided by https://zherczeg.github.io/sljit/regex_perf.html, and the performance of using JIT is much higher than that of not using it.
Hope to support LoongArch architecture
Thank you.


zherczeg:
This looks like quite a lot of work, thank you very much for the contribution. I have no experience with the architecture, nor I have any hardware, so my first question is who will be a (long term) maintainer of the port. The other question is 32 bit support, is there any plan to support it? If yes, the 64 bit specific part should be in a separate file similar to other architectures.

Xiao-Tao:
I think it's an interesting job to support LoongArch. 
1. At present, all support except atomic instructions has been completed, and I can maintain the implementation of new functions and bug fixes of the LoongArch architecture for a long term
2. In the absence of hardware, use the CLFS (Cross Linux From Scratch) cross-compilation toolchain provided by https://github.com/sunhaiyong1978/CLFS-for-LoongArch/releases and the system environment with qemu to complete the test (latest The version of qemu already supports LoongArch64 architecture)
The debian community has started to support the LoongArch architecture
https://wiki.debian.org/Ports/loong64
3. At present, gcc/glibc/binutils does not fully support LoongArch32. After the subsequent LoongArch32 toolchain support is improved, I will implement sljit’s support for LoongArch32 and then modify it to sljitNativeLOONGARCH_(common/32/64).c

zherczeg:
I have no resources to maintain this port, so it is up to you to do it.

carenas:
> 1. all support except atomic instructions has been completed,

why wasn't atomic instructions completed?

also, what is the correct spelling for your name so it could be added to the README?

Xiao-Tao:
> > 1. all support except atomic instructions has been completed,
> 
> why wasn't atomic instructions completed?
> 
1. Use LL.{W/D}-SC.{W/D} in the LoongArch64 architecture to implement atomic memory access.
Reference link: https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html#_ll_wd_sc_wd

2. LL-SC only supports 32bit and 64bit, there may be problems with SLJIT_MOV_U8/U16 implementation

3. "SC.{W/D} rd, rj, si14" After storing the value in the rd register into the memory, the result 0/1 (1-success, 0-failure) will be written into the rd register, which cannot be stored in the rd register retain the original content.

> also, what is the correct spelling for your name so it could be added to the README?

Mingtao Zhou

carenas:
> > > 1. all support except atomic instructions has been completed,
> > 
> > why wasn't atomic instructions completed?

my question wasn't answered, I was curious on the why, not the how.
I presume that since it is not needed for PCRE2 it just wasn't prioritized?

the reason why I am interested is that I think a working implementation of atomics for sljit, is likely to help guide the design of this API, which is otherwise experimental, and I have personally some doubts on its fit.

> 1. Use `LL.{W/D}-SC.{W/D}` in the LoongArch64 architecture to implement atomic memory access.
>    Reference link: https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html#_ll_wd_sc_wd

I had a WIP version of it, which has still issues; indeed I have 2 versions of it (both with issues, but I
suspect the one that ALSO uses LAM is more likely to be ready first)

I'll publish them as soon as they are ready, but it won't be as flawless of a PR as yours, for sure.

> 2. LL-SC only supports 32bit and 64bit, there may be problems with SLJIT_MOV_U8/U16 implementation

indeed, this is one of the particle reasons why I am interested on seeing how bad it will be.
I suspect that we would need to provide a "chooser" API, so users of sljit will be able to decide based on cost if they would want to get those smaller data sizes or not, maybe even return SLJIT_ERR_UNSUPPORTED for some.

BTW, one of the parts I am also having issues with, is the lack of zero extended modes for the instructions, specially since we only support unsigned types with atomics.

> 3. `SC.{W/D} rd, rj, si14` After storing the value in the `rd` register into the memory, the result `0/1 (1-success, 0-failure) `will be written into the `rd` register, which cannot be stored in the rd register retain the original content.
> 
> > also, what is the correct spelling for your name so it could be added to the README?
> 
> Mingtao Zhou

I was secretly hoping you would use some non latin version, as I seem to be the only one with a non-ascii character on mine, but glad otherwise.

Just to be sure, you are also using the "English" order here instead of the "Chinese" one with family name first, right?

Xiao-Tao:
There are 2 main reasons for not completing

1. LL-SC most supports 32/64bit and does not support non-aligned memory access. Too many instructions are used in the implementation of SLJIT_MOV_U8/U16. I think this may have problems in multi-thread scheduling
2. There is no multi-threaded concurrency test in 'test92', I have no way to ensure that my SLJIT_MOV_U8/U16 implementation is correct and efficient

https://github.com/Xiao-Tao/sljit-loongarch64/commit/4b83bc85de36f02dbb3676ef3cdb783f9252e8ff 
I have a patch for atomic instruction support under LoongArch64, I think this code is ugly in compatibility with SLJIT_MOV_U8/U16.
I haven't thought of a way to optimize this code so far, do you have any suggestions?
@carenas

zherczeg:
You can check the s390x code:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitNativeS390X.c#L3950

carenas:
@zherczeg: I don't think the problem is on how to implement, but if it will be "usable" and how to test that, considering all we have now is a test case that is not representative of real usage.

Indeed, I suspect that the s390x code might not perform as well as it could, and that the code using it would rather make sure NOT to need atomics that are smaller than 32bit.  Of course the advantage of s390x is that is CISC and strongly ordered, so it actually has the fastest atomics I'd seen as shown by my own threaded "test case" which is also not representative of real usage.

I am undecided, if we should have a `sljit_atomic_info(type, ptr)` that could be called to decide which size to use at runtime (just like it is done for `cmp`) and to mimic what is done by C11[1], or if the functions should just return `SLJIT_ERR_UNSUPPORTED` for types that are not supported without "emulation".

MIPS is an interesting case[2] and indeed with need a similar "fix" when implemented.

[1] https://en.cppreference.com/w/c/atomic/atomic_is_lock_free
[2] https://devblogs.microsoft.com/oldnewthing/20180410-00/?p=98475

zherczeg:
A single define for`sljit_has_cpu_feature` should be enough. The question is whether 8/16 bit atomics are emulated or supported. Note: WebAssembly, stdlib all supports 8/16 bit atomics, so emulation is necessary.

carenas:
the reason why I think an `sljit_atomic_info(type, ptr)` might be more appropriate is that there is also the possibility that 8/16 bit atomics are supported natively but only if they are 32-bit aligned like MIPS seems to "require".

zherczeg:
Bigger data size includes small data sizes, so if you have a 64 bit atomic op, you also have 8/16/32 as well, regardless of alignment. You do a 64 bit load, update the appropriate bytes (keep the others), and do a read/modify/write.

carenas:
not really, indeed that is another problem I see with the current API design and why I think it will be a better fit if we would include CAS like operations instead to map them better to the CPU own implementations (which in the case of loongarch requires the LAM extension)[1].

since we only have load and store, we can't do a full read/modify/write cycle without losing data for smaller sizes (which in the case of loongarch might even include 32bits because of the sign extension and even if it has support for 32bit atomics in hardware, but lets put aside for now)

if we change the API to include a caveat that says in a per CPU model: "anything below this bit width has to be aligned to a minimum bitsize" then the load would just ignore the rest of the bits (assumed to be zero as it is an unsigned type and "should" be zero extended to the size of the register) and there is no need to do any fancy masking/shifting and even reloading of data as the data will be written back padded with 0s.

I'd yet to find any code using atomics that would require byte aligned atomics, because as you saw with ARM when using LL/SC, packing atomics too closely causes serious problems.  I'd also only see code using LL/SC with super tiny critical code that do absolutely no memory access and most of the time, code compiled directly to do CAS like extensions even in RISC cpus.

[1] https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html#_amswapaddandorxormaxmindb_wd_ammaxmin_db_wudu

PR #172, opened 2023-06-09
loongarch: sljit_emit_atomic_load/store implementation

carenas:
assumes all atomics are at least 32bit aligned through setting SLJIT_ATOMIC_MIN_WIDTH.

only the bits that correspond to the operand size are taking into consideration with the rest pressumed to be 0.

operations with lower alignment are undefined.


test_src/sljitTest.c
```diff
@@ -11674,6 +11683,19 @@ static void test92(void)
...
+	sljit_emit_atomic_load(compiler, SLJIT_MOV_P, SLJIT_R0, SLJIT_R2);
+	sljit_emit_op1(compiler, SLJIT_MOV_P, SLJIT_S1, 0, SLJIT_R0, 0);
+	sljit_emit_op1(compiler, SLJIT_MOV_P, SLJIT_R1, 0, SLJIT_R0, 0);
+	sljit_emit_op1(compiler, SLJIT_MOV_P, SLJIT_R0, 0, SLJIT_R2, 0);
+	/* buf[38] */
+	sljit_emit_atomic_store(compiler, SLJIT_MOV_P | SLJIT_SET_ATOMIC_STORED, SLJIT_R0, SLJIT_R2, SLJIT_R1);
+	sljit_set_label(sljit_emit_jump(compiler, SLJIT_ATOMIC_NOT_STORED), label);
+	/* buf[39] */
+	sljit_emit_op1(compiler, SLJIT_MOV_P, SLJIT_MEM1(SLJIT_S0), 39 * sizeof(sljit_sw), SLJIT_S1, 0);
+
```
carenas, line 11698: this test will be moved to its right place with all indexes renumbered in a fixup!

test_src/sljitTest.c
```diff
@@ -11813,34 +11844,45 @@ static void test92(void)
...
+#endif
 	FAILED(((sljit_u32*)(buf + 33))[1] != 0xdeadbeef, "test92 case 47 failed\n");
-	FAILED(buf[34] != 0xffffffff, "test92 case 48 failed\n");
+	FAILED(((sljit_u32*)(buf + 34))[0] != 0xffffffff, "test92 case 48 failed\n");
 #endif /* SLJIT_64BIT_ARCHITECTURE */
 	FAILED(buf[35] != 1, "test92 case 49 failed\n");
+#if (!defined SLJIT_ATOMIC_MIN_WIDTH || SLJIT_ATOMIC_MIN_WIDTH <= 8)
 	FAILED(buf[36] != 0x11, "test92 case 50 (load) failed\n");
 	FAILED(((sljit_u8*)(buf + 37))[1] != buf[36], "test92 case 50 (store) failed\n");
+#endif /* SLJIT_ATOMIC_MIN_WIDTH <= 8 */
```
carenas, line 11880: this test will need rewriting once we better understand the constrains that NOT supporting byte addressable/aligned bytes in hardware causes.

when "emulation" is enabled to allow code that can pack atomics in bytes like this test does, I would expect failures, and indeed have a version that failed so often that I added a loop to it with 8 attempts and that is not included.

I also implemented similar code using 32bit and 64bit atomics (that are supported natively) and never saw it fail once, so I suspect that the most efficient type for an "atomic_flag" should be `int` instead of `char`, as was assumed by the original code (which never fails in x86. arm or s390x)

FWIW gcc on this architecture running linux shows:
```
sizeof(atomic_flag) == 1
#define __SIG_ATOMIC_TYPE__ int
#define __SIG_ATOMIC_MAX__ 0x7fffffff
#define __GCC_ATOMIC_INT_LOCK_FREE 2
```

test_src/sljitTest.c
```diff
@@ -11813,34 +11844,45 @@ static void test92(void)
...
 	FAILED(((sljit_u16*)(buf + 31))[1] != 51403, "test92 case 44 failed\n");
 	FAILED(buf[32] != 14876, "test92 case 45 failed\n");
+#endif /* SLJIT_ATOMIC_MIN_WIDTH <= 16 */
 #if (defined SLJIT_64BIT_ARCHITECTURE && SLJIT_64BIT_ARCHITECTURE)
+#if PADBYTE != 0
 	FAILED(((sljit_u32*)(buf + 33))[0] != 0x55555555, "test92 case 46 failed\n");
+#endif
 	FAILED(((sljit_u32*)(buf + 33))[1] != 0xdeadbeef, "test92 case 47 failed\n");
-	FAILED(buf[34] != 0xffffffff, "test92 case 48 failed\n");
+	FAILED(((sljit_u32*)(buf + 34))[0] != 0xffffffff, "test92 case 48 failed\n");
```
carenas, line 11874: this change "hides a bug" because loads and moves in 32bit values are sign extended to the full register.

there was an ugly patch that "solves" the issue (at least partially) but I suspect that since it has no external side effects it might be covered by the sljit constrain of ONLY using registers with the correct type from:

https://github.com/zherczeg/sljit/blob/818706a8027312831bc954fbcb0e0d66c20e0748/doc/overview.txt#L41-L44

zherczeg: No. ARM64 and x64 zero extends them.

sljit_src/sljitConfigInternal.h
```diff
@@ -786,6 +786,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_sw sljit_exec_offset(void* ptr);
 #define SLJIT_NUMBER_OF_SAVED_FLOAT_REGISTERS 12
 #define SLJIT_MASKED_SHIFT 1
 #define SLJIT_MASKED_SHIFT32 1
+#define SLJIT_ATOMIC_MIN_WIDTH 32
```
zherczeg, line 789: There should be no such constraint. When a 16 bit is loaded, and you have a 64 bit atomic rmw, you need to do the following steps:
- Align address to 64 bit
- Load 64 bit value
- Insert the 16 bit data in the correct bit position (xor can help)
- RMW the 64 bit value

You can do this with any data type.

test_src/sljitTest.c
carenas: but doing zero extension is a feature which is not supported by the instructions we are using in loongarch (which signs extends by default, just like MIPS did at least when that comment was added).

the important part though, is that if coded correctly that sign extension doesn't need to be visible, and is not in this case as both atomic loads and stores ONLY see the lower 32bit part which is all sljit warranties to be accurate.

sljit_src/sljitConfigInternal.h
carenas: you are correct if we will be implementing an atomic rmw operation direcly, but we are not.

the load has to return a valid value for the type used and so, all non relevant bits from the wider register that are loaded from memory with the correct alignment and that include the smaller type had to be discarded.

by the time the store comes, you can't write the value back to memory or will risk wiping several unrelated bytes, so then a second load must be done to insert the smaller type in the right place before an atomic store can be done.

all that is very expensive, specially when you have a very simple RISC CPU that doesn't have a cmpxchg like CAS instruction either.

I suspect the "problem" is also presented at least on MIPS and the "solution" was to put in the specification what I put in code here.

you can LL/SC any type you want as far as it is aligned to the minimum width that is supported by the instructions; not sure if the data pulled from memory gets masked during transfer like ARM seems to be doing and therefore it is technically a 4 byte aligned type where 2 or 3 bytes get discarded.

if our API would implement instead a "cmpxchg" like interface, then it could be implemented just like you mention.

note that loongarch does support CAS like RMW operations in hardware with its LAM extension, so it would be even easier to implement atomic add, xor and the rest of the usual RMW operations.

zherczeg: You can add a flag to cpuinfo to tell that 8/16 bit updates are not efficient. Still, the emulation can use instructions which are not available otherwise, so it will be still better that the user can do.

carenas: there is also the counterpart of that, which is to (like I am proposing here) make the limitation of the CPU visible at compile time and optimize for the fast use case by default.

of course, will add a way to enable it (had some WIP code that does using an `SLJIT_ATOMIC_EMULATION` define and that by now has been mostly implemented for `SLJIT_MOV_U16` only) and that I could share for feedback once it actually works.  The end result will probably look like the [initial proposal ](https://github.com/Xiao-Tao/sljit-loongarch64/tree/loongarch-atomic-dev)from Mingtao, only uglier.

I frankly suspect that in real world scenarios all atomics will be aligned correctly and have enough padding regardless and so the added code from the emulation is just dead weight.  I was thinking on splitting the change so `SLJIT_ATOMIC_WIDTH` could be defined independently and then implement `SLJIT_ATOMIC_EMULATION` also for s390x, which unlike loongarch is likely to be enabled by default, so it can be tested independently.

zherczeg: We have an infrastructure for checking cpu features. And only two options are here: 8/16 bit is available or emulated. From user perspective, it is hard to use a "bit width" define, because it can have any value.

carenas: while implementing the atomic store for SLJIT_MOV_U8, I found another scenario where the test will infloop.  LL/SC is IMHO too fragile and it would seem that with most CPUs (even the traditionally RISC ones) providing CAS it would seem worth looking if an API more in that line might be worth exploring.

Xiao-Tao: > There should be no such constraint. When a 16 bit is loaded, and you have a 64 bit atomic rmw, you need to do the following steps:
> 
> * Align address to 64 bit
> * Load 64 bit value
> * Insert the 16 bit data in the correct bit position (xor can help)
> * RMW the 64 bit value
> 
> You can do this with any data type.

LoongArch64 requires **16** instructions (maybe optimized to 12) when implementing RMW operations compatible with `SLJIT_MOV_U8`, and `LL-SC ` should use fewer instructions.

zherczeg: 64 bit byte reverse on systems without byte reverse instruction takes more. The thing is, if a code (such as WebAssembly) wants to do an 8 bit atomic update, it does not really matter how many instructions are needed, the job needs to be done somehow. And cpu specific instruction sequences are always shorter than using sljit instructions for emulation.

carenas: > 4 bit byte reverse on systems without byte reverse instruction takes more

that is not a fair comparison, because there is no such thing as restrictions to the number of instruction that you can run in any CPU for that operation, but for LL/SC (which is conceptually one single instruction that can be used to implement "RMW") apparently LoongArch has, and for my reading it would seem RISCV does as well.

the end result is that in those cases the reservation gets always invalidated, and then you are left with a process that infloops.

for all practical reasons the compiler doesn't let you emit a load independently of a store, and deduces the type of operation used, opting for the CAS like equivalent, for example the following C11 code:

```
char step = 32;
volatile char  m = 127;
atomic_char c;
c = (c > m) ? 0 : c + step;
```

will be most likely compiled into something closer to what you get now for loongarch with emulation disabled, and ensure that `c` is aligned to 32bit, or even better will detect LAM is available and then it can use `ammax.wu` for the comparison and `amadd.w` for the increment.

zherczeg: An option could be to create a cmpxchg for all architectures, and create an optional LL/SC based variant, which has limited support.

PR #177, opened 2023-08-02
Fix incorrect branch offset calculation

ilink-splunk:
This now uses the same offset calculation as sljit_generate_code. Before, if you had an address that was 16 bytes off from the maximum, it would overflow the 26 bit signed int available to the "bl" instruction. In other words, it would choose "bl" when it should have chosen "blr". This typically caused a crash because the jump was to an invalid address.


zherczeg (approved):
LGTM

PR #184, opened 2023-08-20
simd: prevent SIGILL with armv6

carenas:
NEON was introduced as an optional feature with armv7, so disable it for earlier versions.


zherczeg (approved):
LGTM

zherczeg:
There should be some compiler define for NEON.

carenas:
there is as shown in #144, but almost all popular armv7 (except maybe tegra 2) in use include support even if optional

PR #187, opened 2023-08-21
arm: provide emulation for SLJIT_CTZ in ARMv6

carenas:
Avoids SIGILL when running in older CPUs than armv7 (ex: Raspberry PI < 2)

Tested with QEMU using a softfp userland, but would work as well with hardfp (FPU is optional but common with those CPUs, even if it is not NEON).

Thumb2 with `march=armv6t2` not addressed yet

### DIFF
--- a/sljit_src/sljitNativeARM_32.c
+++ b/sljit_src/sljitNativeARM_32.c
@@ -95,13 +95,13 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define AND		0xe0000000
 #define B		0xea000000
 #define BIC		0xe1c00000
+#define BKPT		0xe1200070
 #define BL		0xeb000000
 #define BLX		0xe12fff30
 #define BX		0xe12fff10
 #define CLZ		0xe16f0f10
 #define CMN		0xe1600000
 #define CMP		0xe1400000
-#define BKPT		0xe1200070
 #define EOR		0xe0200000
 #define LDR		0xe5100000
 #define LDR_POST	0xe4100000
@@ -115,7 +115,6 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define ORR		0xe1800000
 #define PUSH		0xe92d0000
 #define POP		0xe8bd0000
-#define RBIT		0xe6ff0f30
 #define REV		0xe6bf0f30
 #define REV16		0xe6bf0fb0
 #define RSB		0xe0600000
@@ -168,6 +167,7 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 /* Arm v7 specific instructions. */
 #define MOVT		0xe3400000
 #define MOVW		0xe3000000
+#define RBIT		0xe6ff0f30
 #endif
 
 #if (defined SLJIT_CONFIG_ARM_V6 && SLJIT_CONFIG_ARM_V6)
@@ -994,7 +994,6 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 
 	case SLJIT_SIMD_REGS_ARE_PAIRS:
 	case SLJIT_HAS_CLZ:
-	case SLJIT_HAS_CTZ:
 	case SLJIT_HAS_ROT:
 	case SLJIT_HAS_CMOV:
 	case SLJIT_HAS_REV:
@@ -1002,6 +1001,13 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 	case SLJIT_HAS_COPY_F32:
 	case SLJIT_HAS_COPY_F64:
 		return 1;
+
+	case SLJIT_HAS_CTZ:
+#if defined(SLJIT_CONFIG_ARM_V6) && SLJIT_CONFIG_ARM_V6
+		return 2;
+#else
+		return 1;
+#endif /* SLJIT_CONFIG_ARM_V6 */
 
 	default:
 		return 0;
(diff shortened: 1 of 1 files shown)


sljit_src/sljitNativeARM_T2_32.c
```diff
@@ -509,6 +509,9 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 {
 	switch (feature_type) {
+	case SLJIT_HAS_CTZ:
```
zherczeg, line 512: This should be armv6, and should be somewhere at the end.

sljit_src/sljitNativeARM_32.c
```diff
@@ -1502,8 +1507,16 @@ static SLJIT_INLINE sljit_s32 emit_single_op(struct sljit_compiler *compiler, sl
 	case SLJIT_CTZ:
 		SLJIT_ASSERT(!(flags & INV_IMM) && !(src2 & SRC2_IMM));
 		SLJIT_ASSERT(src1 == TMP_REG1 && !(flags & ARGS_SWAPPED));
+#if defined(SLJIT_CONFIG_ARM_V7) && SLJIT_CONFIG_ARM_V7
 		FAIL_IF(push_inst(compiler, RBIT | RD(dst) | RM(src2)));
 		return push_inst(compiler, CLZ | RD(dst) | RM(dst));
+#else
+		FAIL_IF(push_inst(compiler, RSBS | RD(dst) | RN(src2)));
```
zherczeg, line 1514: Is this RSB? Setting flags are not part of the instruction, should be added after a `|`

sljit_src/sljitNativeARM_32.c
```diff
@@ -1502,8 +1507,16 @@ static SLJIT_INLINE sljit_s32 emit_single_op(struct sljit_compiler *compiler, sl
...
+		FAIL_IF(push_inst(compiler, AND | RD(dst) | RN(dst) | RM(src2)));
+		FAIL_IF(push_inst(compiler, CLZ | RD(dst) | RM(dst)));
+		FAIL_IF(push_inst(compiler, RSBS | RD(dst) | RN(dst) | 31));
+		return push_inst(compiler, 0x43a00020 | RD(dst)); /* movmi dst, #32 */
```
zherczeg, line 1518: The normal instruction should be used, and the condition codes should be changed. There are some places for this in the code.

sljit_src/sljitNativeARM_32.c
```diff
@@ -1502,8 +1507,16 @@ static SLJIT_INLINE sljit_s32 emit_single_op(struct sljit_compiler *compiler, sl
 	case SLJIT_CTZ:
 		SLJIT_ASSERT(!(flags & INV_IMM) && !(src2 & SRC2_IMM));
 		SLJIT_ASSERT(src1 == TMP_REG1 && !(flags & ARGS_SWAPPED));
+#if defined(SLJIT_CONFIG_ARM_V7) && SLJIT_CONFIG_ARM_V7
 		FAIL_IF(push_inst(compiler, RBIT | RD(dst) | RM(src2)));
 		return push_inst(compiler, CLZ | RD(dst) | RM(dst));
+#else
+		FAIL_IF(push_inst(compiler, RSBS | RD(dst) | RN(src2)));
+		FAIL_IF(push_inst(compiler, AND | RD(dst) | RN(dst) | RM(src2)));
```
zherczeg, line 1515: What happens if src2 == dst?

zherczeg (changes requested):
There was a code for this when armv5 support was removed, maybe it can be restored.

sljit_src/sljitNativeARM_T2_32.c
carenas: Since the end has a `default: return 0`, thought it was more appropriate for it to be at the top to help distinguish it.

About it being armv6, I have to admit I was surprised to find out that `CTZ` was returning 1 originally, after all it is not a direct translation and requires at least 2 instructions when `RBIT` is available; wouldn't that count as "emulation"?

sljit_src/sljitNativeARM_32.c
carenas: it is a second encoding[1] for `RSB` which uses an immediate as one of the parameters and that was previously unused.  I enabled the "reporting" bit by default but I see why you would rather have that set explicitly at invocation time, and probably only to the one invocation that matters, eventhough it probably doesn't make any difference performance/logic wise in this context.

would it better to use `RSBI` as a mnemonic?

[1] https://developer.arm.com/documentation/ddi0597/2023-03/Base-Instructions/RSB--RSBS--immediate---Reverse-Subtract--immediate--

carenas: except it was being used through `SRC2_IMM` but I wasn't familiar enough with the code, eitherway code has been discarded for now

sljit_src/sljitNativeARM_T2_32.c
zherczeg: Well, 2 should means: emulation is costly, so if you can avoid this instruction by a different code path, avoid it. The two instruction sequence is simple enough.

However, I am open to use a better description system, maybe use define constants instead of 0-2 return values. If you have time, patches are welcome!

zherczeg:
According to my documentation, RBIT is ARMv6T2, ARMv7 instruction, so Thumb code should not be affected.

sljit_src/sljitNativeARM_T2_32.c
carenas: one way to do that is to return the number of native instructions that are required by emulation so the cost can be also communicated back to the callers in an objective way instead of being implicit.

zherczeg: I am not sure that helps too much from developer perspective. I prefer something which refers to complexity. The question is: am I recommended to use this, or should avoid it when possible.

PR #189, opened 2023-08-22
arm: restrict NEON code in sljit_emit_fset{32,64}

carenas:
vmov.f32(NEON) and vmov.f64(VFPv3) will trigger SIGILL otherwise in test93 when running in an older CPU (ex: arm1136 in Raspberry PI < 2)

### DIFF
--- a/sljit_src/sljitNativeARM_32.c
+++ b/sljit_src/sljitNativeARM_32.c
@@ -2700,8 +2700,10 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fop2(struct sljit_compiler *compil
 SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fset32(struct sljit_compiler *compiler,
 	sljit_s32 freg, sljit_f32 value)
 {
+#if defined(__ARM_NEON) && __ARM_NEON
 	sljit_u32 exp;
 	sljit_ins ins;
+#endif /* NEON */
 	union {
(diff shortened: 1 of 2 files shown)


sljit_src/sljitNativeARM_32.c
```diff
@@ -2712,14 +2710,16 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fset32(struct sljit_compiler *comp
 
 	u.value = value;
 
+#if defined(SLJIT_CONFIG_ARM_V7) && SLJIT_CONFIG_ARM_V7
 	if ((u.imm << (32 - 19)) == 0) {
-		exp = (u.imm >> (23 + 2)) & 0x3f;
+		sljit_u32 exp = (u.imm >> (23 + 2)) & 0x3f;
```
zherczeg, line 2715: This form is not used in the project. Add an ifdef around the definition.

zherczeg (changes requested):

sljit_src/sljitNativeARM_32.c
carenas: FWIW, locally scoped variables are valid C89

zherczeg (approved):
LGTM

zherczeg:
Is __ARM_NEON defined by compilers by default?

carenas:
It is part of ACLE (the ARM standard of how features are defined by compilers) and is therefore defined if the architecture targeted (through `-march`) includes NEON support.

PR #190, opened 2023-08-25
arm: fix not aligned register pair thumb2 operations

carenas:
Mask up to 12bits for the offset to avoid truncation. Triggering in test79 case 21 when running in QEMU, somehow


zherczeg (approved):
LGTM

PR #191, opened 2023-08-27
build: fix warnings with gcc12 and related improvements

carenas:
Mostly simple "fixes" to avoid warnings, but also some improvements to the Makefile to help with future related work.

The new makefile options have defaults that keep the old behaviour.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -15,8 +15,20 @@ ifndef EXTRA_LDFLAGS
 EXTRA_LDFLAGS=
 endif
 
+ifndef EXTRA_LIBS
+EXTRA_LIBS=
+endif
+
+ifndef OPT_FLAGS
+OPT_FLAGS = -O2
+endif
+
+ifndef WERROR
+WERROR = -Werror
+endif
+
 CPPFLAGS = $(EXTRA_CPPFLAGS) -Isljit_src
-CFLAGS += -O2 -Wall -Wextra -Wconversion -Wsign-compare -Werror
+CFLAGS += $(OPT_FLAGS) -Wall -Wextra -Wconversion -Wsign-compare -Wdeclaration-after-statement $(WERROR)
 REGEX_CFLAGS += $(CFLAGS) -fshort-wchar
 LDFLAGS = $(EXTRA_LDFLAGS)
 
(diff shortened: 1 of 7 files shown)


zherczeg:
Is there a negative effect of putting the ldflags at the end? Unusual for me.

carenas:
> Is there a negative effect of putting the ldflags at the end? Unusual for me.

none that I know of, but you are correct it is unusual so better drop it for now

GNUmakefile
```diff
@@ -67,33 +79,33 @@ $(BINDIR)/regexJIT.o : $(REGEXDIR)/regexJIT.c $(BINDIR)/.keep $(SLJIT_HEADERS) $
 	$(CC) $(CPPFLAGS) $(REGEX_CFLAGS) -c -o $@ $(REGEXDIR)/regexJIT.c
 
 $(BINDIR)/sljit_test: $(BINDIR)/.keep $(BINDIR)/sljitMain.o $(TESTDIR)/sljitTest.c $(SRCDIR)/sljitLir.c $(SLJIT_LIR_FILES) $(SLJIT_HEADERS) $(TESTDIR)/sljitConfigPre.h $(TESTDIR)/sljitConfigPost.h
-	$(CC) $(CPPFLAGS) -DSLJIT_HAVE_CONFIG_PRE=1 -I$(TESTDIR) $(CFLAGS) $(LDFLAGS) $(BINDIR)/sljitMain.o $(TESTDIR)/sljitTest.c $(SRCDIR)/sljitLir.c -o $@ -lm -lpthread
+	$(CC) -pthread $(CPPFLAGS) -DSLJIT_HAVE_CONFIG_PRE=1 -I$(TESTDIR) $(CFLAGS) $(LDFLAGS) $(BINDIR)/sljitMain.o $(TESTDIR)/sljitTest.c $(SRCDIR)/sljitLir.c -o $@ -lm $(EXTRA_LIBS)
```
zherczeg, line 82: What is the difference?

carenas: When using `-pthread` then the coprocessor and compiler are also aware that the object needs to be "thread safe" (ex:  `-D_REENTRANT` gets defined).

also it doesn't fail to link if there is no -lpthread (ex: OpenWRT); most libc have the pthread functions included in the main library and optionally use a "stub pthread" library.

GNUmakefile
```diff
@@ -15,8 +15,20 @@ ifndef EXTRA_LDFLAGS
...
+OPT_FLAGS = -O2
+endif
+
+ifndef WERROR
+WERROR = -Werror
+endif
+
 CPPFLAGS = $(EXTRA_CPPFLAGS) -Isljit_src
-CFLAGS += -O2 -Wall -Wextra -Wconversion -Wsign-compare -Werror
+CFLAGS += $(OPT_FLAGS) -Wall -Wextra -Wconversion -Wsign-compare -Wdeclaration-after-statement $(WERROR)
```
zherczeg, line 31: Is `-Wdeclaration-after-statement` widely used? Maybe all warnings should go to a macro as well, so people can define it in any way they want.

On the other hand, this makefile does not produce any libraries, so never used by any real project. So complicating it is not worth it. I am not sure.

carenas: I think it is worth adding, unless we are no longer aiming for C89 compatibility anymore as it is only supported starting with C99

GNUmakefile
zherczeg: Normally libraries are defined at the end, since it is used for checking unresolved references. If pthread is a real library, this might not work.

carenas: It will always work, unless your gcc doesn't know how to handle the `-pthread` flag.  Some really old versions of clang (google seems to think that older than 8) didn't knew how to handle that flag, but hopefully nobody using this still cares about those.

When `-pthread` is provided, and gcc is driving the linking then -lpthread is added as needed.

`-pthread` has been part of gcc since 2.7.0 (literally from last century) and glibc only created the "stub" library with release 2.34 (end of 2021) so that is how it worked until then.

zherczeg: I worry more about other compilers like Intel C compiler. Defining both early and late could work?

zherczeg: Btw as far as I know all command line arguments are processed by the compiler before doing anything, so pthread at the end should have its side effect.

carenas: FWIW for a hold of clang 3.4.2 and it recognized the `-pthread` option and linked correctly as if `-lpthread` was provided, so maybe Google was referring to an old version of Apple clang which is even less likely to be relevant.

For the sake of expediency, I have removed this change though but still think is worth considering, after all even the non opensource compilers are based on clang nowadays.

sljit_src/sljitNativeRISCV_common.c
```diff
@@ -1068,7 +1068,7 @@ static sljit_s32 emit_clz_ctz(struct sljit_compiler *compiler, sljit_s32 op, slj
 {
 	sljit_s32 is_clz = (GET_OPCODE(op) == SLJIT_CLZ);
 #if (defined SLJIT_CONFIG_RISCV_64 && SLJIT_CONFIG_RISCV_64)
-	sljit_ins word = (op & SLJIT_32) >> 5;
+	sljit_ins word = (sljit_u32)(op & SLJIT_32) >> 5;
```
zherczeg, line 1071: sljit_u32 -> sljit_ins 
There are similar cases below.

zherczeg (changes requested):
Good patch, only a minor thing remained.

GNUmakefile
zherczeg: Yes, we can do this change if it does not negatively impact systems where pthreds is just a simple library without any special casing. I had issues with using libraries too early and getting unresolved references in the past.

zherczeg (approved):
LGTM

PR #192, opened 2023-08-27
s390x: sanitize warnings with clang 15

carenas:
Mostly to avoid UB with shifts of negative numbers.

### DIFF
--- a/sljit_src/sljitNativeS390X.c
+++ b/sljit_src/sljitNativeS390X.c
@@ -454,7 +454,7 @@ static SLJIT_INLINE sljit_ins disp_s20(sljit_s32 d)
 	SLJIT_ASSERT(is_s20(d));
 
 	sljit_uw dh = (d >> 12) & 0xff;
-	sljit_uw dl = (d << 8) & 0xfff00;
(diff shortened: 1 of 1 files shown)


carenas:
while it is true that only 2 compilers are supported and both do the right thing with this code (even if it is technically "undefined behaviour"), there is no way to know if future versions will do as well (specially considering how aggressively `clang` has been optimizing in UB in recent versions).

it will also avoid crashes if built with `-fsanitize=undefined -fsanitize-trap=all` which is also more fashionable lately.

zherczeg:
Isn't it enough to cast `(sljit_uw)` without & operation?

carenas:
> Isn't it enough to cast `(sljit_uw)` without & operation?

only for gcc; clang will stil complain that the shifted value doesn't fit in the target type since the higher word now will need to discard all those sign extended bits.

zherczeg:
The `uw` does not extend sign, it kills it. Can clang explain why it exactly complains or is it a bug in there?

carenas:
removed all unnecessary masks

zherczeg (approved):
LGTM

PR #193, opened 2023-08-29
ppc: fix SLJIT_REV for 64bit little endian

carenas:
```
[carenasgm@gcc120 sljit.base]$ bin/sljit_test 

test87 case 3 failed
SLJIT tests: 1 (0%) tests FAILED on PowerPC 64bit (little endian + unaligned) (with fpu)
```
where:

```
test87 case 3 expected 0x0807060504030201 got 0x0403020108070605
test87 case 5 expected 0x8877665544332211 got 0x4433221188776655
```

### DIFF
--- a/sljit_src/sljitNativePPC_common.c
+++ b/sljit_src/sljitNativePPC_common.c
@@ -150,6 +150,9 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define BCx		(HI(16))
 #define BCCTR		(HI(19) | LO(528) | (3 << 11))
 #define BLR		(HI(19) | LO(16) | (0x14 << 21))
+#if defined(_ARCH_PWR10) && _ARCH_PWR10
+#define BRD		(HI(31) | LO(187))
+#endif /* POWER10 */
 #define CNTLZD		(HI(31) | LO(58))
 #define CNTLZW		(HI(31) | LO(26))
 #define CMP		(HI(31) | LO(0))
@@ -184,6 +187,9 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define LD		(HI(58) | 0)
 #define LFD		(HI(50))
 #define LFS		(HI(48))
+#if defined(_ARCH_PWR7) && _ARCH_PWR7
+#define LDBRX		(HI(31) | LO(532))
+#endif /* POWER7 */
 #define LHBRX		(HI(31) | LO(790))
 #define LWBRX		(HI(31) | LO(534))
 #define LWZ		(HI(32))
@@ -222,6 +228,9 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define SRD		(HI(31) | LO(539))
 #define SRW		(HI(31) | LO(536))
 #define STD		(HI(62) | 0)
+#if defined(_ARCH_PWR7) && _ARCH_PWR7
+#define STDBRX		(HI(31) | LO(660))
+#endif /* POWER7 */
 #define STDU		(HI(62) | 1)
 #define STDUX		(HI(31) | LO(181))
 #define STFD		(HI(54))
@@ -266,11 +275,15 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #endif /* SLJIT_CONFIG_PPC_32 */
 
 #if (defined SLJIT_LITTLE_ENDIAN && SLJIT_LITTLE_ENDIAN)
-#define TMP_MEM_OFFSET_LOW TMP_MEM_OFFSET
-#define TMP_MEM_OFFSET_HI (TMP_MEM_OFFSET + sizeof(sljit_s32))
+#define TMP_MEM_OFFSET_LO	(TMP_MEM_OFFSET)
+#define TMP_MEM_OFFSET_HI	(TMP_MEM_OFFSET + sizeof(sljit_s32))
+#define LWBRX_FIRST_REG		S(TMP_REG1)
+#define LWBRX_SECOND_REG	S(dst)
 #else /* !SLJIT_LITTLE_ENDIAN */
-#define TMP_MEM_OFFSET_LOW (TMP_MEM_OFFSET + sizeof(sljit_s32))
-#define TMP_MEM_OFFSET_HI TMP_MEM_OFFSET
+#define TMP_MEM_OFFSET_LO	(TMP_MEM_OFFSET + sizeof(sljit_s32))
+#define TMP_MEM_OFFSET_HI	(TMP_MEM_OFFSET)
+#define LWBRX_FIRST_REG		S(dst)
+#define LWBRX_SECOND_REG	S(TMP_REG1)
 #endif /* SLJIT_LITTLE_ENDIAN */
 
 #if (defined SLJIT_INDIRECT_CALL && SLJIT_INDIRECT_CALL)
@@ -673,7 +686,12 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 		/* Available by default. */
 		return 1;
 #endif
-
+	case SLJIT_HAS_REV:
+#if defined(_ARCH_PWR10) && _ARCH_PWR10
+		return 1;
+#else /* !POWER10 */
+		return 2;
+#endif /* POWER10 */
 	/* A saved register is set to a zero value. */
 	case SLJIT_HAS_ZERO_REGISTER:
 	case SLJIT_HAS_CLZ:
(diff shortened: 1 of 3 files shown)


sljit_src/sljitNativePPC_common.c
```diff
@@ -1421,15 +1425,27 @@ static sljit_s32 emit_rev(struct sljit_compiler *compiler, sljit_s32 op,
 
 		if (memw != 0) {
 			FAIL_IF(push_inst(compiler, SLWI_W(memw) | S(offs_reg) | A(TMP_REG2)));
+#if (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64)
+			if (op == SLJIT_REV && !is_32 && TMP_MEM_START)
+				FAIL_IF(push_inst(compiler, ADDI | D(TMP_REG2) | A(TMP_REG2) | IMM(TMP_MEM_START)));
```
zherczeg, line 1430: Don't you need to add this offset when `memw == 0`

zherczeg (changes requested):
Something I don't understand how reverse has any effect on endianness. It swaps the lower and upper part, and write them back in reversed order. In theory endianness does not matter.

carenas:
> I don't understand how reverse has any effect on endianness

It is because of the implementation that reads/writes 4 bytes at a time, hence needs to know if the next word to operate on is forward or backward in memory.

zherczeg:
No. It reads two words in the 8 byte case, and swaps them (and their values). It should work regardless of endianness.

carenas:
POWER didn't have a register to register `bswap` instruction until Power ISA v.3.1, which is supported by POWER10, so the implementation does the swapping while writting or reading from memory (even when both parameters are registers).

Added support for that instruction so it will be easier to see the differences.  Notice that the instructions that do the swapping only work with 4 bytes at a time as explained earlier.

carenas:
> (ABCD EFGH) its reversed form is (HGFE DCBA) regardless of the endianness of the cpu

and that is why I made the Power10 implementation, that does one byteswap AND one memory write and works the same regardless of endianess.

> So why do we need any specialization here?

because the original implementation uses 4 byte operations with an 8 byte pointer and therefore happened to work fine in Big Endian because the address for the lower 4 byte integer and the 8 byte one is the same, but that assumption breaks with Little Endian.

I am not the author of the original code, but there was logic already to figure out which 4 byte integer were low/hi and that was used in a somehow convoluted way for the register to register operation, so I assumed the additional logic was just forgotten for the other operations and therefore wrote the proposed code to fill that gap; but I woud rather had implemented swapping the 4 byte integers in the register (which will require 1 extra instruction but only in "little endian") and defintely not using the `STWBRX/LWBRX` instructions when not dealing with memory.

zherczeg:
The "extra instruction" raised my attention, because in theory, we should not need it. I might have an idea about the problem and the solution. Moving data between registers and memory is the problem, not the endianness. E.g:
```
A = low_word
B = high_word
big_endian: (B << 32) | A
little_endian: (A << 32) | B
```
Therefore the dst/tmp (load) and tmp1/tmp2 (store) usage must be swapped to tmp/dst and tmp2/tmp1 in little endian. So the code is the same, but uses different register order. Perhaps we could create local variables for the register indicies (both for big and little endian), and let the compiler use the correct constant. Or use defines, which is a bit more complex.

sljit_src/sljitNativePPC_common.c
```diff
@@ -1371,12 +1380,16 @@ static sljit_s32 emit_rev(struct sljit_compiler *compiler, sljit_s32 op,
 
 #if (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64)
 		if (!is_32) {
+#if defined(_ARCH_PWR10) && _ARCH_PWR10
+			return push_inst(compiler, BRD | S(src) | A(dst));
+#else
```
zherczeg, line 1385: Missing comment

sljit_src/sljitNativePPC_common.c
```diff
@@ -1452,16 +1477,23 @@ static sljit_s32 emit_rev(struct sljit_compiler *compiler, sljit_s32 op,
 #if (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64)
 	if (!is_32) {
 		if (dst & SLJIT_MEM) {
+#if defined(_ARCH_PWR7) && _ARCH_PWR7
+			return push_inst(compiler, STDBRX | S(src) | A(mem) | B(offs_reg));
+#else
```
zherczeg, line 1482: Please fix all of them

sljit_src/sljitNativePPC_common.c
```diff
@@ -1424,12 +1437,24 @@ static sljit_s32 emit_rev(struct sljit_compiler *compiler, sljit_s32 op,
 			offs_reg = TMP_REG2;
 		}
 #if (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64)
-	} else if (memw > 0x7fff7fffl || memw < -0x80000000l) {
+#if !(defined(_ARCH_PWR7) && _ARCH_PWR7)
+		if ((op == SLJIT_REV) && !is_32 && (TMP_MEM_START != 0))
+			FAIL_IF(push_inst(compiler, ADDI | D(offs_reg) | A(offs_reg) | IMM(TMP_MEM_START)));
```
zherczeg, line 1442: The question I raised still stands.

big endian:
`swap[base,offset] = reg; offset += 4; tmp =reg >> 32; swap[base,offset] = tmp`
little endian:
`tmp =reg >> 32; swap[base,offset] = tmp; offset += 4; swap[base,offset] = reg`

Therefore we don't need to add an extra instruction for little endian, just reorder the operations. Is there a reason why this cannot be done?

zherczeg (changes requested):

sljit_src/sljitNativePPC_common.c
carenas: Just wish it would look as clean as the MIPS code, but at least most of it won't be executed by default since POWER7 is more than 10 years old and targeted by compilers by default

zherczeg: I would not mind moving to a newer minimum requirement. I need to figure out how to use this with qemu. I don't even understand the difference between powerpc and power.

carenas: the additional instructions are enabled by the compiler, so first need to compile (assuming `gcc` and POWER10 for example) with `-mcpu=power10`

`qemu` has a `-cpu` parameter that indicates the CPU type it is to emulate so there it would be `-cpu power10` assuming you have a recent enough version.

you can get the full list with `-cpu help`

sljit_src/sljitNativePPC_common.c
```diff
@@ -1452,16 +1469,37 @@ static sljit_s32 emit_rev(struct sljit_compiler *compiler, sljit_s32 op,
...
+#if defined(SLJIT_LITTLE_ENDIAN) && SLJIT_LITTLE_ENDIAN
+		FAIL_IF(push_inst(compiler, LWBRX | S(TMP_REG1) | A(mem) | B(offs_reg)));
+		FAIL_IF(push_inst(compiler, ADDI | D(TMP_REG2) | A(offs_reg) | IMM(SSIZE_OF(s32))));
+		FAIL_IF(push_inst(compiler, LWBRX | S(dst) | A(mem) | B(TMP_REG2)));
+		return push_inst(compiler, RLDIMI | S(TMP_REG1) | A(dst) | RLDI_SH(32) | RLDI_MB(0));
+#else /* !SLJIT_LITTLE_ENDIAN */
 		FAIL_IF(push_inst(compiler, LWBRX | S(dst) | A(mem) | B(offs_reg)));
 		FAIL_IF(push_inst(compiler, ADDI | D(TMP_REG2) | A(offs_reg) | IMM(SSIZE_OF(s32))));
 		FAIL_IF(push_inst(compiler, LWBRX | S(TMP_REG1) | A(mem) | B(TMP_REG2)));
 		return push_inst(compiler, RLDIMI | S(TMP_REG1) | A(dst) | RLDI_SH(32) | RLDI_MB(0));
```
zherczeg, line 1500: It looks like the last two instructions are the same, so no guard is needed.

zherczeg (changes requested):
Nice patch! Only a minor change is needed.

sljit_src/sljitNativePPC_common.c
carenas: I intentionally avoided using the same code as it was IMHO less intuitive, but implemented as requested

sljit_src/sljitNativePPC_common.c
```diff
@@ -181,9 +184,12 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define FRSP		(HI(63) | LO(12))
 #define FSUB		(HI(63) | LO(20))
 #define FSUBS		(HI(59) | LO(20))
-#define LD		(HI(58) | 0)
+#define LD		(HI(58))
```
zherczeg, line 187: Why `| 0` is removed? As far as I remember the last two bits define the instruction types for 64 bit load/stores, and that 0 wants to represent it.

zherczeg (commented):
Just one question.

zherczeg (approved):
LGTM

PR #194, opened 2023-08-30
atomic: avoid unused checks if not implemented

carenas:
mainly to avoid unused function warnings with clang on the architectures that are yet to be implemented


zherczeg (approved):
LGTM

PR #195, opened 2023-08-30
mips: add missing ADJUST_LOCAL_OFFSET for sljit_emit_select

carenas:


carenas:
> We should have a test for that

It did trigger an error with test54 case 14 at least on MIPS III 32bit big endian, I checked the code from all other architectures and this was the only one missing it for that function.

zherczeg (approved):
LGTM

carenas:
> I cannot test mips32 anymore, because qemu cannot do float comparison properly and lots of tests are failing

After looking at if for a while, I don't think it is fundamentally a qemu issue, but real bugs.

The first problem is that our FPU detection is only looking at the flag that indicates FPU registers are 64bit so it fails to work with the original MIPS32 and is inconsistent with the code that assumes FPU registers are 32bit (See: the WIP [branch](https://github.com/carenas/sljit/tree/mips32fpu) for a fix).

The bigger problem is that the code assumes FPU registers are always 32bit, but that is not always the case since MIPSr2 (see Section 6.4.1 from the [documentation](https://s3-eu-west-1.amazonaws.com/downloads-mips/documents/MD00082-2B-MIPS32INT-AFP-06.01.pdf)), and qemu is setting the optional flag (StatusFR) that indicates its registers are 64bit and it does NOT want to use them in legacy mode, hence why manipulating doubles with a pair of registers break.

I have a [WIP](https://github.com/carenas/sljit/tree/test95v2) version of the copysign solution from #199 that includes code for that as a POC, but you will need to modify it to be enable with SLJIT_MIPS_REV >= 2 for testing.

zherczeg:
They couldn't make this more complex. The future of mips has lots of questions, so spending a lot of effort on this code is not my priority. Maybe it will be removed from the code at some point.

PR #196, opened 2023-08-31
mips: fix implementation of sljit_emit_fset64 in 32bit mips (BE)

carenas:
This fixes errors in test93 that are interestingly enough only showing for big endian.

### DIFF
--- a/sljit_src/sljitNativeMIPS_32.c
+++ b/sljit_src/sljitNativeMIPS_32.c
@@ -62,8 +62,13 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fset64(struct sljit_compiler *comp
 	if (u.imm[1] != 0)
 		FAIL_IF(load_immediate(compiler, DR(TMP_REG2), u.imm[1]));
 
+#if defined(SLJIT_LITTLE_ENDIAN) && SLJIT_LITTLE_ENDIAN
 	FAIL_IF(push_inst(compiler, MTC1 | (u.imm[0] != 0 ? T(TMP_REG1) : TA(0)) | FS(freg), MOVABLE_INS));
 	return push_inst(compiler, MTC1 | (u.imm[1] != 0 ? T(TMP_REG2) : TA(0)) | FS(freg) | (1 << 11), MOVABLE_INS);
+#else
+	FAIL_IF(push_inst(compiler, MTC1 | (u.imm[1] != 0 ? T(TMP_REG2) : TA(0)) | FS(freg), MOVABLE_INS));
+	return push_inst(compiler, MTC1 | (u.imm[0] != 0 ? T(TMP_REG1) : TA(0)) | FS(freg) | (1 << 11), MOVABLE_INS);
+#endif /* SLJIT_LITTLE_ENDIAN */
 }
 
 SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fcopy(struct sljit_compiler *compiler, sljit_s32 op,


sljit_src/sljitNativeMIPS_common.c
```diff
@@ -243,6 +243,9 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #endif /* SLJIT_MIPS_REV >= 6 */
 #define MOV_S		(HI(17) | FMT_S | LO(6))
 #define MTC1		(HI(17) | (4 << 21))
+#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 2
+#define MTHC1		(MTC1 | (1 << 11))
```
zherczeg, line 247: I think that is flag which is not part of the instruction.

sljit_src/sljitNativeMIPS_32.c
```diff
@@ -56,14 +56,28 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fset64(struct sljit_compiler *comp
...
 
+#if defined(SLJIT_LITTLE_ENDIAN) && SLJIT_LITTLE_ENDIAN
 	FAIL_IF(push_inst(compiler, MTC1 | (u.imm[0] != 0 ? T(TMP_REG1) : TA(0)) | FS(freg), MOVABLE_INS));
-	return push_inst(compiler, MTC1 | (u.imm[1] != 0 ? T(TMP_REG2) : TA(0)) | FS(freg) | (1 << 11), MOVABLE_INS);
+	return push_inst(compiler, MTHC1 | (u.imm[1] != 0 ? T(TMP_REG2) : TA(0)) | FS(freg), MOVABLE_INS);
+#else
+	FAIL_IF(push_inst(compiler, MTC1 | (u.imm[1] != 0 ? T(TMP_REG2) : TA(0)) | FS(freg), MOVABLE_INS));
+	return push_inst(compiler, MTHC1 | (u.imm[0] != 0 ? T(TMP_REG1) : TA(0)) | FS(freg), MOVABLE_INS);
+#endif /* SLJIT_LITTLE_ENDIAN */
+#else
```
zherczeg, line 72: `/* SLJIT_MIPS_REV < 2 */`

sljit_src/sljitNativeMIPS_common.c
carenas: If it would be a flag then the incorrect assumption that the same would apply to `MTHC1` would had been correct but instead there is a pending patch that does:

```
diff --git a/sljit_src/sljitNativeMIPS_common.c b/sljit_src/sljitNativeMIPS_common.c
index acfb28d..71b5fc7 100644
--- a/sljit_src/sljitNativeMIPS_common.c
+++ b/sljit_src/sljitNativeMIPS_common.c
@@ -234,6 +234,9 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define LWR            (HI(38))
 #define LWC1           (HI(49))
 #define MFC1           (HI(17) | (0 << 21))
+#if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 2)
+#define MFHC1          (HI(17) | (3 << 21))
+#endif /* SLJIT_MIPS_REV >= 2 */
 #if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
 #define MOD            (HI(0) | (3 << 6) | LO(26))
 #define MODU           (HI(0) | (3 << 6) | LO(27))
```

another advantage of having it as an instruction with the same mnemonic used in the documentation for MIPSr2 is that it is easier to track and makes the need for fallbacks obvious and detectable at build time.

zherczeg: This is confusing for me. On mips 32, there are 32 single precision registers, and 16 double values mapped to this 32. There is no "high" part of the register. Because all float register indices are even the `1 << 11` is used to access the odd one. Isn't this instruction is 64 bit only?

zherczeg: Historical Information:
For MIPS I, MIPS II, and MIPS III the contents of GPR rt are UNPREDICTABLE for the instruction immediately
following MFC1.

So mfc1 should be there for a long time, why do we need emulation?

carenas: > So mfc1 should be there for a long time, why do we need emulation?

not emulation, but an alternative implementation that doesn't use `MFHC1` which is only available since MIPSr2, and yes all calls to `MFC1` with the `1 << 11` "flag" are probably broken already and also need fixing.

there is also (from the maual) on `MFHC1`:

```
Restrictions:
In implementations prior to Release 2 of the architecture, this instruction resulted in a Reserved Instruction exception.
The results are UNPREDICTABLE if StatusFR = 0 and fs is odd.
```

zherczeg: I try to say we don't need MFHC1 on 32 bit mips, and `1 << 11` is not a flag, it adds 1 to an odd numbered register value.

zherczeg (approved):
LGTM

PR #199, opened 2023-09-02
mips: fix SLJIT_COPYSIGN in 32bit

carenas:
While at it split the implementation into bit specific functions and move some helpful macros used by the 64bit version up as well as other moslty mechanical refactoring.

### DIFF
--- a/sljit_src/sljitNativeMIPS_32.c
+++ b/sljit_src/sljitNativeMIPS_32.c
@@ -26,6 +26,23 @@
 
 /* mips 32-bit arch dependent functions. */
 
+static sljit_s32 emit_copysign(struct sljit_compiler *compiler, sljit_s32 op,
+		sljit_sw src1, sljit_sw src2, sljit_sw dst)
+{
+	int is_32 = (op & SLJIT_32);
+	sljit_ins fr_pair = is_32 ? 0 : (1 << 11);
+
+	FAIL_IF(push_inst(compiler, MFC1 | T(TMP_REG2) | FS(src1) | fr_pair, DR(TMP_REG2)));
+	FAIL_IF(push_inst(compiler, MFC1 | T(TMP_REG3) | FS(src2) | fr_pair, DR(TMP_REG3)));
+	if (!is_32 && src1 != dst)
+		FAIL_IF(push_inst(compiler, MOV_fmt(FMT_S) | FS(src1) | FD(dst), MOVABLE_INS));
+	FAIL_IF(push_inst(compiler, XOR | T(TMP_REG2) | D(TMP_REG3) | S(TMP_REG3), DR(TMP_REG3)));
+	FAIL_IF(push_inst(compiler, SRL | T(TMP_REG3) | D(TMP_REG3) | SH_IMM(31), DR(TMP_REG3)));
+	FAIL_IF(push_inst(compiler, SLL | T(TMP_REG3) | D(TMP_REG3) | SH_IMM(31), DR(TMP_REG3)));
+	FAIL_IF(push_inst(compiler, XOR | T(TMP_REG3) | D(TMP_REG2) | S(TMP_REG2), DR(TMP_REG2)));
+	return push_inst(compiler, MTC1 | T(TMP_REG2) | FS(dst) | fr_pair, MOVABLE_INS);
+}
+
 static sljit_s32 load_immediate(struct sljit_compiler *compiler, sljit_s32 dst_ar, sljit_sw imm)
 {
 	if (!(imm & ~0xffff))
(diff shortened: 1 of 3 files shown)


sljit_src/sljitNativeMIPS_32.c
```diff
@@ -26,6 +26,25 @@
...
+		FAIL_IF(push_inst(compiler, MFC1 | T(TMP_REG1) | FS(src1), DR(TMP_REG1)));
+	FAIL_IF(push_inst(compiler, MFC1 | T(TMP_REG2) | FS(src1) | fr_pair, DR(TMP_REG2)));
+	FAIL_IF(push_inst(compiler, MFC1 | T(TMP_REG3) | FS(src2) | fr_pair, DR(TMP_REG3)));
+	FAIL_IF(push_inst(compiler, XOR | T(TMP_REG2) | D(TMP_REG3) | S(TMP_REG3), DR(TMP_REG3)));
+	FAIL_IF(push_inst(compiler, SRL | T(TMP_REG3) | D(TMP_REG3) | SH_IMM(31), DR(TMP_REG3)));
+	FAIL_IF(push_inst(compiler, SLL | T(TMP_REG3) | D(TMP_REG3) | SH_IMM(31), DR(TMP_REG3)));
+	FAIL_IF(push_inst(compiler, XOR | T(TMP_REG3) | D(TMP_REG2) | S(TMP_REG2), DR(TMP_REG2)));
+	if (!is_32)
+		FAIL_IF(push_inst(compiler, MTC1 | T(TMP_REG1) | FS(dst), MOVABLE_INS));
+	return push_inst(compiler, MTC1 | T(TMP_REG2) | FS(dst) | fr_pair, MOVABLE_INS);
```
zherczeg, line 45: I am not sure I understand this code. Copysign copies only one bit, you never need to touch the second 32 bit part for doubles.

carenas: It is there so the same code could be used with floats; TMP_REG1 (that is needed to hold the lower int in doubles and that is otherwise copied AS-IS from src to dst) goes unused

zherczeg: Ok, I get it now. We have two cases:
a) src1 == dst, when only the sign bit needs to be copied into src
b) src1 != dst, where the lower word needs to be copied as well

I think the original code only did (a), the new only (b), so we should support both.

sljit_src/sljitNativeMIPS_32.c
```diff
@@ -26,6 +26,25 @@
...
 /* mips 32-bit arch dependent functions. */
 
+static sljit_s32 emit_copysign(struct sljit_compiler *compiler, sljit_s32 op,
+		sljit_sw src1, sljit_sw src2, sljit_sw dst)
+{
+	int is_32 = (op & SLJIT_32);
+	sljit_ins fr_pair = is_32 ? 0 : (1 << 11);
+
+	if (!is_32 && src1 != dst)
+		FAIL_IF(push_inst(compiler, MFC1 | T(TMP_REG1) | FS(src1), DR(TMP_REG1)));
```
zherczeg, line 36: One more thing. You can directly copy single precision registers, you don't need to move them into a cpu register. It is also one less instruction.

zherczeg: Put this copy before the XOR to improve parallel execution in the cpu pipeline.

sljit_src/sljitNativeMIPS_32.c
```diff
@@ -26,6 +26,23 @@
...
 /* mips 32-bit arch dependent functions. */
 
+static sljit_s32 emit_copysign(struct sljit_compiler *compiler, sljit_s32 op,
+		sljit_sw src1, sljit_sw src2, sljit_sw dst)
+{
+	int is_32 = (op & SLJIT_32);
+	sljit_ins fr_pair = is_32 ? 0 : (1 << 11);
+
+	if (!is_32 && src1 != dst)
+		FAIL_IF(push_inst(compiler, MOV_fmt(FMT_D) | FS(src1) | FD(dst), DR(TMP_REG1)));
```
zherczeg, line 36: Move this down before the xor, and change draft to ready for review. Is it better to move the whole 64 bit instead of the upper 32? I have no idea which one is better.

carenas: > Is it better to move the whole 64 bit instead of the upper 32?

I don't think that is supported by the instruction as described in the documentation:

```
The operand must be a value in format fmt; if it is not, the result is UNPREDICTABLE and the value of the operand
FPR becomes UNPREDICTABLE.
```

zherczeg: The format would be single precision, and +1 is added or src and dst. Anyway, if you say this is good, I am ok with it. I don't know if it causes dependency issues or the opposite.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -324,6 +323,8 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define LOAD_W		LD
 #endif
 
+#define MOV_fmt(f)	(HI(17) | f | LO(6))
```
zherczeg, line 326: All other float instructions use the `_S` postfix (because fmt s is zero). Do you plan to change all of them?

zherczeg (approved):
LGTM

sljit_src/sljitNativeMIPS_common.c
carenas: `FMT_S` is actually 16 and all had been changed already as part of the "refactoring" done here

zherczeg: I am talking about the other float instructions such as add.fmt, sub.fmt, etc.

PR #200, opened 2023-09-07
mips: correctly detect valid FPU and support StatusFR=1 in 32bit

carenas:
Mostly fixes issues when running through `qemu` and compatibility with newer CPUs that don't support or discourage the use of emulated 32bit FPU registers.

It doesn't address all endianess issues though but will run cleanly in both 32/64bit big endian and addresses manipulation of unaligned doubles which need to be loaded in the right order to avoid unpredictable values as described in:

```
When paired with MTC1 to write a value to a 64-bit FPR, the MTC1 must be executed first, followed by the MTHC1.
This is because of the semantic definition of MTC1, which is not aware that software is using an MTHC1 instruction
to complete the operation, and sets the upper half of the 64-bit FPR to an UNPREDICTABLE value. 
```

### DIFF
--- a/sljit_src/sljitNativeMIPS_32.c
+++ b/sljit_src/sljitNativeMIPS_32.c
@@ -30,17 +30,43 @@ static sljit_s32 emit_copysign(struct sljit_compiler *compiler, sljit_s32 op,
 		sljit_sw src1, sljit_sw src2, sljit_sw dst)
 {
 	int is_32 = (op & SLJIT_32);
-	sljit_ins fr_pair = is_32 ? 0 : (1 << 11);
+	sljit_ins mfhc = MFC1, mthc = MTC1;
+	sljit_ins src1_r = FS(src1), src2_r = FS(src2), dst_r = FS(dst);
 
-	FAIL_IF(push_inst(compiler, MFC1 | T(TMP_REG2) | FS(src1) | fr_pair, DR(TMP_REG2)));
-	FAIL_IF(push_inst(compiler, MFC1 | T(TMP_REG3) | FS(src2) | fr_pair, DR(TMP_REG3)));
+	if (!is_32) {
+		switch (cpu_feature_list & CPU_FEATURE_FR) {
+#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 2
+		case CPU_FEATURE_FR:
+			mfhc = MFHC1;
+			mthc = MTHC1;
+			break;
+#endif /* SLJIT_MIPS_REV >= 2 */
+		default:
+			src1_r |= (1 << 11);
+			src2_r |= (1 << 11);
+			dst_r |= (1 << 11);
+			break;
+		}
+	}
+
+	FAIL_IF(push_inst(compiler, mfhc | T(TMP_REG1) | src1_r, DR(TMP_REG1)));
+	FAIL_IF(push_inst(compiler, mfhc | T(TMP_REG2) | src2_r, DR(TMP_REG2)));
 	if (!is_32 && src1 != dst)
 		FAIL_IF(push_inst(compiler, MOV_fmt(FMT_S) | FS(src1) | FD(dst), MOVABLE_INS));
-	FAIL_IF(push_inst(compiler, XOR | T(TMP_REG2) | D(TMP_REG3) | S(TMP_REG3), DR(TMP_REG3)));
-	FAIL_IF(push_inst(compiler, SRL | T(TMP_REG3) | D(TMP_REG3) | SH_IMM(31), DR(TMP_REG3)));
-	FAIL_IF(push_inst(compiler, SLL | T(TMP_REG3) | D(TMP_REG3) | SH_IMM(31), DR(TMP_REG3)));
-	FAIL_IF(push_inst(compiler, XOR | T(TMP_REG3) | D(TMP_REG2) | S(TMP_REG2), DR(TMP_REG2)));
-	return push_inst(compiler, MTC1 | T(TMP_REG2) | FS(dst) | fr_pair, MOVABLE_INS);
+#if !defined(SLJIT_MIPS_REV) || SLJIT_MIPS_REV <= 1
+	else
+		FAIL_IF(push_inst(compiler, NOP, UNMOVABLE_INS));
+#endif /* MIPS III */
+	FAIL_IF(push_inst(compiler, XOR | T(TMP_REG1) | D(TMP_REG2) | S(TMP_REG2), DR(TMP_REG2)));
+	FAIL_IF(push_inst(compiler, SRL | T(TMP_REG2) | D(TMP_REG2) | SH_IMM(31), DR(TMP_REG2)));
+	FAIL_IF(push_inst(compiler, SLL | T(TMP_REG2) | D(TMP_REG2) | SH_IMM(31), DR(TMP_REG2)));
+	FAIL_IF(push_inst(compiler, XOR | T(TMP_REG2) | D(TMP_REG1) | S(TMP_REG1), DR(TMP_REG1)));
+	FAIL_IF(push_inst(compiler, mthc | T(TMP_REG1) | dst_r, MOVABLE_INS));
+#if !defined(SLJIT_MIPS_REV) || SLJIT_MIPS_REV <= 1
+	if (mthc == MTC1)
+		return push_inst(compiler, NOP, UNMOVABLE_INS);
+#endif /* MIPS III */
+	return SLJIT_SUCCESS;
 }
 
 static sljit_s32 load_immediate(struct sljit_compiler *compiler, sljit_s32 dst_ar, sljit_sw imm)
@@ -65,7 +91,15 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fset64(struct sljit_compiler *comp
 	sljit_s32 freg, sljit_f64 value)
 {
 	union {
-		sljit_s32 imm[2];
+		struct {
+#if defined(SLJIT_LITTLE_ENDIAN) && SLJIT_LITTLE_ENDIAN
+			sljit_s32 lo;
+			sljit_s32 hi;
+#else /* !SLJIT_LITTLE_ENDIAN */
+			sljit_s32 hi;
+			sljit_s32 lo;
+#endif /* SLJIT_LITTLE_ENDIAN */
+		} bin;
 		sljit_f64 value;
 	} u;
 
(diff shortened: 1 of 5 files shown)


sljit_src/sljitNativeMIPS_32.c
```diff
@@ -30,17 +30,37 @@ static sljit_s32 emit_copysign(struct sljit_compiler *compiler, sljit_s32 op,
...
+	sljit_ins mfhc = MFC1, mthc = MTC1;
+	sljit_ins src1_r = FS(src1), src2_r = FS(src2), dst_r = FS(dst);
+
+	if (!is_32) {
+#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 2
+		if (cpu_feature_list & CPU_FEATURE_FR) {
+			mfhc = MFHC1;
+			mthc = MTHC1;
+#else
+		if (0) {
```
zherczeg, line 42: This construct does not looks nice. Maybe we don't need `#if` at all. Or use a `switch` or a `do while(0)` with `break`

sljit_src/sljitNativeMIPS_32.c
```diff
@@ -110,17 +139,31 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fcopy(struct sljit_compiler *compi
...
 
-		inst = FS(freg) | (1 << 11);
+#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 2
+		if (cpu_feature_list & CPU_FEATURE_FR) {
+			mthc = MTHC1;
+			mfhc = MFHC1;
+			inst = FS(freg);
+		} else
+#endif /* SLJIT_MIPS_REV >= 2 */
+			inst = FS(freg) | (1 << 11);
```
zherczeg, line 149: You don't need else here, just keep the original code. One less branch instruction as well. `inst ^= (1 << 11)` can be used as well.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -353,6 +366,56 @@ static sljit_s32 function_check_is_freg(struct sljit_compiler *compiler, sljit_s
...
+{
+#if defined(__GNUC__)
+#if !defined(SLJIT_IS_FPU_AVAILABLE)
+	sljit_u32 fir;
+#endif
+#if SLJIT_CONFIG_MIPS_32
+#if (!defined(SLJIT_DETECT_FR) || (defined(SLJIT_DETECT_FR) && SLJIT_DETECT_FR == 1)) \
+	&& (!defined(SLJIT_DETECT_FR) || SLJIT_DETECT_FR == 1) \
+	&& (!defined(SLJIT_MIPS_REV) || (defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV < 6))
+	double zero = 0.0;
```
zherczeg, line 379: sljit_f64

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -353,6 +366,56 @@ static sljit_s32 function_check_is_freg(struct sljit_compiler *compiler, sljit_s
...
+	int flag = -1;
```
zherczeg, line 380: `sljit_u32`

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -2727,9 +2800,9 @@ static SLJIT_INLINE sljit_s32 sljit_emit_fop1_conv_sw_from_f64(struct sljit_comp
 
 	if (FAST_IS_REG(dst)) {
 		FAIL_IF(push_inst(compiler, MFC1 | flags | T(dst) | FS(TMP_FREG1), MOVABLE_INS));
-#if (!defined SLJIT_MIPS_REV || SLJIT_MIPS_REV <= 1)
```
zherczeg, line 2730: What is the problem with this check?

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -3783,15 +3860,19 @@ static sljit_s32 update_mem_addr(struct sljit_compiler *compiler, sljit_s32 *mem
 }
 
 #if (defined SLJIT_LITTLE_ENDIAN && SLJIT_LITTLE_ENDIAN)
-#define MEM16_IMM_FIRST(memw) IMM((memw) + 1)
-#define MEM16_IMM_SECOND(memw) IMM(memw)
-#define MEMF64_FS_FIRST(freg) FS(freg)
-#define MEMF64_FS_SECOND(freg) (FS(freg) | ((sljit_ins)1 << 11))
+#define IMM_F64_FIRST_LEFT(memw)	IMM((memw) + SSIZE_OF(s32) - 1)
```
zherczeg, line 3863: Why -1 at the end? It overlaps with IMM_F64_FIRST_RIGHT

test_src/sljitTest.c
```diff
@@ -11913,7 +11913,7 @@ static void test94(void)
 	struct sljit_compiler* compiler;
 	sljit_f64 dbuf[9];
 	sljit_f32 sbuf[9];
-	sljit_s32 i;
+	int i;
```
zherczeg, line 11916: Why do we need this change?

zherczeg (changes requested):

test_src/sljitTest.c
carenas: It is easier to read and more common, but you are correct it isn't needed

sljit_src/sljitNativeMIPS_common.c
carenas: There were several versions of it so I standarized in one that is more likely to be correct.

The need for `nop` is only required for MIPS III or older, but SLJIT_MIPS_REV == 1 in MIPS64 corresponds to MIPS IV.  It might be correct for MIPS32 though, but I think it is also unlikely anyone would have a mips32r1 CPU with FPU to test, and indeed until this code change no mips32r1 CPU would have FPU enabled because they have 32bit registers.

sljit_src/sljitNativeMIPS_32.c
carenas: `#if` is needed at least to protect the use of `MFHC1` and `MTHC1` which are only defined (and available) if SLJIT_MIPS_REV >= 2.

not sure I understood how to use the alternatives you proposed but tried to make it a little less ugly

sljit_src/sljitNativeMIPS_32.c
carenas: While I agree the else is a little off, there is IMHO no way to get rid of the branch (at least for `SLJIT_MIPS_REV >= 2` which I hope is the most common path.

Will do the double assignment of `inst` to remove the else as instructed, with the hope that the compiler will optimize it somehow.

sljit_src/sljitNativeMIPS_common.c
zherczeg: Shall we make mipsr1 as minimum?

sljit_src/sljitNativeMIPS_32.c
zherczeg:
```
switch (cpu_feature_list & CPU_FEATURE_FR) {
#if ...
case CPU_FEATURE_FR:
  break;
#endif
default:
  break;
}

File: sljit_src/sljitNativeMIPS_32.c (hunk 2 above)
Comment (zherczeg) [reply sljit_src/sljitNativeMIPS_32.c:149]: The point is, if the compiler does not optimize it, it is the better. Setting a value is cheap, and you don't need a branch at the end.

**Review by zherczeg — CHANGES_REQUESTED** [review]

File: sljit_src/sljitNativeMIPS_common.c (hunk 5 above)
Comment (carenas) [reply sljit_src/sljitNativeMIPS_common.c:2730]: brought the check back with a more conservative setting so at least `NOP` could be skipped for most MIPS64 CPUs.

**Review by zherczeg — CHANGES_REQUESTED** [review]
The patch is quite good. There are some unanswered questions before, please check them.

**Comment by zherczeg** [comment]
Thank you for maintaining mips. It would be good to discuss the future of this port. 

The following link says: MIPS32 was introduced in 1999.
https://en.wikipedia.org/wiki/MIPS_architecture#MIPS32/MIPS64

Maybe we could drop mips III support, and make that mandatory.

File: sljit_src/sljitNativeMIPS_common.c [hunk 8]
```diff
@@ -353,6 +366,60 @@ static sljit_s32 function_check_is_freg(struct sljit_compiler *compiler, sljit_s
...
+	int feature_list = CPU_FEATURE_DETECTED;
+
+#if defined(__GNUC__) && !defined(SLJIT_IS_FPU_AVAILABLE)
+	__asm__ ("cfc1 %0, $0" : "=r"(fir));
+#endif
+	if ((fir & (1 << 22)) != 0)
+		feature_list |= CPU_FEATURE_FPU;
+
+	if (!(feature_list & CPU_FEATURE_FPU) &&
+		(fir & (1 << 17)) != 0 && (fir & (1 << 21)) != 0)
```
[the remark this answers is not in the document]
carenas: since the bits are not adjacent, it is easier to see which ones are relevant when looking at the documentation this way.

frankly I suspect that only one of them is required for our use case, but since I have no mips32r1 CPU to test, this is safer.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -353,6 +366,60 @@ static sljit_s32 function_check_is_freg(struct sljit_compiler *compiler, sljit_s
...
+		feature_list |= CPU_FEATURE_FP32;
+
+#if defined(SLJIT_CONFIG_MIPS_32) && SLJIT_CONFIG_MIPS_32
+	if ((feature_list & CPU_FEATURE_FPU) && ((feature_list & CPU_FEATURE_FP32) == 0)) {
+#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 6
+		feature_list |= CPU_FEATURE_FR;
+#elif defined(SLJIT_DETECT_FR) && SLJIT_DETECT_FR == 0
+#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 5
+		feature_list |= CPU_FEATURE_FR;
```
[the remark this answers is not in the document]
carenas: `SLJIT_DETECT_FR` is not defined by default, and therefore the fact that it is set to off is the most relevant part of this test.

if it was set to ON, then the detection will be done even when SLJIT_MIPS_REV == 5, but never when SLJIT_MIPS_REV >= 6 since it is no longer relevant per the documentation.

Note that the detection is basically a "hack".  The "correct" way will be to look for the StatusFR flag directly (bit 26 in the status register), but doing so requires a privileged instruction (except in MIPSr5, but in a way that is no longer supported by MIPSr6) and the "only" way that would be compatible up to MIPSr2 is to `ptrace(GETREGS)`.

carenas:
Ironically, I think that the best value of this codebase isn't in the availability of CPUs but the fact that by being bi-endian and RISC could be used as inspiration for other architectures far more relevant.

It might be even the only one where softfp is becoming more relevant, and that might come useful when RISCV32 gets deployed instead of MIPS.

sljit_src/sljitNativeMIPS_common.c
carenas: There is no overlap, for example, in a double that is offset by 2 bytes and assuming the address from the aligned buffer  is in `v0` the following will be needed to load `f0` (in legacy mode):

```
	lwl    t9,6(v0)
	lwr    t9,9(v0)
	mtc1   t9,$f0
	lwl    t9,2(v0)
	lwr    t9,5(v0)
	mtc1   t9,$f1
```

zherczeg (approved):
LGTM

zherczeg:
Half of the codebase is guarded by `#if`-s. This is by far the most complex code, so I doubt anybody would use this as inspiration. 

As for me, JIT compiler is a performance oriented system. If you rely it on a 15 year old system, something is wrong. If you have that old machine, you are happy that it works, and don't want to touch it because it will break. You don't need the latest software, just something that works. This is why I dropped armv5 support, you can find armv5 systems in a museum, but you definitely don't want to touch their software.

What do you think, what would be the optimal requirement for systems we support? I would say 10 year old systems.

PR #202, opened 2023-09-12
mips: allow detection of StatusFR with prctl()

carenas:
Improve StatusFR detection by using a system provided interface


zherczeg (approved):
LGTM

PR #203, opened 2023-09-13
mips: enable SLJIT_HAVE_REV for SLJIT_MIPS_REV >= 2

carenas:
Adds 3 instructions from "revision 2" that allow for a new implementation of `SLJIT_REV` (and friends) that should be good enough to enable the corresponding `sljit_has_cpu_feature()` flag.

[documentation](https://github.com/zherczeg/sljit/blob/3dcdb25754363db2cd34dbd1cd9a6a6b880cadae/doc/overview.txt#L41-L44) for what is an invalid operation has been also updated.

### DIFF
--- a/sljit_src/sljitNativeMIPS_common.c
+++ b/sljit_src/sljitNativeMIPS_common.c
@@ -218,12 +218,18 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define DMULTU		(HI(0) | LO(29))
 #endif /* SLJIT_MIPS_REV >= 6 */
 #define DIV_S		(HI(17) | FMT_S | LO(3))
+#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 2
 #define DINSU		(HI(31) | LO(6))
+#endif /* SLJIT_MIPS_REV >= 2 */
 #define DMFC1		(HI(17) | (1 << 21))
 #define DMTC1		(HI(17) | (5 << 21))
+#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 2
 #define DROTR		(HI(0) | (1 << 21) | LO(58))
 #define DROTR32		(HI(0) | (1 << 21) | LO(62))
 #define DROTRV		(HI(0) | (1 << 6) | LO(22))
+#define DSBH		(HI(31) | (2 << 6) | LO(36))
+#define DSHD		(HI(31) | (5 << 6) | LO(36))
+#endif /* SLJIT_MIPS_REV >= 2 */
 #define DSLL		(HI(0) | LO(56))
 #define DSLL32		(HI(0) | LO(60))
 #define DSLLV		(HI(0) | LO(20))
@@ -281,8 +287,10 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define NOR		(HI(0) | LO(39))
 #define OR		(HI(0) | LO(37))
 #define ORI		(HI(13))
+#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 2
 #define ROTR		(HI(0) | (1 << 21) | LO(2))
 #define ROTRV		(HI(0) | (1 << 6) | LO(6))
+#endif /* SLJIT_MIPS_REV >= 2 */
 #define SD		(HI(63))
 #define SDL		(HI(44))
 #define SDR		(HI(45))
@@ -304,6 +312,9 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define SWR		(HI(46))
 #define SWC1		(HI(57))
 #define TRUNC_W_S	(HI(17) | FMT_S | LO(13))
+#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 2
+#define WSBH		(HI(31) | (2 << 6) | LO(32))
+#endif /* SLJIT_MIPS_REV >= 2 */
 #define XOR		(HI(0) | LO(38))
 #define XORI		(HI(14))
 
@@ -325,8 +336,10 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #endif /* SLJIT_MIPS_REV >= 6 */
 #define PREF		(HI(51))
 #define PREFX		(HI(19) | LO(15))
+#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 2
 #define SEB		(HI(31) | (16 << 6) | LO(32))
 #define SEH		(HI(31) | (24 << 6) | LO(32))
+#endif /* SLJIT_MIPS_REV >= 2 */
 #endif /* SLJIT_MIPS_REV >= 1 */
 
 #if (defined SLJIT_CONFIG_MIPS_32 && SLJIT_CONFIG_MIPS_32)
@@ -882,6 +895,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 		return 2;
 #endif /* SLJIT_MIPS_REV >= 1 */
 #if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 2)
+	case SLJIT_HAS_REV:
 	case SLJIT_HAS_ROT:
 		return 1;
 #endif /* SLJIT_MIPS_REV >= 2 */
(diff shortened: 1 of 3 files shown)


zherczeg:
Could you explain the problem with the test?

carenas:
> Could you explain the problem with the test?

`ibuf[4]` related code is missing, but more importantly the original case 12 sets junk in the upper half of the register instead of the full sign extension that would be expected, if it is meant to be truncated and used later as a 32bit register.

I understand that being a test is trying to trigger edge cases, but it goes against the "principle" that defines all other sljit operations as valid, and it requires additional code to guard against.

zherczeg:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitLir.h#L1106

It looks like rev32 handles 32 bit input, the rest should handle word sized input. So no sign extension is needed.

carenas:
not sure I understand your comment, but the problem isn't in sljit code but the fact that the instructions in MIPS have a restriction that says:

```
If GPR rt does not contain a sign-extended 32-bit value (bits 63..31 equal), then the result of the operation is
UNPREDICTABLE.
```

I added a case for test89 that hopefully helps clarify the issue and that does NOT fail on either QEMU or the Octeon CPU I had been testing with, and my assumption is that hopefully "UNPREDICTABLE" means that since the sign is ambiguous when the CPU completes the operation and does the sign extension of the result to put it back in the destination register it might get it wrong. Luckily we already have to "fix" (MIPS always sign extends) the sign extension anyway for `SLJIT_REV_{U,S}16` so hopefully that would help there.

carenas:
in a related theme, is `SLJIT_REV_S16` and other signed variants really needed?

zherczeg:
There is a reason we have `SLJIT_REV_U16` and `SLJIT_REV32_U16`. The former must not expect that the value is sign extended or anything. The latter can expect that the upper 32 bit is the sign extension of bit 31 (on MIPS). The signed/unsigned forms should be useful for loading signed values.

carenas:
My point was, why would you use `SLJIT_REV_S16` or `SLJIT_REV_S32`, if it is semantically better to use `SLJIT_REV_U16` or `SLJIT_REV_U32`.

For example, `gcc` does provide builtins for all `bswap` functions, but they are only expected to work with unsigned values, and so the whole "sign extension" issue is irrelevant.

in MIPS we will then set the upper word to 0 initially and any "automatic" sign extension coul be safely ignored.

zherczeg:
I mentioned above the primary use case is loading signed reversed values, which can be optimized depending on the cpu capabilities.

test_src/sljitTest.c
```diff
@@ -11288,37 +11301,57 @@ static void test90(void)
 
 	sljit_emit_enter(compiler, 0, SLJIT_ARGS2(VOID, W, W), 5, 5, 0, 0, 2 * sizeof(sljit_sw));
 
-	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_IMM, WCONST(0xffffa1b2c3d4, 0xa1b2c3d4));
```
zherczeg, line 11291: This is not need to be changed

carenas: OK, but then the "description" of the case should change, or even better a different case which specifically handles the "junk" in the upper word is ignored by this instruction should be added instead IMHO

test_src/sljitTest.c
```diff
@@ -11339,11 +11376,6 @@ static void test90(void)
 	/* ibuf[4] */
 	ibuf[4] = -824;
 
-	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R4, 0, SLJIT_IMM, WCONST(0xffff8192a3b4, 0x8192a3b4));
-	sljit_emit_op1(compiler, SLJIT_REV_S32, SLJIT_S4, 0, SLJIT_R4, 0);
-	/* buf[6] */
-	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(SLJIT_S0), 6 * sizeof(sljit_sw), SLJIT_S4, 0);
-
```
carenas, line 11346: maybe this test should had been `IS_64BIT` and focus on the "store" of a truncated 32bit signed instead?

see the case added to test89 for details.

test_src/sljitTest.c
zherczeg: The instruction reverses the lower 4 bytes, and sign/unsign extend the result. The upper 4 byte is ignored, since its value is not specified. The value is usually the result of some computation, and the upper bits depend on that. There is a REV32 variant, which uses the result of a 32 bit computation, so the upper bits might be defined by the cpu.

GNUmakefile
```diff
@@ -16,7 +16,7 @@ EXTRA_LDFLAGS=
 endif
 
 CPPFLAGS = $(EXTRA_CPPFLAGS) -Isljit_src
-CFLAGS += -O2 -Wall -Wextra -Wconversion -Wsign-compare -Werror
+CFLAGS += -g -O0 -Wall -Wextra -Wconversion -Wsign-compare -Werror
```
zherczeg, line 19: Keep the original.

test_src/sljitTest.c
```diff
@@ -11262,6 +11275,12 @@ static void test89(void)
 	FAILED(hbuf[5] != (sljit_s16)0x0201, "test89 case 14 failed\n");
 	FAILED(hbuf[6] != (sljit_s16)0x0201, "test89 case 15 failed\n");
 	FAILED(hbuf[7] != -367, "test89 case 16 failed\n");
+#if IS_64BIT
+	FAILED(hbuf[8] != 0x55aa, "test89 case 17 failed\n");
+	FAILED(buf[5] != hbuf[8], "test89 case 17 (zero extension) failed\n");
```
zherczeg, line 11280: this can be 18, 19,...

test_src/sljitTest.c
```diff
@@ -11272,51 +11291,59 @@ static void test90(void)
...
 
 	sljit_emit_enter(compiler, 0, SLJIT_ARGS2(VOID, W, W), 5, 5, 0, 0, 2 * sizeof(sljit_sw));
 
+	/* zero extend any integer */
 	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_IMM, WCONST(0xffffa1b2c3d4, 0xa1b2c3d4));
 	sljit_emit_op1(compiler, SLJIT_REV_U32, SLJIT_R0, 0, SLJIT_R0, 0);
 	/* buf[0] */
 	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(SLJIT_S0), 0, SLJIT_R0, 0);
 
-	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R2, 0, SLJIT_IMM, WCONST(0xffffa1b2c3d4, 0xa1b2c3d4));
```
zherczeg, line 11296: All similar tests should be kept.

zherczeg (changes requested):

test_src/sljitTest.c
carenas: but using a few extra characters, provide additional information on what part of the behaviour is broken, as well as clarifying that the incorrect value was generated as part of a single "case" code block and therefore aid with troubleshooting.

test_src/sljitTest.c
carenas: This test is still doing the same, with the only caveat that is now no longer undefined, because the value stored in the register matches what the operation (SLJIT_REV_S32) expects regarding sign extension of the upper 32bit.

The special case were that is not undefined is now covered by the improved case12 that does so by using a MEM destination and ensuring the relevant value was written and that write didn't overflow.

test_src/sljitTest.c
zherczeg: I am ok with the characters, just increase the numbers.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -1671,25 +1694,34 @@ static sljit_s32 emit_rev(struct sljit_compiler *compiler, sljit_s32 op, sljit_s
...
+#endif /* SLJIT_MIPR_REV >= 2 */
 	return SLJIT_SUCCESS;
 }
 
 static sljit_s32 emit_rev16(struct sljit_compiler *compiler, sljit_s32 op, sljit_s32 dst, sljit_sw src)
 {
+#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 2
+#if defined(SLJIT_CONFIG_MIPS_32) && SLJIT_CONFIG_MIPS_32
+	FAIL_IF(push_inst(compiler, WSBH | T(src) | D(dst), DR(dst)));
+#else
```
zherczeg, line 1711: comment `/* !SLJIT_CONFIG_MIPS_32 */` missing

test_src/sljitTest.c
```diff
@@ -11293,30 +11316,37 @@ static void test90(void)
 	/* buf[0] */
 	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(SLJIT_S0), 0, SLJIT_R0, 0);
 
+	/* sign extend negative integer */
```
zherczeg, line 11319: Please use full sentences/
/* Sign extend negative integer. */
Similar cases below.

test_src/sljitTest.c
```diff
@@ -11354,16 +11383,17 @@ static void test90(void)
...
+	FAILED(buf[2] != 0x4d3c2b1a, "test90 case 3 failed\n");
+	FAILED(buf[3] != WCONST(0xc6d7e8f9, 0xc6d7e8f9), "test90 case 4 failed\n");
+	FAILED(buf[4] != WCONST(-0x39281707, 0xc6d7e8f9), "test90 case 5 failed\n");
+	FAILED(buf[5] != WCONST(0xddccbbaa, 0xddccbbaa), "test90 case 6 failed\n");
+	FAILED(buf[6] != WCONST(-0x22334456, 0xddccbbaa), "test90 case 7 failed\n");
+	FAILED(ibuf[0] != 0x04030201, "test90 case 8 failed\n");
+	FAILED(ibuf[1] != 0x04030201, "test90 case 9 failed\n");
+	FAILED(ibuf[2] != (sljit_s32)0xc0d0e0f0, "test90 case 10 failed\n");
+	FAILED(ibuf[3] != (sljit_s32)0xc0d0e0f0, "test90 case 11 failed\n");
+	FAILED(ibuf[4] != (sljit_s32)0xb4a39281, "test90 case 13 failed\n");
```
zherczeg, line 11395: 13 after 11, so 12 is missing

zherczeg (changes requested):
Only some style issues remained

zherczeg (approved):
LGTM

PR #205, opened 2023-09-16
mips: improve FPU detection also for MIPS64

carenas:
Improves support of older FPUs in 64bit while hopefully cleaning up the code slightly.

As a bonus would allow detecting the FPU in: `qemu-mips64-static -cpu 5Kf`

### DIFF
--- a/sljit_src/sljitNativeARM_32.c
+++ b/sljit_src/sljitNativeARM_32.c
@@ -1005,7 +1005,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 	case SLJIT_HAS_FPU:
 	case SLJIT_HAS_F64_AS_F32_PAIR:
 #ifdef SLJIT_IS_FPU_AVAILABLE
-		return SLJIT_IS_FPU_AVAILABLE;
+		return (SLJIT_IS_FPU_AVAILABLE) != 0;
 #else
 		/* Available by default. */
 		return 1;
@@ -1015,7 +1015,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 		return 0;
 #else
 #ifdef SLJIT_IS_FPU_AVAILABLE
-		return SLJIT_IS_FPU_AVAILABLE;
+		return (SLJIT_IS_FPU_AVAILABLE) != 0;
 #else
 		/* Available by default. */
 		return 1;
--- a/sljit_src/sljitNativeMIPS_common.c
+++ b/sljit_src/sljitNativeMIPS_common.c
@@ -367,8 +367,8 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define UIMM_MAX	(0xffff)
 
 #define CPU_FEATURE_DETECTED	(1 << 0)
-#define CPU_FEATURE_FP32	(1 << 1)
-#define CPU_FEATURE_FPU		(1 << 2)
+#define CPU_FEATURE_FPU		(1 << 1)
+#define CPU_FEATURE_FP64	(1 << 2)
 #define CPU_FEATURE_FR		(1 << 3)
 
 static sljit_u32 cpu_feature_list = 0;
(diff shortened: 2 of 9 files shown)


zherczeg:
I get some new errors on mips.
```
/tmp/ccmvccaB.s: Assembler messages:
/tmp/ccmvccaB.s:306: Warning: tried to set unrecognized symbol: oddspreg
```

```
test94 case 7 failed
SLJIT tests: 1 (0%) tests FAILED on MIPS III 64bit (big endian + aligned) (with fpu)
```

carenas:
the test failure is unrelated; I'd seen it with some of my compilers but only when `-march=mips64`.

apparently the whole call to `sljit_emit_fop1_conv_f64_from_sw()` gets somehow dropped from the expansion of `SELECT_FOP1_OPERATION_WITH_CHECKS`.

as for the warning, you could probably get rid of it by using `prctl()` in your system for that detection (ex: `EXTRA_CFLAGS=-DHAVE_PRCTL`) but should be otherwise harmless.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -864,23 +870,12 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
...
 		if (!cpu_feature_list)
 			get_cpu_features();
 
 		fpu_is_valid = (cpu_feature_list & CPU_FEATURE_FPU) != 0;
-#if defined(SLJIT_CONFIG_MIPS_32) && SLJIT_CONFIG_MIPS_32
-		if (!fpu_is_valid)
-			fpu_is_valid = (cpu_feature_list & CPU_FEATURE_FP32) != 0;
-#endif /* SLJIT_CONFIG_MIPS_32 */
+
 		return fpu_is_valid;
```
zherczeg, line 878: Do we need a local variable for this?

zherczeg (changes requested):

zherczeg:
Are we really need that `.set oddspreg`? We could use a local value for loading data instead of a global, so we could delete the `.data` part as well.

carenas:
> really need that `.set oddspreg`?

without it, the assembler is potentially invalid because gcc could be building in FP64A mode or simply `-mno-odd-spreg` which forbids using odd floating point registers and therefore fail to build.

r5 CPUs, can even be made (FRE mode) to trap those instructions so they will be transparently translated into `mfhc1/mthc1` as needed.

sljit_src/sljitNativeARM_32.c
```diff
@@ -1005,7 +1005,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 	case SLJIT_HAS_FPU:
 	case SLJIT_HAS_F64_AS_F32_PAIR:
 #ifdef SLJIT_IS_FPU_AVAILABLE
-		return SLJIT_IS_FPU_AVAILABLE;
+		return SLJIT_IS_FPU_AVAILABLE != 0;
```
zherczeg, line 1008: What is the aim of this change?

zherczeg:
The "float register should be even" looks an invalid error, since 32 bit registers should work with any index regardless of mode. Btw any scratch float register, e.g. 0 should work here, since this is not a jit code.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -435,27 +443,25 @@ static void get_cpu_features(void)
 		if (flag < 0) {
 			SLJIT_ASSERT(freg_map[TMP_FREG3] == 16);
```
zherczeg, line 444: There should be no such requirement because this is not a jit code refer to this check. We could simply use f0 or any scratch register.

sljit_src/sljitNativeARM_32.c
carenas: `sljit_has_cpu_feature()` returns 1 if the feature is available, but `SLJIT_IS_FPU_AVAILABLE` can be set to any non zero value to mean "available".

before this change it was always assumed that `SLJIT_IS_FPU_AVAILABLE` will either be 0 or 1 which was not technically correct.

carenas:
> 32 bit registers should work with any index regardless of mode.

I don't think the message indicates that they don't work, but only that (as I mentioned earlier) they are not allowed by policy to prevent ABI incompatibilities or traps.

`.set oddspreg` to the assembler is what indicates that is not a concern by this code.  clang's message might be more clear:

```
In file included from sljit_src/sljitLir.c:3004:
sljit_src/sljitNativeMIPS_common.c:446:5: error: -mno-odd-spreg prohibits the use of odd FPU registers
  446 |                                 "lwc1 $f17, %0\n"
      |                                 ^
<inline asm>:1:7: note: instantiated into assembly here
    1 |         lwc1 $f17, 20($fp)
      |              ^
```

sljit_src/sljitNativeMIPS_common.c
carenas: It was meant to document what the magic numbers below meant, but agree was a little harsh

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -390,72 +390,76 @@ static sljit_s32 function_check_is_freg(struct sljit_compiler *compiler, sljit_s
...
-				"0:\n"
-				"	.quad 0\n"
-				".text\n"
-				".set oddspreg\n"
-				"	mtc1 %0, $f17\n"
-				"	ldc1 $f16, 0b\n"
-				"	nop\n"
-				"	mfc1 %0, $f17\n"
-			: "+r" (flag) : : "$f16", "$f17");
+			/* TMP_FREG3 is expected to be mapped to register f16 */
```
zherczeg, line 444: The point is, it does not matter which fpu register you use. The comment makes it look like it does. I would use the typical function return value (`f0` `f1` I think).

sljit_src/sljitNativeARM_32.c
zherczeg: Change to `(SLJIT_IS_FPU_AVAILABLE) != 0`, so it even can be an expression (some `syscall() & flag` for example).

zherczeg (changes requested):

sljit_src/sljitNativeMIPS_common.c
carenas: Fair enough, but then I will rather remove the comment than use a register pair that "might" need saving/restoring.  f16/f17 is a "temporary register" per the n32 ABI which is indirectly why the comment meant.

zherczeg: Ok, then mention that f16 is an assembler temporary in the MIPS ABI. Anyway since "$f16", "$f17" is passed as clobbered, the compiler will not use them for anything. The function itself does not use floating point at all.

zherczeg (approved):
LGTM

PR #206, opened 2023-09-18
add SLJIT_REV16PACK and SLJIT_REV32PACK

carenas:
Useful when bytes swapping in bulk and therefore as it is meant to be performance sensitive, not Including (at least for now) emulation.

Not particularly fond of he name, but at least keeps the bit width of the instruction away from the bit width of the register size, even if a mouthful with `SLJIT_REV16PACK32`


zherczeg:
Is there a use case for this? When I was thinking about this before, I decided not to do it until the support needed for something. For example, the normal rev is needed for big-endian WebAssembly support, since WebAssembly must use little endian memory accesses.

carenas:
> Is there a use case for this?

no, was more interested on feedback on the design and code, as well as the potential performance improvements if bulk bswaps were needed as well as if it was worth doing it here or going all the way to vector instructions (most likely the later).

Issue #207, opened 2023-09-21
Reclaiming unused space in executable allocations


invertego:
I understand that `sljit_generate_code` can allocate executable memory in excess of the necessary amount, as the exact amount of required space is not known at the start of the call. That said, I can see no way to cleanly reclaim the unused space at the end of every allocation.

I have a custom executable allocator, and currently I can use `sljit_get_generated_code_size` to determine the actual amount of used memory... on most architectures. However, on PPC (when SLJIT_INDIRECT_CALL=1) and s390x, there may be additional data at the end. It makes me feel that I probably shouldn't be using this API in that way.

I see a few ways forward:
1) A new getter API that includes total amount of used space, including any data after the executable code
2) A new SLJIT_REALLOC_EXEC [1] macro invoked at the end of the generate function to signal the reduced size to the allocator; the default implementation could do nothing, as it would only be used to shrink allocations.
3) Live life and stop worrying about a few unused bytes

I'm curious if there is any preference here, or if I've overlooked another possible solution.

[1] Maybe a bad name, because unlike `realloc` it should never move things around

carenas:
> I've overlooked another possible solution

provide another customizable allocator that will be called when additional memory for data is needed and therefore you could always trust that `sljit_get_generated_code_size` is always accurate.

I wonder though, why isn't option 3 the most obvious choice?, how much space (as a % of the generated code, and including any alignment needs) are you expecting to save?

invertego:
> I wonder though, why isn't option 3 the most obvious choice?, how much space (as a % of the generated code, and including any alignment needs) are you expecting to save?

I just took some measurements, and for [our use case](https://github.com/ares-emulator/ares), on x86_64, we're seeing ~20%-40% wasted space. We generate branch-heavy code and the extreme pessimization of direct branch size estimates (13 bytes!) seems to be the killer.

> provide another customizable allocator that will be called when additional memory for data is needed and therefore you could always trust that `sljit_get_generated_code_size` is always accurate.

That could work, but I'm having a hard time seeing how one could implement such an API while still making it easy for the data to be placed immediately after the executable code as it is today.

zherczeg:
I think this is a valid use case. It is true that branches are overestimated, and you cannot reallocate code, since the absolute offsets embedded into the code would become invalid. The estimation could be improved as well.

https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitNativeX86_common.c#L790
[...]

PR #209, opened 2023-09-22
s390x: make sljit_generated_code_size() accurate

carenas:
Include the additional data added in some architectures at the end of the generated code to the reported `executable_size`

Fixes: #207

### DIFF
--- a/sljit_src/sljitNativePPC_common.c
+++ b/sljit_src/sljitNativePPC_common.c
@@ -452,6 +452,7 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 	reverse_buf(compiler);
 
 #if (defined SLJIT_INDIRECT_CALL && SLJIT_INDIRECT_CALL)
+	/* add to compiler->size additional instruction space to hold the trampoline and padding */
 #if (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64)
 	compiler->size += (compiler->size & 0x1) + (sizeof(struct sljit_function_context) / sizeof(sljit_ins));
 #else
@@ -652,7 +653,6 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 
 	compiler->error = SLJIT_ERR_COMPILED;
 	compiler->executable_offset = executable_offset;
-	compiler->executable_size = (sljit_uw)(code_ptr - code) * sizeof(sljit_ins);
 
 	code = (sljit_ins *)SLJIT_ADD_EXEC_OFFSET(code, executable_offset);
 
@@ -670,8 +670,12 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 	SLJIT_UPDATE_WX_FLAGS(code, code_ptr, 1);
 
 #if (defined SLJIT_INDIRECT_CALL && SLJIT_INDIRECT_CALL)
+	compiler->executable_size = (sljit_uw)(code_ptr - code) * sizeof(sljit_ins) + sizeof(struct sljit_function_context);
+
 	return code_ptr;
 #else
+	compiler->executable_size = (sljit_uw)(code_ptr - code) * sizeof(sljit_ins);
+
 	return code;
 #endif
 }
--- a/sljit_src/sljitNativeS390X.c
+++ b/sljit_src/sljitNativeS390X.c
@@ -1651,6 +1651,8 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 	compiler->error = SLJIT_ERR_COMPILED;
 	compiler->executable_offset = executable_offset;
 	compiler->executable_size = ins_size;
+	if (pool_size)
+		compiler->executable_size += (pad_size + pool_size);
 	code = SLJIT_ADD_EXEC_OFFSET(code, executable_offset);
 	code_ptr = SLJIT_ADD_EXEC_OFFSET(code_ptr, executable_offset);
 	SLJIT_CACHE_FLUSH(code, code_ptr);


zherczeg:
This will not fix #207 because on ppc, the extra function structure is (correctly) not included in the code size.

carenas:
in AIX at least, where SLJIT_INDIRECT_CALL is being set, the "code" pointer is 1 or 2 instructions after the end of the generated function, so at least from the point of view of a custom memory allocator, that could be used to infer where the generated code starts and ends.

sljit_src/sljitNativePPC_common.c
```diff
@@ -653,6 +653,9 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 	compiler->error = SLJIT_ERR_COMPILED;
 	compiler->executable_offset = executable_offset;
 	compiler->executable_size = (sljit_uw)(code_ptr - code) * sizeof(sljit_ins);
+#if defined(SLJIT_INDIRECT_CALL) && SLJIT_INDIRECT_CALL
+	compiler->executable_size += (2 * sizeof(sljit_ins));
```
zherczeg, line 657: There is an alignment code (`code_ptr++`) below.

carenas: I am not sure how the trampoline is supposed to work for 64bit though (from a disassembly of a generated `void fn(void)` in AIX):

```
(dbx) p code.code
0x0a00000000000038 
(dbx) p compiler->executable_size
64 
(dbx) 0xa00000000000010/16i
0xa00000000000010 (???) 7c0802a6           mflr   r0
0xa00000000000014 (???) f821ff81           stdu   r1,-128(r1)
0xa00000000000018 (???) fbe10078            std   r31,0x78(r1)
0xa0000000000001c (???) f8010090            std   r0,0x90(r1)
0xa00000000000020 (???) 3be00000             li   r31,0x0
0xa00000000000024 (???) e8010090             ld   r0,0x90(r1)
0xa00000000000028 (???) ebe10078             ld   r31,0x78(r1)
0xa0000000000002c (???) 7c0803a6           mtlr   r0
0xa00000000000030 (???) 38210080           addi   r1,0x80(r1)
0xa00000000000034 (???) 4e800020            blr
0xa00000000000038 (???) 0a000000          tdlti   r0,0x0
0xa0000000000003c (???) 00000010       Invalid opcode.
0xa00000000000040 (???) 00000001       Invalid opcode.
0xa00000000000044 (???) 10001208        vmuleub   vr0,vr0,vr2
0xa00000000000048 (???) 00000000       Invalid opcode.
0xa0000000000004c (???) 00000000       Invalid opcode.
```

sljit_src/sljitNativePPC_common.c
```diff
@@ -652,7 +650,7 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 
 	compiler->error = SLJIT_ERR_COMPILED;
 	compiler->executable_offset = executable_offset;
-	compiler->executable_size = (sljit_uw)(code_ptr - code) * sizeof(sljit_ins);
+	compiler->executable_size = (sljit_uw)(code_ptr - code) * sizeof(sljit_ins) + trampoline_size;
```
zherczeg, line 653: I would just move this after sljit_set_function_context

carenas: but if you do, then have to track the 1 instruction padding independently of trampoline size

sljit_src/sljitNativePPC_common.c
```diff
@@ -446,19 +446,21 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
...
 	CHECK_PTR(check_sljit_generate_code(compiler));
 	reverse_buf(compiler);
 
-#if (defined SLJIT_INDIRECT_CALL && SLJIT_INDIRECT_CALL)
 #if (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64)
-	compiler->size += (compiler->size & 0x1) + (sizeof(struct sljit_function_context) / sizeof(sljit_ins));
-#else
-	compiler->size += (sizeof(struct sljit_function_context) / sizeof(sljit_ins));
-#endif
+	pad_size = (compiler->size & 0x1) * sizeof(sljit_ins);
```
zherczeg, line 461: This is a different code than the original. Indirect calls may or may not be present regardless of 32/64 bit.

sljit_src/sljitNativePPC_common.c
```diff
@@ -652,20 +650,19 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
...
 
 	code = (sljit_ins *)SLJIT_ADD_EXEC_OFFSET(code, executable_offset);
 
 #if (defined SLJIT_INDIRECT_CALL && SLJIT_INDIRECT_CALL)
-#if (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64)
-	if (((sljit_sw)code_ptr) & 0x4)
+	if (pad_size)
 		code_ptr++;
-#endif
 	sljit_set_function_context(NULL, (struct sljit_function_context*)code_ptr, (sljit_uw)code, (void*)sljit_generate_code);
```
zherczeg, line 659: Just add the size to `code_ptr`, and the `executable_size` computation below is enough.

zherczeg (changes requested):

sljit_src/sljitNativePPC_common.c
carenas: but the padding is only needed for 64bit.

sljit_src/sljitNativePPC_common.c
```diff
@@ -652,20 +653,21 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
...
 	code = (sljit_ins *)SLJIT_ADD_EXEC_OFFSET(code, executable_offset);
 
 #if (defined SLJIT_INDIRECT_CALL && SLJIT_INDIRECT_CALL)
 #if (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64)
-	if (((sljit_sw)code_ptr) & 0x4)
+	if ((sljit_sw)code_ptr & 0x4)
 		code_ptr++;
-#endif
+#endif /* SLJIT_CONFIG_PPC_64 */
 	sljit_set_function_context(NULL, (struct sljit_function_context*)code_ptr, (sljit_uw)code, (void*)sljit_generate_code);
```
zherczeg, line 664: I feel this patch is over complicated. My suggestion is add the size to code_ptr in this case, and the computation moved below is correct. Then no other change is needed.

carenas: While I agree it is more involved than needed and had indeed several bugs originally, I think the latest version is actually cleaner and simpler than the original.

Indeed it was the overloading of `code->size` part of the reason why it was far more complicated to understand and code than the s390x version, hence why I proposed that as (while at it) cleanup.

zherczeg: Adding the context size after `sljit_set_function_context(NULL, (struct sljit_function_context*)code_ptr, (sljit_uw)code, (void*)sljit_generate_code);` is still missing. It is something like:

`code_ptr += sizeof (struct sljit_function_context) / sizeof(type pointed by code_ptr)`

This way the extra allocation at the end is also included in the size.

zherczeg: A bit of background: function pointers on ppc may point to a descriptor structure, which contains the initial value of some global registers. These global values are not used by the jit compiler. However, I am not sure we correctly call functions from jit. We may not set these values as well. As long as the called function is in the same module, this is not a problem, but it can be when the application is big.

carenas: > `code_ptr += sizeof (struct sljit_function_context) / sizeof(type pointed by code_ptr)`

we can't do this, because `code_ptr` must point to the beginning of the trampoline at the end.

> This way the extra allocation at the end is also included in the size.

now you see why I did the "cleanup"; this code is confusing and fragile and I suspect still buggy.

sljit_src/sljitNativePPC_common.c
```diff
@@ -666,12 +666,18 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 
 	code_ptr = (sljit_ins *)SLJIT_ADD_EXEC_OFFSET(code_ptr, executable_offset);
 
-	SLJIT_CACHE_FLUSH(code, code_ptr);
-	SLJIT_UPDATE_WX_FLAGS(code, code_ptr, 1);
-
 #if (defined SLJIT_INDIRECT_CALL && SLJIT_INDIRECT_CALL)
+	compiler->executable_size = (sljit_uw)(code_ptr - code) * sizeof(sljit_ins) + sizeof(struct sljit_function_context) / sizeof(sljit_ins);
+	SLJIT_CACHE_FLUSH(code, code_ptr + sizeof(struct sljit_function_context) / sizeof(sljit_ins));
```
zherczeg, line 671: Why this change is added? The function context is read only memory structure, and it is never executed. The size computation here is a good idea.

sljit_src/sljitNativePPC_common.c
```diff
@@ -667,11 +667,18 @@ SLJIT_API_FUNC_ATTRIBUTE void* sljit_generate_code(struct sljit_compiler *compil
 	code_ptr = (sljit_ins *)SLJIT_ADD_EXEC_OFFSET(code_ptr, executable_offset);
 
 	SLJIT_CACHE_FLUSH(code, code_ptr);
-	SLJIT_UPDATE_WX_FLAGS(code, code_ptr, 1);
 
 #if (defined SLJIT_INDIRECT_CALL && SLJIT_INDIRECT_CALL)
+	compiler->executable_size = (sljit_uw)(code_ptr - code) * sizeof(sljit_ins) + sizeof(struct sljit_function_context);
+
+	SLJIT_UPDATE_WX_FLAGS(code, code_ptr + sizeof(struct sljit_function_context), 1);
```
zherczeg, line 674: The function context is readable only, no need to enable executable permission to it.

zherczeg (approved):
LGTM

PR #210, opened 2023-09-23
arm: small improvements for atomics with LSE

carenas:
Mostly to avoid some warning when building for aarch64 with `-march=armv8.1-a` or higher, but also adds test cases for bugs found in LoongArch and s390x implementations among other improvements.


zherczeg (approved):
LGTM

PR #212, opened 2023-09-27
atomic: avoid unused function declarations if not implemented

carenas:
Avoid spurious warnings when building in architectures that are yet to have an implementation for this experimental API


zherczeg:
Usually there are templates which returns with unsupported error.

zherczeg:
I can land this patch as well.

carenas:
while it would work AS-IS, and since it has been sitting here for so long, was actually planning on refactoring yet again

carenas:
shouldn't be needed anymore

PR #179, opened 2023-08-11
Be explicit about OS version

sevan:
via macports, allows pcre2 to build with jit support on OS X 10.4

PR #213, opened 2023-10-02
config: use standard interface to detct macOS version

carenas:
Implements a portability suggestion from #179

It is important to note that albeit undocumented, Apple compilers set the original macro as well, so this will only prevent the unnecessary fallback that was required for other compilers in that architecture.


zherczeg (approved):
LGTM

PR #214, opened 2023-10-03
allocator: add support for PowerPC in Apple

carenas:
needed at least for macOS versions older than 10.6 and corrects a regression from a previous refactoring that changed the intention of the code as documented to use MAP_JIT in non macOS if defined.


zherczeg:
I am sorry, I totally forgot about these patches. This looks ready, I can land it if you have no plans for other changes. Please ping me when I don't reply for a longer time.

zherczeg (approved):
LGTM

PR #215, opened 2023-11-01
Fix build error in MSVC

clover2123:


sljit_src/sljitLir.h
```diff
@@ -344,7 +344,7 @@ extern "C" {
 #define SLJIT_ARG_TYPE_SCRATCH_REG 0x8
 
 /* Void result, can only be used by SLJIT_ARG_RETURN. */
-#define SLJIT_ARG_TYPE_VOID	0
+#define SLJIT_ARG_TYPE_V	0
```
clover2123, line 347: When I try to build by msvc, `SLJIT_ARG_TYPE_void` is used instead of `SLJIT_ARG_TYPE_VOID`.
So, I change the name and it works.
I'm not sure the reason of error

zherczeg:
I have also noticed the windows errors, and started fixing them:
https://github.com/zherczeg/sljit/commit/e09319736d051a248dffdf81257162ea331a2b4f

I have added a V variant for the macros where the return value is void.

clover2123:
@zherczeg 
Thanks for the quick update!
I'll revise [walrus#182](https://github.com/Samsung/walrus/pull/182) based on it!

zherczeg:
Ok it should be simple. I can also do it if you prefer that way (but you can do it of course if you have the time).

zherczeg:
Btw I think the error is that windows defines a `VOID` macro as `void` for whatever reason.

PR #216, opened 2023-11-07
loongarch: sljit_emit_atomic_load/store implementation using LoongArch V1.10

lrzlin:
Loongson has released the LoongArch Version 1.10, which provides enhanced atomic instructions including `amcas.b/h/w/d`, this patch using these instructions to implement the `sljit_emit_atomic_load/store` functions. Meanwhile, for older processors' compatibility (3A5000 and current qemu), `cpucfg`  instruction was used for dynamically checking the cpu features.

LoongArch V1.10 Instruction Set Manual: https://www.loongson.cn/uploads/images/2023102309132647981.%E9%BE%99%E8%8A%AF%E6%9E%B6%E6%9E%84%E5%8F%82%E8%80%83%E6%89%8B%E5%86%8C%E5%8D%B7%E4%B8%80_r1p10.pdf


lrzlin:
Sorry for the Chinese version document, due to Loongson hasn't update their english document yet, so if you reviewers have any questions about these newer instructions, I'll answer them to the best of my ability.

test_src/sljitTest.c
```diff
@@ -11571,6 +11577,18 @@ static void test92(void)
 	sljit_sw buf[45];
 	sljit_s32 i;
 
+/* Check whether the processor supports LoongArch V1.10's AMCAS instruction*/
```
zherczeg, line 11580: I am not happy about this direction. Maybe we should introduce a feature for atomics, and if it is not available, skip the test. Then the large ifdef could go away.

lrzlin: Just like the SLJIT_HAS_SIMD I think? This maybe a good idea for both side.

zherczeg (approved):
LGTM

PR #217, opened 2023-11-12
x86: avoid clobbering "ebx register" during CPU detection (32-bit)

carenas:
Fixes a FTBFS when building a PCRE2 library that includes this change in 32bit Linux.

The problem might be only triggered by older compilers (using gcc 4.9.2-10+deb8u2) as using clang version 3.5.0-10 in the same system succeeds even without the patch

### DIFF
--- a/sljit_src/sljitNativeX86_common.c
+++ b/sljit_src/sljitNativeX86_common.c
@@ -426,9 +426,11 @@ static void execute_cpu_id(sljit_u32 info[4])
 		"movl %0, %%esi\n"
 		"movl (%%esi), %%eax\n"
 		"movl 8(%%esi), %%ecx\n"
+		"pushl %%ebx\n"
 		"cpuid\n"
 		"movl %%eax, (%%esi)\n"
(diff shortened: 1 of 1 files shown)


zherczeg:
Basically this is a workaround for old compilers? I also noticed that gcc inlines the function, and if it improperly support clobbered registers it would be a problem.

zherczeg:
Just one question: why the code works on 64 bit? And what about `esi`, which is also a saved register on 32 bit?

carenas:
I don't know if it works in 64 bit; don't have a system at hand to check that but AFAIK it is less likely to happen there because there are a lot more registers that can be used, and therefore less of a chance for conflict.

`esi` (as well as `ebx`) should be protected by the clobbered list, but are indeed also unnecessarily clobbered in this case.

this PR was meant to address this specific "Failure To Build From Source" that was IMHO more problematic that making the code more efficient, but I could amend it if desired, although I would rather do that change (that affects `esi`) independently.

zherczeg (approved):
LGTM

PR #218, opened 2023-11-14
x86: avoid hardcoding rsi in execute_cpu_id

carenas:
Simplifies the recently introduced `execute_cpu_id` and brings the logic a little closer to the original, while avoiding to hardcode `rsi`.

Tested in macOS (only 64bit available), Linux, NetBSD, OpenBSD, FreeBSD, Solaris and Windows, but it will likely need additional testing for other systems affected and maybe even require similar changes for at least the Windows specific paths.

### DIFF
--- a/sljit_src/sljitNativeX86_common.c
+++ b/sljit_src/sljitNativeX86_common.c
@@ -414,41 +414,40 @@ static SLJIT_INLINE void sljit_unaligned_store_sw(void *addr, sljit_sw value)
 
 static void execute_cpu_id(sljit_u32 info[4])
 {
+#if (defined(SLJIT_CONFIG_X86_32) && SLJIT_CONFIG_X86_32) && defined(__GNUC__)
+	register int *linfo __asm__ ("esi");
+#else
+	int *linfo;
+#endif
+	linfo = (int *)info;
+
 #if defined(_MSC_VER) && _MSC_VER >= 1400
 
-	__cpuidex((int*)info, (int)info[0], (int)info[2]);
+	__cpuidex(linfo, linfo[0], linfo[2]);
 
 #elif defined(__GNUC__) || defined(__INTEL_COMPILER) || defined(__SUNPRO_C) || defined(__TINYC__)
 
 	/* AT&T syntax. */
-	__asm__ (
+	__asm__ volatile (
 #if (defined SLJIT_CONFIG_X86_32 && SLJIT_CONFIG_X86_32)
-		"movl %0, %%esi\n"
-		"movl (%%esi), %%eax\n"
-		"movl 8(%%esi), %%ecx\n"
 		"pushl %%ebx\n"
-		"cpuid\n"
-		"movl %%eax, (%%esi)\n"
-		"movl %%ebx, 4(%%esi)\n"
-		"popl %%ebx\n"
-		"movl %%ecx, 8(%%esi)\n"
-		"movl %%edx, 12(%%esi)\n"
-#else /* !SLJIT_CONFIG_X86_32 */
-		"movq %0, %%rsi\n"
-		"movl (%%rsi), %%eax\n"
-		"movl 8(%%rsi), %%ecx\n"
-		"cpuid\n"
-		"movl %%eax, (%%rsi)\n"
-		"movl %%ebx, 4(%%rsi)\n"
-		"movl %%ecx, 8(%%rsi)\n"
-		"movl %%edx, 12(%%rsi)\n"
 #endif /* SLJIT_CONFIG_X86_32 */
-		:
-		: "r" (info)
+		"movl (%0), %%eax\n"
+		"movl 8(%0), %%ecx\n"
+		"cpuid\n"
+		"movl %%eax, (%0)\n"
+		"movl %%ebx, 4(%0)\n"
+		"movl %%ecx, 8(%0)\n"
+		"movl %%edx, 12(%0)\n"
 #if (defined SLJIT_CONFIG_X86_32 && SLJIT_CONFIG_X86_32)
-		: "memory", "eax", "ecx", "edx", "esi"
+		"popl %%ebx\n"
+#endif /* SLJIT_CONFIG_X86_32 */
+		: "=r" (linfo)
+		: "0" (linfo)
+#if (defined SLJIT_CONFIG_X86_32 && SLJIT_CONFIG_X86_32)
+		: "memory", "eax", "ecx", "edx"
 #else /* !SLJIT_CONFIG_X86_32 */
-		: "memory", "rax", "rbx", "rcx", "rdx", "rsi"
+		: "memory", "rax", "rbx", "rcx", "rdx"
 #endif /* SLJIT_CONFIG_X86_32 */
 	);
 
@@ -518,7 +517,7 @@ static void get_cpu_features(void)
 	}
 
 	info[0] = 0x80000001;
-	info[2] = 0; /* Silences an incorrect compiler warning. */
+	info[2] = 0; /* Silences a compiler warning. */
 	execute_cpu_id(info);
 
 	if (info[2] & 0x20)


sljit_src/sljitNativeX86_common.c
```diff
@@ -412,7 +412,7 @@ static SLJIT_INLINE void sljit_unaligned_store_sw(void *addr, sljit_sw value)
 /*    Utility functions                               */
 /******************************************************/
 
-static void execute_cpu_id(sljit_u32 info[4])
+static void execute_cpu_id(const sljit_u32 *info)
```
zherczeg, line 415: The block is replaced, it cannot be const.

sljit_src/sljitNativeX86_common.c
```diff
@@ -423,32 +423,24 @@ static void execute_cpu_id(sljit_u32 info[4])
 	/* AT&T syntax. */
 	__asm__ (
 #if (defined SLJIT_CONFIG_X86_32 && SLJIT_CONFIG_X86_32)
-		"movl %0, %%esi\n"
-		"movl (%%esi), %%eax\n"
-		"movl 8(%%esi), %%ecx\n"
 		"pushl %%ebx\n"
+#endif /* SLJIT_CONFIG_X86_32 */
+		"movl (%0), %%eax\n"
```
zherczeg, line 428: How do you ensure, that `%0` is not `eax`-`edx` ?

sljit_src/sljitNativeX86_common.c
```diff
@@ -484,16 +476,17 @@ static void get_cpu_features(void)
 {
 	sljit_u32 feature_list = CPU_FEATURE_DETECTED;
 	sljit_u32 info[4];
+	sljit_u32 *pinfo = info;
```
zherczeg, line 479: info is pointer, why do you need an alias?

sljit_src/sljitNativeX86_common.c
carenas: only the pointer is const, the data inside is changed and updated by both "cpuid" implementations AFAIK (tested also in Windows with MSVC)

sljit_src/sljitNativeX86_common.c
carenas: the info "label" is a pointer, but `info` is really an array and you need a pointer to it for the assembler to use the correct type on its options.

sljit_src/sljitNativeX86_common.c
carenas: eax, ecx and edx are not used because of the clobber, it could be ebx which then looks really funny, because it is first pushed to the stack by the epilogue and then again by the inlined assembler.

sljit_src/sljitNativeX86_common.c
zherczeg: My MSVC complained when there was no `(int*)` and `(int)` type casts.

sljit_src/sljitNativeX86_common.c
zherczeg: I don't know what is the best. But it looks like we cannot trust in the compiler too much. Somehow we can create fully asm functions, maybe that is the best.

sljit_src/sljitNativeX86_common.c
carenas: not sure what you mean, I removed the `int` casts in the last version, because the local alias should work as an alternative and didn't see any warnings, neither would expect any.

sljit_src/sljitNativeX86_common.c
```diff
@@ -518,7 +517,7 @@ static void get_cpu_features(void)
 	}
 
 	info[0] = 0x80000001;
-	info[2] = 0; /* Silences an incorrect compiler warning. */
+	info[2] = 0;
```
zherczeg, line 520: Why the comment is removed? It is a bug in my compiler, it thinks the variable is never set (which is a false assumption).

sljit_src/sljitNativeX86_common.c
```diff
@@ -414,41 +414,40 @@ static SLJIT_INLINE void sljit_unaligned_store_sw(void *addr, sljit_sw value)
 
 static void execute_cpu_id(sljit_u32 info[4])
 {
+#if (defined(SLJIT_CONFIG_X86_32) && SLJIT_CONFIG_X86_32) && defined(__GNUC__)
+	register int *linfo __asm__ ("esi");
```
zherczeg, line 418: This is too much magic for me. Why using esi is a problem? We could use the same was as ebx: push/pop and not mentioning it in the clobbered list. Is is possible to use the stack directly in asm? Then we could allocate two local ariables on the stack and use `[esp/rsp+offset]` to save/restore ebp/esp.

Another alternative: since `info[1]` is never used, we could use ebx instead of esp, and pushes it twice on the stack.

zherczeg (changes requested):

zherczeg:
Btw pusing/poping values may break the 16 byte stack alignment on 64 bit. Since we don't call any functions, this should not be a problem in practice, but worth noting.

sljit_src/sljitNativeX86_common.c
carenas: It is not a problem if used this way, the objective is that it does NOT use ebx, and both are possible since they are both valid on the ABI, the interesting part is that if they are not valid, then the compiler will save "esi" and restore automatically for us.

The problem with using it any other way, is that it can trigger bugs on the way the compiler allocates registers, since it is one of the ones it could chose to host the argument.

carenas:
> Btw pusing/poping values may break the 16 byte stack alignment on 64 bit

I think the original code had a comment about that but it was removed, that is also why the push/pop is compiled conditionally and will only be applied in 32bit

sljit_src/sljitNativeX86_common.c
zherczeg: We should reduce relying on the compiler, that is the main focus, if I understand correctly. Then be it minimal by manually saving/restoring everything by us, and tell nothing about it for the compiler.

sljit_src/sljitNativeX86_common.c
carenas: It was about it not being incorrect (under some circumstances), see the [pentium branch](https://github.com/carenas/sljit/tree/pentium) with one additional commit on top which might also do the initialization that was required.

sljit_src/sljitNativeX86_common.c
carenas: One way to do that is to rely more in the system provided calls (at least for Windows) and that is what the [mingw32 branch](https://github.com/carenas/sljit/tree/mingw32) was proposing.

When that is not possible, inlined assembler makes for the most optimal alternative IMHO.

zherczeg: We agree on that. I just want to remove the gcc specific part from the picture, and make the inlined assembly rely on minimal things. I still remember the TinyC issues.

zherczeg:
What do you think about using the same code as before jut using ebx instead of esi. The `info[1]` will not be filled, but it is not used. We just need to mention it somewhere. The intel assembly code could be updated as well.

carenas:
Not sure if I understand what you are suggesting, but `info[1]` is used to detect a couple of features that seem to be relevant to the rest of the codebase.

what is the point of using `ebx` instead of `esi`?, isn't it better to avoid `ebx`, because `cpuid` will clover it, regardless?

zherczeg:
I thought only 0/2/3 are used, but it seems yes, even 1 is used. Can we use a value on the stack in some way? Because we could save anything there, and use it for temporary storage. Or push both rbx/rsi onto the stack, and they don't need to be mentioned in the clobbered list.

carenas:
> Can we use a value on the stack in some way? Because we could save anything there, and use it for temporary storage

No; the original code got around it by NEVER using `ebx` and basically doing:

```
; setup eax and ecx for call
push ebx
cpuid
pop ebx
; extract values
```

sljit_src/sljitNativeX86_common.c
carenas: implemented differently with  #251

PR #219, opened 2023-11-17
s390x: fix FTBFS with clang

carenas:
While at it, add the relevant warning to the default list so this kind of issues are easier to detect.


zherczeg (approved):
LGTM

PR #220, opened 2023-11-24
loongarch: fix FTBFS when building as part of PCRE2

carenas:
When building aPCRE2 library that included these changes and that supports at least 2 different character sizes the following is shown:

```
/usr/bin/ld: ./.libs/libpcre2-16.a(libpcre2_16_la-pcre2_jit_compile.o): in function `.LANCHOR9':
pcre2_jit_compile.c:(.bss+0x160): multiple definition of `cpu_feature_list'; ./.libs/libpcre2-8.a(libpcre2_8_la-pcre2_jit_compile.o):pcre2_jit_compile.c:(.bss+0x160): first defined here
```


zherczeg (approved):
LGTM

PR #221, opened 2023-11-27
build: make code mostly ANSI C89 compatible

carenas:
While `-pedantic` is all that is needed together with an `-std` version; adding `-Wpedantic` allows for some "extra" and has been supported by `clang` for  a long while as well as by `gcc >= 4.8`.

The non C89 compliant parts of `s390x` could be addressed later, but are unlikely to be ever needed under the currently supported environments; the same goes for `ppc` where using a `void *` for a function is not valid C (because pointers to data and code can have different sizes), but is required for POSIX, and all systems using that CPU usually conform to that.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -7,16 +7,8 @@ CC = gcc
 endif
 endif
 
-ifndef EXTRA_CPPFLAGS
-EXTRA_CPPFLAGS=
-endif
-
-ifndef EXTRA_LDFLAGS
-EXTRA_LDFLAGS=
-endif
-
-ifndef EXTRA_LIBS
-EXTRA_LIBS=
+ifndef COMPAT_FLAGS
+COMPAT_FLAGS = -std=c89 -pedantic -Wpedantic
 endif
 
 ifndef OPT_FLAGS
@@ -32,7 +24,7 @@ WERROR = -Werror
 endif
 
 CPPFLAGS = $(EXTRA_CPPFLAGS) -Isljit_src
-CFLAGS += $(OPT_FLAGS) $(WARN_FLAGS) $(WERROR)
+CFLAGS += $(COMPAT_FLAGS) $(OPT_FLAGS) $(WARN_FLAGS) $(WERROR)
 REGEX_CFLAGS += $(CFLAGS) -fshort-wchar
 LDFLAGS = $(EXTRA_LDFLAGS)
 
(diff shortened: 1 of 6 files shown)


zherczeg:
What happens with the non conforming targets? Are they still compile with `-Werror`?

carenas:
Yes, but that can also "disabled" by calling `make WERROR=`, so you can always get it to build if you want.

I didn't want to complicate the logic on the Makefile unnecessarily

zherczeg:
I would not add it now, but when it works everywhere.

carenas:
Fair enough (indeed that is why I didn't propose it originally), but could we clarify what you mean by "everywhere", as well as a plan to move forward with the current issues that were being fixed?

zherczeg:
I am confused. You said it won't compile everywhere, so if this patch is merged, you need to manually edit the makefile on s390x, etc. just to compile the project.

carenas:
> ut enabling the feature might be too early. Or is it not?

guess the answer depends on what is meant by "feature".

for me was not too early, because the intention was (as spelled in the subject) and as a conclusion of the previous comments linked, to "fix the places where the code was accidentally made not C89 compatible and add a way to detect future mistakes to prevent regressions".

`s390x` was always meant to be C99 since it only supports running on systems that use C99 compilers by default and where making the code C89 would mean only less flexibility.

yes, it could make it slightly more difficult to build the code with really old compilers or in systems where the code was never ANSI C, but will hopefully prevent any more regressions.

zherczeg:
For me old compilers does not really matter, it is not really worth supporting more than 10-15 year old systems. High perf computing is not the focus on them. But developer convenience matters. I am still unsure what is the best option here.

zherczeg (approved):
Lets try this and see what happens. LGTM

zherczeg:
This patch killed the executable allocators. MAP_ANON is not defined anymore, and O_CLOEXEC as well. Removing `-std=c89` fixes this.

carenas:
Sorry, forgot to mention that was also expected, but frankly I didn't expect either that anyone would be using those with SLJIT's Makefile when they could as well change their invocation to be instead:

```
make EXTRA_CPPFLAGS=-DSLJIT_PROT_EXECUTABLE_ALLOCATOR COMPAT_FLAGS='-std=gnu89'
```

the reason why this will happen is that mmap itself is not C but POSIX compatible

zherczeg:
Maybe we should just add the warnings there which are important for us. Currently I have all kinds of compiling / running issues so I will disable it for now.

carenas:
> Maybe we should just add the warnings there which are important for us

does removing `-Wpedantic` help with your issues, or even `Wno-error=pedantic`?

zherczeg:
I just want it to compile out of the box. Adding the necessary warnings could help (e.g. a warning to not use `//` comments), and sill it would compile everywhere. You could use the stdc option from time to time to check if something wrong is introduced, and adding more warnings to prevent them.

carenas:
I think we are talking about the same but not understanding each other.

The way to generate warnings for issues with the C standard is `-pedantic`, sadly that has the side effect to stop the build because of `-Werror`, but if you add `-Wno-error=pedantic` then it will still be a visible warning and wouldn't stop the build (so you can check the logs periodically and make sure everything is clear)

I wasn't sure if the issues you were seeing was because of `-Wpedantic` and since that is only "nice to have" and if you don't want to add `-Wno-error=pedantic` then removig it would probably solve the issue without a full revert.

I am fine with whatever you would prefer anyway, as I said I ONLY proposed this change to provide a mechanism that would help identify early any non conformant code in the codebase, and knowing that there is already several places where we knowingly were not strictly compliant with the standards.

zherczeg:
My problems with warnings is that if new warnings appear we often don't notice in the long list of warnings. Currently gcc is very verbose, and for simple errors it displays pages of text, and it is hard to see that something new is there. For example, ppc64 complains now that functions cannot be converted to pointers, and do this with 3 pages of text (where a single line would be more than enough).

I think the best would be enabling warnings for certain constructs, which really should not be there. I never insisted on c89 compatiblity, but there are some things, such as // comments I would prefer not adding to the codebase.

PR #223, opened 2023-12-02
riscv: fix getput_arg and load_addr_to_reg

ksco:
This was discovered while running pcre2 tests on QEMU with Sv48 kernel. This 
patch fixes `getput_arg()` and `load_addr_to_reg()` for large immediate cases.
Tested with sljit tests and pcre2 tests.

Reported-by: Felix Yan <felixonmars@archlinux.org>
Reported-by: dramforever <dramforever@live.com>

### DIFF
--- a/sljit_src/sljitNativeRISCV_common.c
+++ b/sljit_src/sljitNativeRISCV_common.c
@@ -348,13 +348,12 @@ static SLJIT_INLINE void load_addr_to_reg(void *dst, sljit_u32 reg)
 		if ((addr & 0x80000000l) != 0)
 			high = ~high;
 
-		if ((high & 0x800) != 0)
-			high += 0x1000;
-
 		if (flags & PATCH_ABS52) {
 			SLJIT_ASSERT(addr <= S52_MAX);
 			inst[0] = LUI | RD(TMP_REG3) | (sljit_ins)(high << 12);
 		} else {
+			if ((high & 0x800) != 0)
+				high += 0x1000;
 			inst[0] = LUI | RD(TMP_REG3) | (sljit_ins)(high & ~0xfff);
 			inst[1] = ADDI | RD(TMP_REG3) | RS1(TMP_REG3) | IMM_I(high);
 			inst++;
@@ -940,7 +939,7 @@ static sljit_s32 getput_arg(struct sljit_compiler *compiler, sljit_s32 flags, sl
 
 	/* Since tmp can be the same as base or offset registers,
 	 * these might be unavailable after modifying tmp. */
-	if ((flags & MEM_MASK) <= GPR_REG && (flags & LOAD_DATA))
+	if ((flags & MEM_MASK) <= GPR_REG && (flags & LOAD_DATA) && reg == TMP_REG2)
 		tmp_r = reg;
 
 	if (SLJIT_UNLIKELY(arg & OFFS_REG_MASK)) {


ksco:
Oh wait, test19 case 6 failed, investigating...

ksco:
Ok, I reimplemented `load_immediate()` function, and all tests passed.

zherczeg:
Is this version generates more instructions in some cases? Currently 6 is the maximum number of instructions needed on 64 bit, which is unfortunately higher than any other cpu.

ksco:
> Which instruction cause this?

It's the `getput_arg` function, base register can be the same as `tmp_r`, so its value gets corrupted if `tmp_r` is used by `load_immediate()`.

> Another option is to fix the offending sljit instruction.

I did try this in the original patch https://github.com/zherczeg/sljit/commit/b0edde65dd669b5544cce002354cad640235a0e7. But later discovered that test19 case 6 is still failing for the same kind of reason. I noticed that we only have 3 scratch registers, so I think it might be a bit tight to have `load_immediate()` occupy one of them. That being said, I'm very new to the codebase, so any suggestions are welcome.

zherczeg:
Could you send em the line info where load_immediate is called? Also I would like a backtrace of how this happens. I am sure we can avoid using it as a base register.

zherczeg:
Sorry I mean it uses `load_immediate()`, but this is a load operation, so the base register can be overwritten. Store is a different thing, but for loads, this should not be an issue.

zherczeg:
So is this a bug in `getput_arg`? `SLJIT_MEM1(SLJIT_R0), (sljit_sw)&data` is incorrectly generated? The code is mostly copied from mips, where this kind of issue does not exists, but here, the immediate part should be constructed in tmp1/tmp2 and tmp3 registers.

ksco:
MIPS `load_immediate()` does not require a scratch register. As mentioned before, this patch https://github.com/zherczeg/sljit/commit/b0edde65dd669b5544cce002354cad640235a0e7 also does not work -- sljit_test goes further with it, but eventually test19 case 6 will fail.

zherczeg:
Could you try this check:

`if ((flags & MEM_MASK) <= GPR_REG && (flags & LOAD_DATA) && reg == TMP_REG2)`

ksco:
Thanks! I tried it, and it seems working. However, after running `sljit_test` a couple of times, it segfaulted in some of the runs. I can't reproduce it in gdb first, but after `set disable-randomization off`, it's happening in gdb too, so looks like it has something to do with ASLR maybe? The segfaulting address is strange, see debug log below. Just to be sure, I applied my patch and did all that again, and it's segfaulting the same way. So maybe it's not related to this bug.

Here is the debug log on your patch: https://ksco.cool/a/5GPX

ksco:
I'll debug this further to see the cause.

zherczeg:
Thank you. Is this an address in jit generated code? Is it always happens there or random location? Sljit has an mmap based stack allocation system, which can be called from jit code, and test38 test this. Maybe the `c` code returns to a wrong address.

ksco:
I think I found the issue:

```diff
diff --git a/sljit_src/sljitNativeRISCV_common.c b/sljit_src/sljitNativeRISCV_common.c
index 3bdd7ff..e78e245 100644
--- a/sljit_src/sljitNativeRISCV_common.c
+++ b/sljit_src/sljitNativeRISCV_common.c
@@ -348,13 +348,12 @@ static SLJIT_INLINE void load_addr_to_reg(void *dst, sljit_u32 reg)
                if ((addr & 0x80000000l) != 0)
                        high = ~high;
 
-               if ((high & 0x800) != 0)
-                       high += 0x1000;
-
                if (flags & PATCH_ABS52) {
                        SLJIT_ASSERT(addr <= S52_MAX);
                        inst[0] = LUI | RD(TMP_REG3) | (sljit_ins)(high << 12);
                } else {
+                       if ((high & 0x800) != 0)
+                               high += 0x1000;
                        inst[0] = LUI | RD(TMP_REG3) | (sljit_ins)(high & ~0xfff);
                        inst[1] = ADDI | RD(TMP_REG3) | RS1(TMP_REG3) | IMM_I(high);
                        inst++;
```

ksco:
Running `sljit_test` 500 times, all passed.

Can you commit your patch above? I'll close this one and then submit a new PR for above patch.

zherczeg:
Nice job! You can combine these two changes together and just open a PR for it.

ksco:
Okay, I force-pushed a new commit, PR title and description are also updated. Ready to go, thanks!

zherczeg (approved):
LGTM

PR #224, opened 2023-12-23
regex: allow building with a C++ compiler

carenas:
While at it, make sure the default "allocator" macros are safe to use in all configurations.


zherczeg (approved):
LGTM

PR #225, opened 2023-12-25
Add vzeroupper support for x86

chen-hu-97:
Refer to "15.3 MIXING AVX CODE WITH SSE CODE" Intel Software Optimization manual, "if software inter-mixes AVX and SSE instructions without using VZEROUPPER properly, it can experience an AVX/SSE transition penalty."

This patch add support for vzeroupper so JIT can emit this instruction and place it in correct place.

### DIFF
--- a/sljit_src/sljitNativeX86_common.c
+++ b/sljit_src/sljitNativeX86_common.c
@@ -966,6 +966,19 @@ static sljit_s32 emit_cmov_generic(struct sljit_compiler *compiler, sljit_s32 ty
 	sljit_s32 dst_reg,
 	sljit_s32 src, sljit_sw srcw);
 
+static SLJIT_INLINE sljit_s32 emit_vzeroupper(struct sljit_compiler *compiler)
+{
+       /* Emit vzeroupper.  */
+       sljit_u8 *inst;
+       inst = (sljit_u8*)ensure_buf(compiler, 1 + 3);
+       FAIL_IF(!inst);
+       INC_SIZE(3);
+       inst[0] = 0xc5;
+       inst[1] = 0xf8;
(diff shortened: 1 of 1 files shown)


chen-hu-97:
I am not sure if such a single patch, add support for one instruction on x86, can be a PR.
However, hope I can ramp up the repo and contribute to AVX2 code generation/emit in coming months.

zherczeg:
This kind of thing does not really fit to the concept of a generic jit compiler, because it more of a special casing for a specific issue. It would be interesting to explore other options as well, such as zeroing the target registers directly with a xor operation, since it could be generated without any extra api call. It is also a big question for me whether avx will ever be faster than sse2. If not, this direction does not really worth the effort.

carenas:
something that might be interesting would be to mix this logic with some other CPU specific codepaths that could be dynamically patched in.

I am sure (for example) that in really modern CPUs with a highly performant AVX-512 circuit the AVX2 code should be able to perform better than SSE2.

Indeed once CPUs with the next variable vector implementation are out, that dynamically changes the vector size, using AVX would be definitely be faster as it will allow also for AVX-512 to work without any changes on the code.

zherczeg:
For the next release I decided to use the SSE2 code path. But it would be good to improve the vector use. I tried some methods, such as zeroing the 256 bit register before using it, but it was still slow. And moving the vzeroupper to somewhere else also has no effect. Overall I don't really understand what is exactly happening here, which makes hard to maintain / redesign the code. If feels "fragile", and if you want to play around with some new idea, you might just break it, and you don't even understand why. On the long run we need to understand how the cpu thinks, and how can we exploit it.

PR #234, opened 2024-02-22
Fix leaks

vorosl:


test_src/sljitTestFloat.h
```diff
@@ -1243,7 +1243,7 @@ static void test_float12(void)
 {
 	/* Test floating point argument passing to sljit_emit_enter. */
 	executable_code code;
-	struct sljit_compiler* compiler = sljit_create_compiler(NULL, NULL);
+	struct sljit_compiler* compiler;
```
zherczeg, line 1246: I would delete the `compiler = sljit_create_compiler(NULL, NULL);` below

test_src/sljitTestFloat.h
```diff
@@ -343,7 +343,7 @@ static void test_float4(void)
 {
 	/* Test inline assembly. */
 	executable_code code;
-	struct sljit_compiler* compiler = sljit_create_compiler(NULL, NULL);
+	struct sljit_compiler* compiler;
```
zherczeg, line 346: Same here

zherczeg (changes requested):

vorosl:
Should I refactor all tests in this way?

zherczeg:
When reasonable.

PR #240, opened 2024-03-10
Fix build on older macOS

jmroot:
The headers on some older macOS versions don't define MAP_JIT at all.


zherczeg:
@carenas could you check this patch?

zherczeg:
Is the code still working on old mac systems? Is this x86 or ppc?

jmroot:
I haven't tried ppc but it works on old x86_64 macOS.

carenas (approved):

PR #242, opened 2024-03-16
Fix LoongArch floating point select tests

lrzlin:
Fix failed floating point select tests for LoongArch 64, tested under AOSC Linux running on Loongson 3A6000.


zherczeg (approved):
LGTM

PR #244, opened 2024-04-02
Fix locking region in sjlit_malloc_exec

lbonn:
The locked section needs to extend until we are done modifying internal block offsets and sizes, otherwise we risk running into data corruption in a multi-threaded context.

This bug was introduced in 7de0fee9a7376b33e31a3903487cd9933da3a606

See: https://github.com/PCRE2Project/pcre2/issues/402


zherczeg (approved):
LGTM

zherczeg:
Thank you!

PR #248, opened 2024-05-17
Fix TMP_REG2 usage fault in sljit_emit_simd_lane_mov for LoongArch

lrzlin:
According to what @carenas pointed, the LoongArch SIMD code will fail to complete test_simd9 case10, it's related to that in case10, we use `SLJIT_TMP_DEST_REG`, this reg is equal to reg `$ra` or `TMP_REG2` in LoongArch, and it's confilct to `sljit_emit_simd_lane_mov`, so replace it with `TMP_REG3`. 

This is due to I forget to update my codebase while developing LoongArch SIMD code, and the test_simd9 case10 wasn't there at the time. Again, big thanks to @carenas, and sorry for my carelessness.

### DIFF
--- a/sljit_src/sljitNativeLOONGARCH_64.c
+++ b/sljit_src/sljitNativeLOONGARCH_64.c
@@ -323,7 +323,7 @@ lower parts in the instruction word, denoted by the “L” and “H” suffixes
 /* Vector Instructions */
 
 /* Vector Arithmetic Instructions */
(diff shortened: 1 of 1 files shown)


zherczeg:
Would it be possible to use TMP_REG1? If TMP_REG2 is used for preferred temporary, TMP_REG1 is usually free

zherczeg (approved):
LGTM. Thank you for the patch

PR #250, opened 2024-06-25
x86: improve execute_get_xcr0_low() portability

carenas:
Address failure to build with [`tcc`](https://en.wikipedia.org/wiki/Tiny_C_Compiler) (and which could be argued is their fault anyway), but also prepare the code to be able to build when Intel Syntax is being used (ex: `gcc -masm=intel`) as well as improve compiler portability (Oracle Developer Studio C++ still needs and additional patch to build)

### DIFF
--- a/sljit_src/sljitNativeX86_common.c
+++ b/sljit_src/sljitNativeX86_common.c
@@ -401,6 +401,11 @@ static sljit_u32 cpu_feature_list = 0;
 #include <intrin.h>
 #endif
 
+#if (defined(_MSC_VER) && _MSC_VER >= 1400) || defined(__INTEL_COMPILER) \
+	|| (defined(__INTEL_LLVM_COMPILER) && defined(__XSAVE__))
+#include <immintrin.h>
+#endif
+
 /******************************************************/
 /*    Unaligned-store functions                       */
 /******************************************************/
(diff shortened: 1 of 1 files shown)


sljit_src/sljitNativeX86_common.c
```diff
@@ -496,35 +501,69 @@ static sljit_u32 execute_get_xcr0_low(void)
 {
 	sljit_u32 xcr0;
 
-#if defined(_MSC_VER) && _MSC_VER >= 1400
+#if (defined(_MSC_VER) && _MSC_VER >= 1400) || defined(__INTEL_COMPILER) \
+	|| (defined(__INTEL_LLVM_COMPILER) && defined(__XSAVE__))
```
zherczeg, line 505: Does `immintrin.h` defines  `__XSAVE__` ?

sljit_src/sljitNativeX86_common.c
```diff
@@ -496,35 +501,69 @@ static sljit_u32 execute_get_xcr0_low(void)
...
+		: "+m" (xcr0)
+		:
+#if defined(SLJIT_CONFIG_X86_32) && SLJIT_CONFIG_X86_32
+		: "eax", "ecx", "edx"
+#else /* !SLJIT_CONFIG_X86_32 */
+		: "rax", "rcx", "rdx"
+#endif /* SLJIT_CONFIG_X86_32 */
+	);
+
+#elif (defined(__INTEL_LLVM_COMPILER) && __INTEL_LLVM_COMPILER < 20220100) \
```
zherczeg, line 526: Why `< 20220100` is needed? Does intel supports AT&T

sljit_src/sljitNativeX86_common.c
carenas: no,  calling the compiler with`-mxsave` or an equivalent does.

`immintrin.h` is needed for the definition of the intrinsic call

sljit_src/sljitNativeX86_common.c
carenas: that is the version that adds support for writing [multiple syntax](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Multiple-assembler-dialects-in-asm-templates) that is used next, the non LLVM compiler (just like gcc) always had it AFAIK

zherczeg: This code guards an AT&T syntax code, and only for older "llvm intel" compilers. Is there such thing btw? I thought intel has its own compiler. If you tested it, then it is ok.

sljit_src/sljitNativeX86_common.c
zherczeg: Do we need `defined(__XSAVE__)` for the include above as well?

carenas: I thought there was no harm on having it included regardless since the file is there anyway, but will do after testing.

sljit_src/sljitNativeX86_common.c
carenas: the original intel compiler is now obsolete and has been replaced by the "llvm intel" compiler going forward.

AT&T syntax is supported by all compilers AFAIK

carenas: > the non LLVM compiler (just like gcc) always had it AFAIK

gcc < 3 doesn't support it (added to the list), so there might be also an ancient legacy intel compiler that would break, but considering that they are all obsolete already it should be ok

zherczeg:
Ok. If you have no plans for more changes, I can land the patch.

zherczeg (approved):
LGTM

PR #251, opened 2024-06-26
x86: improve execute_cpuid() portability

carenas:
Solves a failure to build when using Oracle Developer Studio C++ compiler
and simplifies the logic used in this function. finally making the codebase
safe to build with Intel syntax enforced.


zherczeg (approved):
LGTM

carenas:
FWIW it also fixes building with [pcc](https://en.wikipedia.org/wiki/Portable_C_Compiler)

PR #252, opened 2024-06-28
Add shuffle support for LoongArch

lrzlin:
Add 128-bit shuffle operation support to LoongArch and fix compilation failure due to API changes. Tested on Loongson 3C5000 & Gentoo Linux.

Plus, I've notice that the atomic part has been rewritten, dropped the `fine-grained` only tag. so if LoongArch needs to add the `lr/sc` atomic support? If so, I'll do it as soon as possible.

### DIFF
--- a/sljit_src/sljitNativeLOONGARCH_64.c
+++ b/sljit_src/sljitNativeLOONGARCH_64.c
@@ -348,6 +348,7 @@ lower parts in the instruction word, denoted by the “L” and “H” suffixes
 #define VREPLGR2VR OPC_2R(0x1ca7c0)
 #define VREPLVE OPC_3R(0xe244)
 #define VREPLVEI OPC_2R(0x1cbde0)
+#define VSHUF_B OPC_4R(0xd5)
 #define XVPERMI OPC_2RI8(0x1dfa)
 
 #define I12_MAX (0x7ff)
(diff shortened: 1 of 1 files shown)


lrzlin:
BTW, LoongArch do not have macros like _ARM_FEATURE_ATOMICS by now. So if I want to add lr/sc style atomic operation, I have to use cpucfg at runtime in case of checking whether amcas instruction extension is available or not. Is that OK?

carenas:
> use cpucfg at runtime

that would be similar to how `cpuid` is used in x86

lrzlin:
> > use cpucfg at runtime
> 
> that would be similar to how `cpuid` is used in x86

Thanks, I'll check how x86 use it.

zherczeg:
The lr/sc part is not necessary for all operations, the code generator can reject unsupported forms. Only the 32 bit / word sized operations are mandatory now.

zherczeg:
Patch looks good. Shall I land it?

lrzlin:
> The lr/sc part is not necessary for all operations, the code generator can reject unsupported forms. Only the 32 bit / word sized operations are mandatory now.

Got it, then I'll add 32 bit lr/sc support in a seperate patch for chips which doesn't support amcas. And how about this patch? Is there any issues?

lrzlin:
> Patch looks good. Shall I land it?

Please do so, thanks.

zherczeg (approved):
LGTM

zherczeg:
One more thing, the interface of simd instructions changed to use vreg (vector reg) instead of freg (float reg) because on some systems such as riscv the vector register set is independent from the float register set.

lrzlin:
> One more thing, the interface of simd instructions changed to use vreg (vector reg) instead of freg (float reg) because on some systems such as riscv the vector register set is independent from the float register set.

I've noticed that, and Is it necessary to replace all 'FRD/FRJ/FRK' macros with 'VRD/VRJ/VRK' macros?

zherczeg:
For private macros it is up to you. I didn't change them on some cpus, because they you a shared fpu/simd model.

carenas:
Unrelated, but noticed the following when testing in an emulated archlinux with `gcc -fsanitize=undefined`:

```
$ bin/sljit_test -s
sljit_src/sljitNativeLOONGARCH_64.c:1223:29: runtime error: signed integer overflow: -7520595660820047966 - 9161432420001660707 cannot be represented in type 'long int'
sljit_src/sljitNativeLOONGARCH_64.c:1217:48: runtime error: signed integer overflow: -7520595660820047966 - 9161432420001660928 cannot be represented in type 'long int'
SLJIT tests: all tests PASSED on LOONGARCH 64bit (little endian + unaligned) (with simd)
```

lrzlin:
> Unrelated, but noticed the following when testing in an emulated archlinux with `gcc -fsanitize=undefined`:
> 
> ```
> $ bin/sljit_test -s
> sljit_src/sljitNativeLOONGARCH_64.c:1223:29: runtime error: signed integer overflow: -7520595660820047966 - 9161432420001660707 cannot be represented in type 'long int'
> sljit_src/sljitNativeLOONGARCH_64.c:1217:48: runtime error: signed integer overflow: -7520595660820047966 - 9161432420001660928 cannot be represented in type 'long int'
> SLJIT tests: all tests PASSED on LOONGARCH 64bit (little endian + unaligned) (with simd)
> ```

I'll check this overflow error, thanks.

PR #254, opened 2024-06-30
loongarch: Implement ll/sc support for atomic operations

lrzlin:
Implement ll/sc for loongarch atomic load & store, also fix amcas bugs.
Tested under 3C5000 and 3A6000 (with amcas support).

### DIFF
--- a/sljit_src/sljitNativeLOONGARCH_64.c
+++ b/sljit_src/sljitNativeLOONGARCH_64.c
@@ -805,9 +805,6 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 	case SLJIT_HAS_SIMD:
 		return (LOONGARCH_HWCAP_LSX & get_cpu_features(GET_HWCAP));
 
-	case SLJIT_HAS_ATOMIC:
-		return (LOONGARCH_CFG2_LAMCAS & get_cpu_features(GET_CFG2));
-
 	case SLJIT_HAS_CLZ:
 	case SLJIT_HAS_CTZ:
 	case SLJIT_HAS_REV:
@@ -815,6 +812,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 	case SLJIT_HAS_PREFETCH:
 	case SLJIT_HAS_COPY_F32:
 	case SLJIT_HAS_COPY_F64:
+	case SLJIT_HAS_ATOMIC:
 		return 1;
 
 	default:
@@ -2645,10 +2643,8 @@ static sljit_ins get_jump_instruction(sljit_s32 type)
 {
 	switch (type) {
 	case SLJIT_EQUAL:
-	case SLJIT_ATOMIC_NOT_STORED:
 		return BNE | RJ(EQUAL_FLAG) | RD(TMP_ZERO);
 	case SLJIT_NOT_EQUAL:
-	case SLJIT_ATOMIC_STORED:
 		return BEQ | RJ(EQUAL_FLAG) | RD(TMP_ZERO);
 	case SLJIT_LESS:
 	case SLJIT_GREATER:
@@ -2656,6 +2652,7 @@ static sljit_ins get_jump_instruction(sljit_s32 type)
 	case SLJIT_SIG_GREATER:
 	case SLJIT_OVERFLOW:
 	case SLJIT_CARRY:
+	case SLJIT_ATOMIC_STORED:
 		return BEQ | RJ(OTHER_FLAG) | RD(TMP_ZERO);
 	case SLJIT_GREATER_EQUAL:
 	case SLJIT_LESS_EQUAL:
@@ -2663,6 +2660,7 @@ static sljit_ins get_jump_instruction(sljit_s32 type)
 	case SLJIT_SIG_LESS_EQUAL:
 	case SLJIT_NOT_OVERFLOW:
 	case SLJIT_NOT_CARRY:
+	case SLJIT_ATOMIC_NOT_STORED:
 		return BNE | RJ(OTHER_FLAG) | RD(TMP_ZERO);
 	case SLJIT_F_EQUAL:
 	case SLJIT_ORDERED_EQUAL:
@@ -2934,7 +2932,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_op_flags(struct sljit_compiler *co
 			break;
 		case SLJIT_ATOMIC_STORED:
 		case SLJIT_ATOMIC_NOT_STORED:
-			FAIL_IF(push_inst(compiler, SLTUI | RD(dst_r) | RJ(EQUAL_FLAG) | IMM_I12(1)));
+			FAIL_IF(push_inst(compiler, SLTUI | RD(dst_r) | RJ(OTHER_FLAG) | IMM_I12(1)));
 			src_r = dst_r;
 			invert ^= 0x1;
 			break;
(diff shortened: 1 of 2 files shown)


sljit_src/sljitNativeLOONGARCH_64.c
```diff
@@ -3652,16 +3681,45 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_store(struct sljit_compiler
 	CHECK_ERROR();
 	CHECK(check_sljit_emit_atomic_store(compiler, op, src_reg, mem_reg, temp_reg));
 
+	if (!(LOONGARCH_CFG2_LAMCAS & get_cpu_features(GET_CFG2))) {
+		switch (GET_OPCODE(op)) {
+		case SLJIT_MOV:
+		case SLJIT_MOV_P:
+			ins = SC_D;
+			break;
+		case SLJIT_MOV_S32:
```
carenas, line 3690: SLJIT_MOV_U32?

signed types should return SLJIT_ERR_UNSUPPORTED AFAIK

carenas (changes requested):

sljit_src/sljitNativeLOONGARCH_64.c
lrzlin: RISC-V and MIPS all use almost the same ll/sc for atomic load/store as LoongArch, and they all support signed types, so I think it wouldn't be an issue. BTW, I have noticed that in sljit_src/sljitNativeMIPS_common.c line 4234 has a SLJIT_CONFIG_RISCV_64 comment, should be a typo.  @zherczeg

carenas: > signed types should return SLJIT_ERR_UNSUPPORTED AFAIK

I stand corrected, apparently the "spec" was changed with 8bcd711708fd429eeefe24031c6a5f4decb2e4c5, so guess that the ones still rejecting signed types might be the ones that are not up to spec.

Eitherway you are still missing SLJIT_MOV_U32 there, right?

carenas (changes requested):
could you elaborate on the nature of the bugs that were fixed?

lrzlin:
> could you elaborate on the nature of the bugs that were fixed?

Originally, SLJIT_MOV_S32 was missing so that when it was used, amcas will fall to amcas.d and leading to an incorrect result.

sljit_src/sljitNativeLOONGARCH_64.c
lrzlin: Is it okay to use signed 32-bit ll/sc for 32-bit unsigned atomic load/store directly? I'm worry about boundary cases.

carenas: definitely not but is better than leaving it out as unsupported, and indeed I was surprised when it didn't break (as it did in a previous attempts to implement this as shown by #172).

which is why I was curious about the bug fixed, since it seems obvious that our test coverage needs some serious overhaul.

zherczeg:
Now the atomic instruction can tell the user whether it supports signed or unsigned operations (or both, but that is unlikely). Users can use sign/zero extension if they need it. MOV32 has no sign, it is up to the cpu, if registers are more than 32 bit long. riscv in mips is definitely a typo.

lrzlin:
> MOV32 has no sign, it is up to the cpu, if registers are more than 32 bit long. riscv in mips is definitely a typo.

Understand, I'll fix that typo, besides, are there still any problem for this patch?

sljit_src/sljitNativeLOONGARCH_64.c
```diff
@@ -805,16 +805,14 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
...
-		return (LOONGARCH_CFG2_LAMCAS & get_cpu_features(GET_CFG2));
-
 	case SLJIT_HAS_CLZ:
 	case SLJIT_HAS_CTZ:
 	case SLJIT_HAS_REV:
 	case SLJIT_HAS_ROT:
 	case SLJIT_HAS_PREFETCH:
 	case SLJIT_HAS_COPY_F32:
 	case SLJIT_HAS_COPY_F64:
+	case SLJIT_HAS_ATOMIC:
```
zherczeg, line 815: Am I see it correctly, that the cpu has new LL/SC form but CAS form is also supported? This is true for ARM64, so I am thinking whether we should allow users to pick which form they prefer. E.g. add a flag to pick the type, then return with unsupported if not available.

sljit_src/sljitNativeLOONGARCH_64.c
```diff
@@ -2934,7 +2932,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_op_flags(struct sljit_compiler *co
 			break;
 		case SLJIT_ATOMIC_STORED:
 		case SLJIT_ATOMIC_NOT_STORED:
-			FAIL_IF(push_inst(compiler, SLTUI | RD(dst_r) | RJ(EQUAL_FLAG) | IMM_I12(1)));
+			FAIL_IF(push_inst(compiler, SLTUI | RD(dst_r) | RJ(OTHER_FLAG) | IMM_I12(1)));
```
zherczeg, line 2935: Note: I changed it for convenience, you don't need to change it if using the equal flag is easier for the target cpu.

sljit_src/sljitNativeLOONGARCH_64.c
```diff
@@ -3618,14 +3616,42 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_load(struct sljit_compiler
 	CHECK_ERROR();
 	CHECK(check_sljit_emit_atomic_load(compiler, op, dst_reg, mem_reg));
 
+	if (!(LOONGARCH_CFG2_LAMCAS & get_cpu_features(GET_CFG2))) {
+		switch (GET_OPCODE(op)) {
+		case SLJIT_MOV:
+		case SLJIT_MOV_P:
+			ins = LL_D;
```
zherczeg, line 3623: I don't remember the current status, but I suspect LL_D is not available for 32 bit systems. Do we only support 64 bit?

lrzlin: By now, almost every LoongArch chip are 64-bit, AFAIK, only 1c series mcu are 32-bit, so I think it's safe to support 64-bit only. If there were 32-bit chips come into market, we could follow the way RISC-V did, split this into sljitNativeLOONGARCH_common.c and so-on.

sljit_src/sljitNativeLOONGARCH_64.c
lrzlin: Either way should be okay, but I think it's better to stay the same with other RISC architectures for maintainability.

sljit_src/sljitNativeLOONGARCH_64.c
lrzlin: LoongArch originally only support LL/SC, but add CAS support on LoongArch V1.10 (Currently only 3A6000 support this extension, QEMU do not support it yet). So it's safe to always use LL/SC, but CAS give more fine-grained atomic operations support. 

To allow user's selection, maybe we should add a type param to `sljit_emit_atomic_load` or `sljit_emit_atomic_store`, or we could add a `SLJIT_HAS_CAS` flag for it? But I still think there is not too much need to offer choices. Unless CAS is slower that LL/SC on some specific platform.

Reference: https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html#new-in-loongarch-v1.1

lrzlin: Update: AMCAS is faster than LL/SC on LoongArch

zherczeg: Ok. I will think about it.

sljit_src/sljitNativeLOONGARCH_64.c
```diff
@@ -3684,8 +3745,8 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_store(struct sljit_compiler
 	if (unsign)
```
zherczeg, line 3745: The unsign support can be removed in the future.

zherczeg (approved):
LGTM

PR #255, opened 2024-06-30
x86: check for extended function availability before using

carenas:
Protect against oldest CPUs that might leave bogus values after `cpuid` affecting the feature detection.

The "bug" for leaf 1 could affect several old CPUs as it matches the original specification for `cpuid`, possibly including:

> AMD (Am486 DX2/DX4, Am5x86, K5)
> Centaur (WinChip C6, Winchip 2)
> NexGen (Nx586)
> Cyrix (Cx486DX2/DX4, 5x86, 6x86, 6x86MX, MII)


zherczeg (approved):
LGTM

PR #256, opened 2024-07-01
x86: avoid SIGILL in CPUs with AVX but without AVX2

carenas:
Fix a SIGILL crash in machines with Intel's Sandy Bridge, Ivy Bridge as well as the three first generation of AMD Bulldozer and other equivalent CPUs

### DIFF
--- a/sljit_src/sljitNativeX86_common.c
+++ b/sljit_src/sljitNativeX86_common.c
@@ -3900,7 +3900,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_simd_replicate(struct sljit_compil
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


zherczeg:
This is a workaround not a fix. Which particular instruction form is not supported? AVX should support all SSE instructions, but the code generator might have issues.

carenas:
> Which particular instruction form is not supported?

```
vpbroadcastd %xmm4, %xmm4
```

(as explained in the commit message) and that is only available in AVX2 and has been guarded correctly to be only emitted when AVX2 is available everywhere else.

zherczeg:
The `use_vex` forces vex prefix for SSE instructions. This patch is the real fix.

zherczeg (approved):

PR #257, opened 2024-07-03
x86: improve SIMD detection

carenas:
Avoid SIGILL when AVX has been disabled by removing OSXSAVE, as well as when SSE2 is not enabled (most likely by running a really old OS in modern hardware)


sljit_src/sljitNativeX86_common.c
```diff
@@ -592,7 +593,11 @@ static void get_cpu_features(void)
 			feature_list |= CPU_FEATURE_LZCNT;
 	}
 
-	if ((feature_list & CPU_FEATURE_OSXSAVE) && (execute_get_xcr0_low() & 0x4) == 0)
+#if defined(SLJIT_CONFIG_X86_32) && SLJIT_CONFIG_X86_32
+	if (!(feature_list & CPU_FEATURE_SSE2))
+		feature_list &= ~(sljit_u32)CPU_FEATURE_SSE41;
```
zherczeg, line 598: Who sets this? Does the compiler runs on that old hardware?

carenas: Just like AVX2; SSE2 also needs OS support, which could be missing (ex: Debian 3).

When `gcc` targets a CPU with SSE2 it defines `__SSE2__` just like MSVC does with `_M_IX86_FP`

sljit_src/sljitNativeX86_common.c
```diff
@@ -572,12 +572,13 @@ static void get_cpu_features(void)
...
 		if (info[2] & 0x10000000)
 			feature_list |= CPU_FEATURE_AVX;
-#if (defined SLJIT_DETECT_SSE2 && SLJIT_DETECT_SSE2)
-		if (info[3] & 0x4000000)
-			feature_list |= CPU_FEATURE_SSE2;
-#endif
 		if (info[3] & 0x8000)
 			feature_list |= CPU_FEATURE_CMOV;
+#if (defined(SLJIT_DETECT_SSE2) && SLJIT_DETECT_SSE2) \
+	&& (defined(__SSE2__) || (defined(_WIN64) || (defined(_M_IX86_FP) && _M_IX86_FP >= 2)))
```
carenas, line 578: This will disable SSE2 with SUNPRO, TINYC or other compilers than won't provide the defines.

sljit_src/sljitNativeX86_common.c
zherczeg: The question was how is it possible to set SSE4.1 without setting SSE2. I always worry that we accidentally disables features for newer systems.

PR #258, opened 2024-07-16
Bitmanipulation instructions on RISC-V

vorosl:
clz, clzw, ctz, ctzw, rev8, rol, rolw, ror, rori, roriw, rorw, sext.b, sext.h, sh1add, sh2add, sh3add, zext.h

### DIFF
--- a/sljit_src/sljitNativeRISCV_common.c
+++ b/sljit_src/sljitNativeRISCV_common.c
@@ -88,6 +88,10 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define BGE		(F3(0x5) | OPC(0x63))
 #define BLTU		(F3(0x6) | OPC(0x63))
 #define BGEU		(F3(0x7) | OPC(0x63))
+#if defined __riscv_zbb
+#define CLZ		(F7(0x30) | F3(0x1) | OPC(0x13))
+#define CTZ		(F7(0x30) | F12(0x1) | F3(0x1) | OPC(0x13))
+#endif /* __riscv_zbb */
 #define DIV		(F7(0x1) | F3(0x4) | OPC(0x33))
 #define DIVU		(F7(0x1) | F3(0x5) | OPC(0x33))
 #define EBREAK		(F12(0x1) | F3(0x0) | OPC(0x73))
@@ -124,22 +128,48 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define ORI		(F3(0x6) | OPC(0x13))
 #define REM		(F7(0x1) | F3(0x6) | OPC(0x33))
 #define REMU		(F7(0x1) | F3(0x7) | OPC(0x33))
+#if defined __riscv_zbb
+#if defined SLJIT_CONFIG_RISCV_32
+#define REV8		(F12(0x698) | F3(0x5) | OPC(0x13))
+#elif defined SLJIT_CONFIG_RISCV_64
+#define REV8		(F12(0x6b8) | F3(0x5) | OPC(0x13))
+#endif
+#define ROL		(F7(0x30) | F3(0x1) | OPC(0x33))
+#define ROR		(F7(0x30) | F3(0x5) | OPC(0x33))
+#define RORI		(F7(0x30) | F3(0x5) | OPC(0x13))
+#endif /* __riscv_zbb */
 #define SC		(F7(0xc) | OPC(0x2f))
 #define SD		(F3(0x3) | OPC(0x23))
+#if defined __riscv_zbb
+#define SEXTB		(F7(0x30) | F12(0x4) | F3(0x1) | OPC(0x13))
+#define SEXTH		(F7(0x30) | F12(0x5) | F3(0x1) | OPC(0x13))
+#endif /* __riscv_zbb */
+#if defined __riscv_zba
+#define SH1ADD		(F7(0x10) | F3(0x2) | OPC(0x33))
+#define SH2ADD		(F7(0x10) | F3(0x4) | OPC(0x33))
+#define SH3ADD		(F7(0x10) | F3(0x6) | OPC(0x33))
+#endif /* __riscv_zba */
 #define SLL		(F7(0x0) | F3(0x1) | OPC(0x33))
-#define SLLI		(IMM_I(0x0) | F3(0x1) | OPC(0x13))
+#define SLLI		(F3(0x1) | OPC(0x13))
 #define SLT		(F7(0x0) | F3(0x2) | OPC(0x33))
 #define SLTI		(F3(0x2) | OPC(0x13))
 #define SLTU		(F7(0x0) | F3(0x3) | OPC(0x33))
 #define SLTUI		(F3(0x3) | OPC(0x13))
 #define SRL		(F7(0x0) | F3(0x5) | OPC(0x33))
-#define SRLI		(IMM_I(0x0) | F3(0x5) | OPC(0x13))
+#define SRLI		(F3(0x5) | OPC(0x13))
 #define SRA		(F7(0x20) | F3(0x5) | OPC(0x33))
-#define SRAI		(IMM_I(0x400) | F3(0x5) | OPC(0x13))
+#define SRAI		(F7(0x20) | F3(0x5) | OPC(0x13))
 #define SUB		(F7(0x20) | F3(0x0) | OPC(0x33))
 #define SW		(F3(0x2) | OPC(0x23))
 #define XOR		(F7(0x0) | F3(0x4) | OPC(0x33))
 #define XORI		(F3(0x4) | OPC(0x13))
+#if defined __riscv_zbb
+#if defined SLJIT_CONFIG_RISCV_32
+#define ZEXTH		(F7(0x4) | F3(0x4) | OPC(0x33))
+#elif defined SLJIT_CONFIG_RISCV_64
+#define ZEXTH		(F7(0x4) | F3(0x4) | OPC(0x3B))
+#endif
+#endif /* __riscv_zbb */
 
 #define SIMM_MAX	(0x7ff)
 #define SIMM_MIN	(-0x800)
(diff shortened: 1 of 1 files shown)


sljit_src/sljitNativeRISCV_common.c
```diff
@@ -70,6 +70,7 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define FRS2(rs2)	((sljit_ins)freg_map[rs2] << 20)
 #define IMM_I(imm)	((sljit_ins)(imm) << 20)
 #define IMM_S(imm)	((((sljit_ins)(imm) & 0xfe0) << 20) | (((sljit_ins)(imm) & 0x1f) << 7))
+#define SHAMT(shamt)	((sljit_ins)(shamt) << 20)
```
zherczeg, line 73: Seems the same as `IMM_I`

sljit_src/sljitNativeRISCV_common.c
```diff
@@ -88,6 +89,10 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define BGE		(F3(0x5) | OPC(0x63))
 #define BLTU		(F3(0x6) | OPC(0x63))
 #define BGEU		(F3(0x7) | OPC(0x63))
+#if defined __riscv_zbb
+#define CLZ		(F7(0x30) | F3(0x1) | OPC(0x13))
+#define CTZ		(IMM_I(0x601) | F3(0x1) | OPC(0x13))
```
zherczeg, line 94: IMM is not used for opcode definitions. Combine from F7 and `(((sljit_ins)0x...) << ...)`

sljit_src/sljitNativeRISCV_common.c
```diff
@@ -140,7 +164,13 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
 #define SW		(F3(0x2) | OPC(0x23))
 #define XOR		(F7(0x0) | F3(0x4) | OPC(0x33))
 #define XORI		(F3(0x4) | OPC(0x13))
-
+#if defined __riscv_zbb
+#if defined SLJIT_CONFIG_RISCV_32
+#define ZEXTH		(IMM_I(0x80) | F3(0x4) | OPC(0x33))
+#elif defined SLJIT_CONFIG_RISCV_64
+#define ZEXTH		(IMM_I(0x80) | F3(0x4) | OPC(0x3B))
```
zherczeg, line 171: Does the instruction form change? The register size should not matter.

sljit_src/sljitNativeRISCV_common.c
```diff
@@ -1043,10 +1073,24 @@ static sljit_s32 getput_arg(struct sljit_compiler *compiler, sljit_s32 flags, sl
 		next_arg = 0;
 		next_argw = 0;
 	}
-
```
zherczeg, line 1046: Dont't remove these lines.

sljit_src/sljitNativeRISCV_common.c
```diff
@@ -1043,10 +1073,24 @@ static sljit_s32 getput_arg(struct sljit_compiler *compiler, sljit_s32 flags, sl
...
+				FAIL_IF(push_inst(compiler, ADD | RD(tmp_r) | RS1(OFFS_REG(arg)) | RS2(base)));
+				break;
+			case 1:
+				FAIL_IF(push_inst(compiler, SH1ADD | RD(tmp_r) | RS1(OFFS_REG(arg)) | RS2(base)));
+				break;
+			case 2:
+				FAIL_IF(push_inst(compiler, SH2ADD | RD(tmp_r) | RS1(OFFS_REG(arg)) | RS2(base)));
+				break;
+			case 3:
+				FAIL_IF(push_inst(compiler, SH3ADD | RD(tmp_r) | RS1(OFFS_REG(arg)) | RS2(base)));
```
zherczeg, line 1090: These are essentially the same except the opcode changes. We could put the instruction into a local variable (`sljit_ins ins`) and call push_inst once.

zherczeg (changes requested):
Nice patch, only a few changes needed.

sljit_src/sljitNativeRISCV_common.c
vorosl: Yes.

https://drive.google.com/file/d/1uviu1nH-tScFfgrovvFCrj7Omv8tFtkp/view page 266.

sljit_src/sljitNativeRISCV_common.c
```diff
@@ -124,8 +128,27 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
...
+#define REV8		(IMM_I(0x6b8) | F3(0x5) | OPC(0x13))
+#endif
+#define ROL		(F7(0x30) | F3(0x1) | OPC(0x33))
+#define ROR		(F7(0x30) | F3(0x5) | OPC(0x33))
+#define RORI		((((sljit_ins)0x18) << 26) | F3(0x5) | OPC(0x13))
+#endif /* __riscv_zbb */
 #define SC		(F7(0xc) | OPC(0x2f))
 #define SD		(F3(0x3) | OPC(0x23))
+#if defined __riscv_zbb
+#define SEXTB		(IMM_I(0x604) | F3(0x1) | OPC(0x13))
```
zherczeg, line 144: These are also defined with IMM. Perhaps we should add another F macro F5, and combine these with F7, F5. This could be used for CTZ.

vorosl: I've found F12 macro

sljit_src/sljitNativeRISCV_common.c
```diff
@@ -124,23 +128,48 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 3] = {
...
 #define XOR		(F7(0x0) | F3(0x4) | OPC(0x33))
 #define XORI		(F3(0x4) | OPC(0x13))
-
+#if defined __riscv_zbb
+#if defined SLJIT_CONFIG_RISCV_32
+#define ZEXTH		(F7(0x4) | F3(0x4) | OPC(0x33))
+#elif defined SLJIT_CONFIG_RISCV_64
+#define ZEXTH		(F7(0x4) | F3(0x4) | OPC(0x3B))
+#endif
+#endif /* __riscv_zbb */
```
zherczeg, line 172: Place a newline after (removes the deleted newline above)

sljit_src/sljitNativeRISCV_common.c
```diff
@@ -1669,6 +1763,20 @@ static SLJIT_INLINE sljit_s32 emit_single_op(struct sljit_compiler *compiler, sl
 
 	case SLJIT_ROTL:
 	case SLJIT_ROTR:
+#if defined __riscv_zbb  
+		if (flags & SRC2_IMM) {
+			SLJIT_ASSERT(src2 != 0);
```
carenas, line 1768: wouldn't it be better to return SLJIT_SUCCESS for this?, also no need to duplicate logic IMHO

sljit_src/sljitNativeRISCV_common.c
```diff
@@ -1357,33 +1417,67 @@ static SLJIT_INLINE sljit_s32 emit_single_op(struct sljit_compiler *compiler, sl
...
 		SLJIT_ASSERT(src1 == TMP_ZERO && !(flags & SRC2_IMM));
+#if defined __riscv_zbb
+		FAIL_IF(push_inst(compiler, REV8 | RD(dst) | RS1(src2)));
+#if defined SLJIT_CONFIG_RISCV_64 && SLJIT_CONFIG_RISCV_64
+		return push_inst(compiler, (GET_OPCODE(op) == SLJIT_REV_U16 ? SRLI : SRAI )| RD(dst) | RS1(dst) | IMM_I(48));
+#else /* !SLJIT_CONFIG_RISCV_64 */
+		return push_inst(compiler, (GET_OPCODE(op) == SLJIT_REV_U16 ? SRLI : SRAI) | RD(dst) | RS1(dst) | IMM_I(16));
+#endif /* SLJIT_CONFIG_RISCV_64 */
+#else /* !__riscv_zbb */
 		return emit_rev16(compiler, op, dst, src2);
```
carenas, line 1469: this function shouldn't be needed for !__riscv_zbb, why is not #ifdef out like `emit_clz_ctz`?

zherczeg: Hm, the compiler should complain about that.

sljit_src/sljitNativeRISCV_common.c
zherczeg: How would you merge the code?

carenas: I meant that there is no need to have it twice if it is outside the `#ifdef`

sljit_src/sljitNativeRISCV_common.c
carenas: I think we do a lot of `static` functions with gcc that might just result on it not triggering, did you try `clang`?

sljit_src/sljitNativeRISCV_common.c
zherczeg: A few lines could be shared, but more don't. This is why I ask how could this look better?

sljit_src/sljitNativeRISCV_common.c
carenas: the `#ifdef` that I originally though was only for `emit_clz_ctz` actually covers all 3 functions, so this is a non issue.  should had pulled the code first to review instead of just relying on the github interface.

sljit_src/sljitNativeRISCV_common.c
```diff
@@ -1671,7 +1766,18 @@ static SLJIT_INLINE sljit_s32 emit_single_op(struct sljit_compiler *compiler, sl
...
-
+#if defined __riscv_zbb
+			if (GET_OPCODE(op) == SLJIT_ROTL) {
+#if (defined SLJIT_CONFIG_RISCV_64 && SLJIT_CONFIG_RISCV_64)
+				src2 = ((op & SLJIT_32) ? 32 : 64) - src2;
+#else /* !SLJIT_CONFIG_RISCV_64 */
+				src2 = 32 - src2;
+#endif /* SLJIT_CONFIG_RISCV_64 */
+			}
+			return push_inst(compiler, RORI | WORD | RD(dst) | RS1(src1) | IMM_I(src2));
```
zherczeg, line 1777: There should be an `#endif` here, to make the layout clearer.

zherczeg (approved):
LGTM

carenas (approved):
squashing the tip of https://github.com/carenas/sljit/tree/zb will be better IMHO

zherczeg (approved):
LGTM

PR #260, opened 2024-07-19
riscv: only enable SIMD if supported by system

carenas:
Avoid SIGILL when the CPU/OS/toolchain doesn't support the vector extension and that is fairly common in currently available systems.


zherczeg (approved):
LGTM

PR #261, opened 2024-07-21
Add SLJIT_ATOMIC_USE_LS/CAS support to LoongArch

lrzlin:
This fix test failure on LoongArch platform that support AMCAS. Tested under QEMU & Loongson 3A6000.

BTW, I do think adding a CI that tells us whether a PR will break some architecture or not is a need, which reduces our working strength, and also let us fix that failure in time.

### DIFF
--- a/sljit_src/sljitNativeLOONGARCH_64.c
+++ b/sljit_src/sljitNativeLOONGARCH_64.c
@@ -387,6 +387,8 @@ static sljit_u32 hwcap_feature_list = 0;
 #define GET_CFG2 	0
 #define GET_HWCAP	1
 
+#define LOONGARCH_SUPPORT_AMCAS (LOONGARCH_CFG2_LAMCAS & get_cpu_features(GET_CFG2))
+
 static SLJIT_INLINE sljit_u32 get_cpu_features(sljit_u32 feature_type)
  {
  	if (cfg2_feature_list == 0)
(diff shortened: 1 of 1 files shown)


sljit_src/sljitNativeLOONGARCH_64.c
```diff
@@ -3625,7 +3627,10 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_load(struct sljit_compiler
 	CHECK_ERROR();
 	CHECK(check_sljit_emit_atomic_load(compiler, op, dst_reg, mem_reg));
 
-	if (!(LOONGARCH_CFG2_LAMCAS & get_cpu_features(GET_CFG2))) {
+	if (op & SLJIT_ATOMIC_USE_CAS && !LOONGARCH_SUPPORT_AMCAS)
```
zherczeg, line 3630: I prefer `()` around & to make it clearer. Similar cases below.

zherczeg (changes requested):

sljit_src/sljitNativeLOONGARCH_64.c
lrzlin: Added.

sljit_src/sljitNativeLOONGARCH_64.c
```diff
@@ -3690,7 +3695,10 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_atomic_store(struct sljit_compiler
 	CHECK_ERROR();
 	CHECK(check_sljit_emit_atomic_store(compiler, op, src_reg, mem_reg, temp_reg));
 
-	if (!(LOONGARCH_CFG2_LAMCAS & get_cpu_features(GET_CFG2))) {
+	if ((op & SLJIT_ATOMIC_USE_CAS) && !LOONGARCH_SUPPORT_AMCAS)
```
zherczeg, line 3698: `LOONGARCH_SUPPORT_AMCAS` evaluated twice. `if (op & SLJIT_ATOMIC_USE_CAS) ..` would be moved into the if below, then one evaluation is enough.

lrzlin: Fixed, thanks.

zherczeg (approved):
LGTM

zherczeg:
Thank you!

Issue #262, opened 2024-08-12
Update loongarch


zherczeg:
@lrzlin I made some changes in the jit compiler. There is a new `SLJIT_MEMORY_BARRIER` operation which is a fence operation, and some changes were made in the atomic part. Could you update loongarch?

carenas:
the version of sljit included in latest PCRE2 shows the following warnings when building:

```
src/pcre2_jit_simd_inc.h:1868:9: warning: "VLD" redefined
 1868 | #define VLD           0x2c000000
      |         ^~~
In file included from src/../deps/sljit/sljit_src/sljitLir.c:3461,
                 from src/pcre2_jit_compile.c:81:
src/../deps/sljit/sljit_src/sljitNativeLOONGARCH_64.c:336:9: note: this is the location of the previous definition
  336 | #define VLD OPC_2RI12(0xb0)
      |         ^~~
src/pcre2_jit_simd_inc.h:1869:9: warning: "VOR_V" redefined
 1869 | #define VOR_V         0x71268000
      |         ^~~~~
src/../deps/sljit/sljit_src/sljitNativeLOONGARCH_64.c:330:9: note: this is the location of the previous definition
  330 | #define VOR_V OPC_3R(0xe24d)
      |         ^~~~~
src/pcre2_jit_simd_inc.h:1870:9: warning: "VAND_V" redefined
 1870 | #define VAND_V        0x71260000
      |         ^~~~~~
src/../deps/sljit/sljit_src/sljitNativeLOONGARCH_64.c:332:9: note: this is the location of the previous definition
  332 | #define VAND_V OPC_3R(0xe24c)
      |         ^~~~~~
src/pcre2_jit_simd_inc.h:1882:9: warning: "VREPLGR2VR" redefined
 1882 | #define VREPLGR2VR  0x729f0800
      |         ^~~~~~~~~~
src/../deps/sljit/sljit_src/sljitNativeLOONGARCH_64.c:352:9: note: this is the location of the previous definition
  352 | #define VREPLGR2VR OPC_2R(0x1ca7c0)
      |         ^~~~~~~~~~
```

carenas:
as discussed in PCRE2Project/pcre2#802, this should be fixed before the next scheduled release of PCRE2 (likely mid October)

zherczeg:
This should be fixed in pcre2 side.

PR #263, opened 2024-08-12
loongarch: add memory fence & support tmp register for atomic

lrzlin:
Implement LoongArch `dbar 0` memory fence, which is a full memory barrier just like `fence iorw, iorw` on RISC-V.

Also support tmp register for atomic, it worth noticing that `TMP_REG2` seems to have collision issues, so I use `TMP_REG3` instead.

Done: https://github.com/zherczeg/sljit/issues/262


zherczeg (approved):
LGTM

PR #264, opened 2024-08-17
Enable public usage of SLJIT_LOCALS_OFFSET

invertego:
Remove dependency on internal macro SSIZE_OF.


zherczeg (approved):
LGTM

zherczeg:
Just out of curiosity, how can you use this relative size? I am not aware of any operation where it is needed.

invertego:
I am using it to calculate the unadjusted stack pointer. I wanted a cheap way to determine where in generated code the last callout to C code occurred without needing to walk the stack or pay an additional cost for each call. On entry to generated code, I can do something like the following:
`sljit_get_local_base(compiler, SLJIT_MEM1(SLJIT_S0), offsetof(JitState, stackPointer), -SLJIT_LOCALS_OFFSET);`

Later, after calling out to C code, from an arbitrarily deep call stack I can retrieve the stored stack pointer and use it to determine where the call originated. On x86-64, the next 8 bytes below the stored pointer will hold the return address. I understand this trick can't be generalized to all architectures/ABIs (notably ARM64).

zherczeg:
This operation can also be used to get the return address:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitLir.h#L1478

PR #267, opened 2024-08-20
Emit brk #0xf000 for breakpoint opcode on arm64

### LINKED ISSUE #266: Breakpoint opcode encoding on ARM64
sljit currently implements the breakpoint opcode with `brk #0` on ARM64. The Windows kernel treats `brk` instructions differently depending on the immediate field. For `#0` it produces an illegal instruction exception. To produce a breakpoint exception (like `int3` does on x86) requires an immediate of `#0xf000`. With a debugger attached, you can also resume execution at the next instruction after a breakpoint exception (unlike illegal instruction exceptions).

LLVM 9 changed the implementation of its `debugtrap` intrinsic, exposed from clang as `__builtin_debugtrap()` and also as `__debugbreak()` (for compatibility with MSVC), to emit `brk #0xf000` for ARM64 Windows targets. In LLVM 12 this change was extended to all other ARM64 targets (for which it used to emit `brk #1`).

For sljit, I think it would be preferable for the breakpoint opcode be implemented with `brk #0xf000` on Windows. It would also be an option to follow LLVM's lead and make the change for all target OSes.

I could open a PR for either change, but I wanted to gauge reactions first.

References:
https://learn.microsoft.com/en-us/cpp/intrinsics/debugbreak?view=msvc-170
https://devblogs.microsoft.com/oldnewthing/20220822-00/?p=107032
https://reviews.llvm.org/D63635
https://reviews.llvm.org/D84014

invertego:
Implement SLJIT_BREAKPOINT with the same instruction emitted by MSVC for the __debugbreak intrinsic (brk #0xf000) and by Clang for __builtin_debugtrap. On Windows, this triggers a breakpoint exception just as int3 does on x86. The previously emitted instruction (brk #0) triggers an illegal instruction exception instead.

LLDB also has special handling for this brk variant on Windows and macOS, automatically advancing to the next instruction so execution can be easily continued.

Fixes #266

### DIFF
--- a/sljit_src/sljitNativeARM_64.c
+++ b/sljit_src/sljitNativeARM_64.c
@@ -1548,7 +1548,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_op0(struct sljit_compiler *compile
 	op = GET_OPCODE(op);
 	switch (op) {
 	case SLJIT_BREAKPOINT:
-		return push_inst(compiler, BRK);
+		return push_inst(compiler, BRK | (0xf000 << 5));
 	case SLJIT_NOP:
 		return push_inst(compiler, NOP);
 	case SLJIT_LMUL_UW:


invertego:
I can move the immediate directly into the definition of BRK if desired. I did it this way to make the choice of immediate more readily apparent.

zherczeg (approved):
LGTM

zherczeg:
The hint is not part of the instruction, its an argument, so it is better this way.

Issue #272, opened 2024-11-01
Spurious assert on x86_64


invertego:
When generating x86_64 code for a function that contains a mix of branches to a label and to a target address, the following assert may fire spuriously:

```
Assertion failed at sljit_src/sljitNativeX86_common.c:955
SLJIT_ASSERT((sljit_uw)code_ptr - addr <= ((jump->flags >> JUMP_SIZE_SHIFT) & 0x1f));
```

Here's a self-contained repro:
```
void repro(void)
{
	struct sljit_compiler* compiler;
	struct sljit_jump* jump;
	struct sljit_label* label;
	void* nullsub;
	void* code;

	// emit function that simply returns
	compiler = sljit_create_compiler(NULL);
	sljit_compiler_verbose(compiler, stdout);
	sljit_emit_enter(compiler, 0, SLJIT_ARGS0V(), 0, 0, 0);
	sljit_emit_return_void(compiler);
	nullsub = sljit_generate_code(compiler, 0, NULL);
	sljit_free_compiler(compiler);

	// emit function that calls nullsub
	compiler = sljit_create_compiler(NULL);
	sljit_compiler_verbose(compiler, stdout);
	sljit_emit_enter(compiler, 0, SLJIT_ARGS0V(), 0, 0, 0);

	// emit many calls to nullsub
	for (int i = 0; i < 16; i++)
		sljit_emit_icall(compiler, SLJIT_CALL, SLJIT_ARGS0V(), SLJIT_IMM, SLJIT_FUNC_ADDR(nullsub));

	// emit jump to next instruction (return)
	jump = sljit_emit_jump(compiler, SLJIT_JUMP);
	label = sljit_emit_label(compiler);
	sljit_set_label(jump, label);	
	sljit_emit_return_void(compiler);

	// assert fires here
	code = sljit_generate_code(compiler, 0, NULL);
	sljit_free_compiler(compiler);

	sljit_free_code(nullsub, NULL);
	sljit_free_code(code, NULL);
}
```

From a quick look at the implementation, this seems to be the result of `detect_near_jump_type()` using `label->size` when calculating the branch distance. For a forward branch, this value will still be just an estimate produced by `reduce_code_size()`. With many preceding branches to a target address, this value could be a large overestimate if the actual size of those branches is much less than `JUMP_MAX_SIZE`. As a result, the distance between the actual source address and estimated label address can be larger than the branch distance estimated by `reduce_code_size()`.

zherczeg:
[...]

PR #273, opened 2024-11-02
loongarch: Optimize jump maximum offset prediction & update action co…

lrzlin:
Optimize LoongArch jump maximum offset prediction just like RISC-V and replace the out-dated compiler link used in github actions. Tested on 3A6000.

Done: https://github.com/zherczeg/sljit/issues/272#issuecomment-2452975704

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -237,8 +237,8 @@ jobs:
       run: |
           sudo apt update
           sudo apt install -y make wget ninja-build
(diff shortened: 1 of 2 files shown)


zherczeg:
The patch is ok, thank you!

@carenas the CI says things like:
ERROR: failed to solve: arm32v7/ubuntu:latest: failed to resolve source metadata for docker.io/arm32v7/ubuntu:latest: no match for platform in manifest: not found
make: ./cross-tools/bin/loongarch64-unknown-linux-gnu-gcc-14.1.0: No such file or directory

Do you have an idea what is happening?

carenas:
The loongarch specific issue should be fixed by using the correct gcc binary from the new bundle as shown by fbbc169c26655d2e0c87fd6862491c92913f85aa

zherczeg:
Thank you for the quick update

Issue #232, opened 2024-02-08
Update loongarch


zherczeg:
@lrzlin some updates are needed in the loongarch code after the recent changes. There is a new serializer interface, and `label->addr` needs to be changed to `label->u.addr`.

This update is not necessary, but recommended. Many platforms has a `reduce_code_size` function now:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitNativeRISCV_common.c#L405

This function tries to reduce the space allocated for branches, which improves branch construction and reduces code size. On the regex benchmark the code size reduction is around 30-50% on 64 bit, and 20-40% on 32 bit cpus, so this is quite significant saving.

lrzlin:
OK, I'll working on it as soon as possible, thanks for remind me that.

zherczeg:
@lrzlin I have added a few macros to specify how the upper 32 bit is used by 32 bit operations on 64 bit:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitConfigInternal.h#L53

Could you update LoongArch? Should be a minor thing. Thank you!

PR #276, opened 2024-11-19
loongarch: Add SLJIT_UPPER_BITS_SIGN_EXTENDED

lrzlin:
Notice that in LoongArch V1.1 (LA664 and later), the SLJIT_UPPER_BITS_IGNORED should also be true, however due to a micro-architecture bug, 3A5000 (LA464) will needs the upper 32 bits of source register is sign extended while doing 32 bits division (e.g. div.w), otherwise it will output garbage result.

See: https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html Chapter 1.5.1 Item 7
Also: https://areweloongyet.com/newsletter/this-week-in-loongarch-22 (Chinese Only)

Done: https://github.com/zherczeg/sljit/issues/232


zherczeg (approved):
LGTM, thank you

PR #284, opened 2024-12-19
Fix sljit_emit_simd_sign RISCV

vorosl:

PR #286, opened 2025-01-05
Enable -Wshadow warning

NWilson:
PCRE2 uses `-Wshadow`, so it's helpful if the sljit code does as well.

Some errors seem to have crept in to the Mac-only (or rather, ARM-only) code which wasn't noticed until I tried a PCRE2 build on that platform, in one of the build configurations using `-Wshadow`.

### DIFF
--- a/sljit_src/sljitNativeS390X.c
+++ b/sljit_src/sljitNativeS390X.c
@@ -169,10 +169,10 @@ static SLJIT_INLINE sljit_u8 get_cc(struct sljit_compiler *compiler, sljit_s32 t
 	switch (type) {
 	case SLJIT_EQUAL:
 		if (SLJIT_ADD_SUB_NO_COMPARE(compiler->status_flags_state)) {
-			sljit_s32 type = GET_FLAG_TYPE(compiler->status_flags_state);
-			if (type >= SLJIT_SIG_LESS && type <= SLJIT_SIG_LESS_EQUAL)
+			sljit_s32 flag_type = GET_FLAG_TYPE(compiler->status_flags_state);
+			if (flag_type >= SLJIT_SIG_LESS && flag_type <= SLJIT_SIG_LESS_EQUAL)
 				return cc0;
-			if (type == SLJIT_OVERFLOW)
+			if (flag_type == SLJIT_OVERFLOW)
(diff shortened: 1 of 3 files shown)


zherczeg:
The build updates from  #284 might be needed here as well.
Anyway, s390x also has some shadowed local variables.

NWilson:
Thanks, I've pulled Vörös's build fix and attempted a fix for the S390 shadowing.

sljit_src/sljitNativeS390X.c
```diff
@@ -2243,28 +2243,28 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_op1(struct sljit_compiler *compile
 		}
 		/* STORE and STORE IMMEDIATE */
 		if ((dst & SLJIT_MEM) && (FAST_IS_REG(src) || src == SLJIT_IMM)) {
-			struct addr mem;
+			struct addr mem2;
```
zherczeg, line 2246: Hm, does this local variable even needed? It seems this was a mistake in the original code.

NWilson: Oh I see, the old code was vaguely evil. I updated all instances of `mem → mem2` within the scope of the variable... but it's sneakily accessed via `LEVAL` which is defined higher-up.

I don't really understand the original code very well, which is always a bad thing when making changes! I'll push a commit which attempts to simply remove the shadowed `struct addr mem` and see if the CI passes.

NWilson: I've thought a bit more about the code.

I think it's totally OK to remove the shadowed copies of `mem`, as I have done in the latest revision.

Reasoning: the `struct addr mem` is totally clobbered by make_addr_bxy at the start of every block of code that uses `mem`. All the fields are fully-overwritten each time it's used, so re-using a common object should not cause any problems. I can't see any gotos or other hazards.

zherczeg:
#284 landed, please update the patch

NWilson:
Thanks @zherczeg, done.

zherczeg (approved):
LGTM

zherczeg:
Thank you!

PR #287, opened 2025-01-07
misc: avoid unnecesary shadowed variable

carenas:
Found when building the latest PCRE2 RC with gcc in macOS


zherczeg:
Is this related to #286 or something different?

carenas:
> Is this related to #286 or something different?

it is fundamentally the same issue (sorry, only found out there was one, after this one was already out), but the solution is different and I think could be complimentary.

zherczeg:
I think `total_size` global now got a new name. Are these longer names still necessary?

carenas:
a longer (more descriptive) name is always better IMHO, and changing the one used locally in all those RISC CPU files which seem to be somehow related is good on itself, but you are correct that with the renaming of the global (which hopefully is not being used outside our codebase and therefore doesn't "break ABI") it is no longer _needed_ to avoid `-Wshadow` related warnings.

PR #294, opened 2025-01-21
Implement 16 bit branches for RISCV

zherczeg:


carenas (approved):
Tested in Spacemit(R) X60

zherczeg:
Do you have access to RISCV with 16 bit instruction set? That is great! Thank you!

PR #296, opened 2025-01-24
Implement emit_op2cmp operation

zherczeg:


zherczeg:
@lrzlin this patch fails on loongarch. Loongarch could also use the riscv variant of the code.

lrzlin:
> @lrzlin this patch fails on loongarch. Loongarch could also use the riscv variant of the code.

OK, I'll give it a look.

zherczeg:
Thank you

PR #302, opened 2025-02-06
Implement aligned labels

zherczeg:


zherczeg:
@lrzlin this is another patch which affects loongarch. The API is probably not final, I have some other plans with it, but not in a single patch.

PR #313, opened 2025-03-14
Fix sljit_emit_icall on Linux x86_64

partic2:
In current version SLJIT, sljit_emit_icall only save "RDI" if word_arg_count >= 3. But RDI is the first argument according to Linux x86_64 calling convention,

in other words, below code will cause SegFault on current version SLJIT.

```c
	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_IMM, 10);
	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R1, 0, SLJIT_IMM, 16);
	sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R2, 0, SLJIT_IMM, SLJIT_FUNC_ADDR(func2));
	sljit_emit_icall(compiler, SLJIT_CALL, SLJIT_ARGS2(W, W, W), SLJIT_R2, 0);
```

BTW sljit  failed on windows in test_float13, with large stack test. 
Is it known issue? Should we decrease the SLJIT_MAX_LOCAL_SIZE on windows?

### DIFF
--- a/sljit_src/sljitNativeX86_64.c
+++ b/sljit_src/sljitNativeX86_64.c
@@ -805,7 +805,7 @@ static sljit_s32 call_with_args(struct sljit_compiler *compiler, sljit_s32 arg_t
 	if (word_arg_count == 0)
 		return SLJIT_SUCCESS;
 
-	if (word_arg_count >= 3) {
+	if (word_arg_count >= 3 || src == SLJIT_R2) {
 		if (src == SLJIT_R2)
 			*src_ptr = TMP_REG1;
 		EMIT_MOV(compiler, TMP_REG1, 0, SLJIT_R2, 0);
--- a/test_src/sljitTestCall.h
+++ b/test_src/sljitTestCall.h
@@ -34,13 +34,18 @@ static sljit_sw func4(sljit_sw a, sljit_sw b, sljit_sw c, sljit_sw d)
 	return func(a, b, c) + d;
 }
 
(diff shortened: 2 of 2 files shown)


sljit_src/sljitNativeX86_64.c
```diff
@@ -805,7 +805,7 @@ static sljit_s32 call_with_args(struct sljit_compiler *compiler, sljit_s32 arg_t
 	if (word_arg_count == 0)
 		return SLJIT_SUCCESS;
 
-	if (word_arg_count >= 3) {
+	if (word_arg_count >= 1) {
```
zherczeg, line 808: Shouldn't it be: `(word_arg_count >= 3 || src == SLJIT_R2)`. Perhaps an `xchg` could be used to use a single mov.
Could you add a test as well?

zherczeg:
On Windows, sljit cannot control the maximum stack size, it needs to be set by the application.
In Visual Studio for example: PROJECT->Properties->Configuration Properties->Linker->System->Stack Reserve Size

https://stackoverflow.com/questions/40157847/increase-stack-size-in-c

partic2:
You are right. I send a new patch now.
I'm not familiar with the source of sljit, So I can't find out where to use **xchg** properly.

Thanks for your patient explanation about Windows.

zherczeg:
Ok, I will do a follow-up patch then.

zherczeg (approved):
LGTM

PR #314, opened 2025-03-14
loongarch: Update toolchain and minor code style change

lrzlin:
Update toolchain to the lastest 2025.02.21 version & some minor code style change to maintain the consistency of codebase.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -49,7 +49,7 @@ jobs:
               target: riscv64-linux-user
               name: qemu-riscv64
           - name: riscv64
-            toolchain: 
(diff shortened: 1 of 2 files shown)


zherczeg (approved):
LGTM

zherczeg:
Btw I noticed the immediate generator could be improved.

For example:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitNativeLOONGARCH_64.c#L911
If the lowest 12 bit is zero, that is a nop.

Riscv also do shifted imm32 computations, so a `0xffffff00000` is generated as `0xffffff << 20`. But this might be less interesting here.

PR #315, opened 2025-03-15
loongarch: improve immediate generator

lrzlin:
Stop emitting `ori` while the immediate's lower 12-bit is zero in immediate generator.

### DIFF
--- a/sljit_src/sljitNativeLOONGARCH_64.c
+++ b/sljit_src/sljitNativeLOONGARCH_64.c
@@ -908,14 +908,18 @@ static sljit_s32 load_immediate(struct sljit_compiler *compiler, sljit_s32 dst_r
 
 	if (imm <= 0x7fffffffl && imm >= -0x80000000l) {
 		FAIL_IF(push_inst(compiler, LU12I_W | RD(dst_r) | (sljit_ins)(((imm & 0xffffffff) >> 12) << 5)));
-		return push_inst(compiler, ORI | RD(dst_r) | RJ(dst_r) | IMM_I12(imm));
+		if (IMM_I12(imm) != 0)
+			return push_inst(compiler, ORI | RD(dst_r) | RJ(dst_r) | IMM_I12(imm));
+		return SLJIT_SUCCESS;
 	} else if (imm <= 0x7ffffffffffffl && imm >= -0x8000000000000l) {
 		FAIL_IF(push_inst(compiler, LU12I_W | RD(dst_r) | (sljit_ins)(((imm & 0xffffffff) >> 12) << 5)));
-		FAIL_IF(push_inst(compiler, ORI | RD(dst_r) | RJ(dst_r) | IMM_I12(imm)));
+		if (IMM_I12(imm) != 0)
+			FAIL_IF(push_inst(compiler, ORI | RD(dst_r) | RJ(dst_r) | IMM_I12(imm)));
 		return push_inst(compiler, LU32I_D | RD(dst_r) | (sljit_ins)(((imm >> 32) & 0xfffff) << 5));
 	}
 	FAIL_IF(push_inst(compiler, LU12I_W | RD(dst_r) | (sljit_ins)(((imm & 0xffffffff) >> 12) << 5)));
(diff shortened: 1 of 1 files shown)


sljit_src/sljitNativeLOONGARCH_64.c
```diff
@@ -907,15 +907,21 @@ static sljit_s32 load_immediate(struct sljit_compiler *compiler, sljit_s32 dst_r
 		return push_inst(compiler, ADDI_D | RD(dst_r) | RJ(TMP_ZERO) | IMM_I12(imm));
 
 	if (imm <= 0x7fffffffl && imm >= -0x80000000l) {
-		FAIL_IF(push_inst(compiler, LU12I_W | RD(dst_r) | (sljit_ins)(((imm & 0xffffffff) >> 12) << 5)));
-		return push_inst(compiler, ORI | RD(dst_r) | RJ(dst_r) | IMM_I12(imm));
+		if (IMM_I12(imm) != 0) {
```
zherczeg, line 910: I think this could look the same as below, only put ORI behind an `if`.

sljit_src/sljitNativeLOONGARCH_64.c
```diff
@@ -907,15 +907,21 @@ static sljit_s32 load_immediate(struct sljit_compiler *compiler, sljit_s32 dst_r
...
-		FAIL_IF(push_inst(compiler, ORI | RD(dst_r) | RJ(dst_r) | IMM_I12(imm)));
+		if (IMM_I12(imm) != 0)
+			FAIL_IF(push_inst(compiler, ORI | RD(dst_r) | RJ(dst_r) | IMM_I12(imm)));
 		return push_inst(compiler, LU32I_D | RD(dst_r) | (sljit_ins)(((imm >> 32) & 0xfffff) << 5));
 	}
 	FAIL_IF(push_inst(compiler, LU12I_W | RD(dst_r) | (sljit_ins)(((imm & 0xffffffff) >> 12) << 5)));
-	FAIL_IF(push_inst(compiler, ORI | RD(dst_r) | RJ(dst_r) | IMM_I12(imm)));
+	if (IMM_I12(imm) != 0)
+		FAIL_IF(push_inst(compiler, ORI | RD(dst_r) | RJ(dst_r) | IMM_I12(imm)));
 	FAIL_IF(push_inst(compiler, LU32I_D | RD(dst_r) | (sljit_ins)(((imm >> 32) & 0xfffff) << 5)));
```
zherczeg, line 925: `(imm >> 32) & 0xfffff)` could also be 0.

sljit_src/sljitNativeLOONGARCH_64.c
```diff
@@ -907,15 +907,21 @@ static sljit_s32 load_immediate(struct sljit_compiler *compiler, sljit_s32 dst_r
...
 	return push_inst(compiler, LU52I_D | RD(dst_r) | RJ(dst_r) | IMM_I12(imm >> 52));
```
zherczeg, line 926: These instructions modify certain bits while keeping others, am I right? So `dst` needs to be set to something before they are used.

lrzlin: According to LoongArch Instruction Manual, `lu12i.w` will clear the lower 12-bits of dst and sign-extend the imm, so we could just use `dst` directly.

Reference: https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html#_lu12i_w_lu32i_d_lu52i_d

sljit_src/sljitNativeLOONGARCH_64.c
lrzlin: `lu12i.w` will sign-extend the immediate to 64-bit and write it to `dst`, so I think the `lu32.d` there is a must?

zherczeg: I thought LU52I_D is must, and LU32I_D is optional in this case, but I don't know .the cpu.

zherczeg:
Do you plan any more changes?

lrzlin:
> Do you plan any more changes?

No more for now, let's find out whether this optimize is useful or not.

PR #324, opened 2025-08-10
aarch64: avoid -Wunused_parameter in older CPUs

carenas:
Noticed while building recent PCRE2 in a 64-bit arm Linux, but likely to also affect others


sljit_src/sljitNativeARM_64.c
```diff
@@ -24,9 +24,15 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#ifdef __ARM_FEATURE_ATOMICS
+#define ARM_ATOMIC_INFO " (LSE)"
```
zherczeg, line 28: Is LSE the official name? Not CAS (compare and swap)?

PR #325, opened 2025-08-17
loongarch: add support for soft float ABI

carenas:


zherczeg:
@lrzlin please check this patch

lrzlin:
According to [LoongArch Toolchain Conventions](https://loongson.github.io/LoongArch-Documentation/LoongArch-toolchain-conventions-EN.html), the `__loongarch_soft_float` should only tell wether we would use fp registers for parameters passing, I don't think use it to judge LSX/LASX is exist or not is reasonable, hwcap should be enough. 

Moreover, if we really want to use it safely, we probably need disable all floating point instructions at the same time in case of the FPU doesn't exist at all.

PR #331, opened 2025-09-08
Attempt some further improvements to -Wimplicit-fallthrough support

NWilson:
* Guard use of all macros with defined(...) to avoid warnings with -Wundef
* Guard use of all attributes with appropriate __has_foo_attribute checks

---

When I tried to integrate Carlo's latest changes into PCRE2, I ran into some build failures in CI.

PCRE2 builds use the `-Wundef` warning, which prevents you doing `#if __cplusplus >= XXX`. This is unfortunate - it's not a compiler warning I'd want to use myself. But clearly some customer in the past must have requested that we support it cleanly.

I have attempted to guard against this, and ensure that all branches make use of the appropriate `__has_foo_attribute()` checks.

Finally, the C++ standard makes absolutely no mention of `__fallthrough__`: only `[[fallthrough]]` is standards-compliant.

### DIFF
--- a/sljit_src/sljitConfigInternal.h
+++ b/sljit_src/sljitConfigInternal.h
@@ -972,26 +972,27 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_sw sljit_exec_offset(void *code);
 
 #ifndef SLJIT_FALLTHROUGH
 
-#if __cplusplus >= 201703L
-#define SLJIT_FALLTHROUGH [[__fallthrough__]];
-#elif __STDC_VERSION__ >= 202311L \
-	|| (__STDC_VERSION__ > 201710L && defined(__has_c_attribute))
-#if __STDC_VERSION__ >= 202311L || __has_c_attribute(fallthrough) >= 201904L
-#define SLJIT_FALLTHROUGH [[__fallthrough__]];
+#if defined(__cplusplus) && __cplusplus >= 202002L && \
+	defined(__has_cpp_attribute)
+/* Standards-compatible C++ variant. */
+#if __has_cpp_attribute(fallthrough)
+#define SLJIT_FALLTHROUGH [[fallthrough]];
 #endif
-#elif defined(__clang__)
-#if (__clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 2)) \
-	&& __cplusplus >= 201103L
-#define SLJIT_FALLTHROUGH [[clang::fallthrough]];
-#elif defined(__has_attribute)
-#if ((__clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 6)) \
-	&& __has_attribute(fallthrough) && !defined(__cplusplus) \
-	|| (__clang_major__ >= 6 && __cplusplus == 199711L))
-#define SLJIT_FALLTHROUGH __attribute__((__fallthrough__));
+#elif !defined(__cplusplus) && \
+	  defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L && \
+	  defined(__has_c_attribute)
+/* Standards-compatible C variant. */
+#if __has_c_attribute(fallthrough)
+#define SLJIT_FALLTHROUGH [[fallthrough]];
 #endif
-#endif /* clang */
-#elif __GNUC__ >= 7 && !defined(__INTEL_COMPILER) && !defined(__NVCC__)
-#define SLJIT_FALLTHROUGH __attribute__((__fallthrough__));
+#elif ((defined(__clang__) && __clang_major__ >= 10) || \
+	   (defined(__GNUC__) && __GNUC__ >= 7)) && \
+	  defined(__has_attribute)
+/* Clang and GCC syntax. Rule out old versions because apparently Clang at
+   least has a broken implementation of __has_attribute. */
+#if __has_attribute(fallthrough)
+#define SLJIT_FALLTHROUGH __attribute__((fallthrough));
+#endif
 #endif
 
 #ifndef SLJIT_FALLTHROUGH


sljit_src/sljitConfigInternal.h
```diff
@@ -971,27 +971,41 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_sw sljit_exec_offset(void *code);
...
+#if __has_c_attribute(fallthrough)
+#define SLJIT_FALLTHROUGH [[fallthrough]];
 #endif
 #elif defined(__clang__)
-#if (__clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 2)) \
-	&& __cplusplus >= 201103L
+/* Clang-specific variants. */
+#if (__clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 2)) && \
+	defined(__cplusplus) && __cplusplus >= 201103L && \
+	defined(__has_cpp_attribute)
```
carenas, line 991: not all versions of clang have __has_cpp_attribute, but all versions that target C++11 or newer support the `[[clang::fallthrough]]` attribute which is why this last defined() was not used originally and is ill advised.

sljit_src/sljitConfigInternal.h
```diff
@@ -971,27 +971,41 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_sw sljit_exec_offset(void *code);
...
-#define SLJIT_FALLTHROUGH [[__fallthrough__]];
-#elif __STDC_VERSION__ >= 202311L \
-	|| (__STDC_VERSION__ > 201710L && defined(__has_c_attribute))
-#if __STDC_VERSION__ >= 202311L || __has_c_attribute(fallthrough) >= 201904L
-#define SLJIT_FALLTHROUGH [[__fallthrough__]];
+#if defined(__cplusplus) && __cplusplus >= 201703L && \
+	defined(__has_cpp_attribute)
+/* Standards-compatible C++ variant. */
+#if __has_cpp_attribute(fallthrough)
+#define SLJIT_FALLTHROUGH [[fallthrough]];
```
carenas, line 978: isn't `[[fallthrough]]` warranted to be there by compliance with the C++17 standard?

sljit_src/sljitConfigInternal.h
```diff
@@ -971,27 +971,41 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_sw sljit_exec_offset(void *code);
...
+	defined(__has_attribute)
+#if __has_attribute(fallthrough)
+#define SLJIT_FALLTHROUGH __attribute__((fallthrough));
+#endif
+#endif /* Clang-specific */
+#elif defined(__GNUC__) && __GNUC__ >= 7 && \
+	!defined(__INTEL_COMPILER) && !defined(__NVCC__) && \
+	defined(__has_attribute)
+/* GCC-specific variant. */
+#if __has_attribute(fallthrough)
```
carenas, line 1006: this and the previous `__has_attribute` are already warranted by the fact this is `clang`, a GNU compiler and has a version check, do you know of any reason why this might be needed?

carenas (changes requested):

sljit_src/sljitConfigInternal.h
carenas: FWIW, sljit (although an older version) is in the NetBSD kernel and the currently supported 9.x series uses gcc 7.

sljit_src/sljitConfigInternal.h
```diff
@@ -972,26 +972,27 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_sw sljit_exec_offset(void *code);
...
+#define SLJIT_FALLTHROUGH [[fallthrough]];
+#endif
+#elif !defined(__cplusplus) && \
+	  defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L && \
+	  defined(__has_c_attribute)
+/* Standards-compatible C variant. */
+#if __has_c_attribute(fallthrough)
+#define SLJIT_FALLTHROUGH [[fallthrough]];
+#endif
+#elif ((defined(__clang__) && __clang_major__ >= 9) || \
```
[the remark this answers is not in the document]
carenas: FWIW, this [branch](https://github.com/carenas/sljit/tree/pr331) has a commit on top that fixes clang 9. Note that as explained before, since clang ignores the `-Wimplicit-fallthrough` unless it has a syntax that it supports, to make sure that the code is getting exercised the right `-std` flag MUST be used.

clang <= 10 ONLY support c2x, and expect to see the `[[clang::fallthrough]]`  attribute which the new version no longer supports.

zherczeg:
I agree that this takes up so much effort. For me, the whole point of fallthrough is pointless, since case labels are just gotos. For example:

```
int f(int a, int b) {
    int i = 100;
    switch (a) {
    case 0:
        if (b > 1) {
        case 1:
            i = 50;
            while (--i > 0) {
            case 2:
                b++;
            }
        }
    }
    return b;
}
```
Where do you put the fallthrough here?

NWilson:
@carenas I've taken your commit - thank you.

I wasn't able to take the change to downgrade the C requirement from C23 to C17. Clang issues a warning if you use the `[[...]]` syntax for C17.

```txt
== CLANG c17 ==
macro-test.c:44:7: error: [[]] attributes are a C23 extension [-Werror,-Wc23-extensions]
   44 |       SLJIT_FALLTHROUGH
      |       ^
macro-test.c:14:27: note: expanded from macro 'SLJIT_FALLTHROUGH'
   14 | #define SLJIT_FALLTHROUGH [[fallthrough]];
      |                           ^
1 error generated.
```

I have also added back the exclusion for ICC and NVCC, since it's harmless, and they are current production compilers (although did you actually test on these - or was it code you added speculatively?)

---

Finally, here is the latest output of the testing script.

We are clean of warnings for all `--std=` arguments, on modern GCC and Clang.

```
== GCC c89 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== GCC c90 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== GCC c99 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== GCC c11 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== GCC c17 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== GCC c2x ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== GCC gnu89 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== GCC gnu90 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== GCC gnu99 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== GCC gnu11 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== GCC gnu17 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== GCC gnu2x ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
========================
== CLANG c89 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG c90 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG c99 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG c11 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG c17 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG c2x ==
SLJIT_FALLTHROUGH is defined as: C [[fallthrough]]
== CLANG gnu89 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG gnu90 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG gnu99 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG gnu11 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG gnu17 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG gnu2x ==
SLJIT_FALLTHROUGH is defined as: C [[fallthrough]]
========================
== G++ c++98 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== G++ c++03 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== G++ c++11 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== G++ c++17 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== G++ c++20 ==
SLJIT_FALLTHROUGH is defined as: C++ [[fallthrough]]
== G++ c++23 ==
SLJIT_FALLTHROUGH is defined as: C++ [[fallthrough]]
== G++ gnu++98 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== G++ gnu++03 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== G++ gnu++11 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== G++ gnu++17 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== G++ gnu++20 ==
SLJIT_FALLTHROUGH is defined as: C++ [[fallthrough]]
== G++ gnu++23 ==
SLJIT_FALLTHROUGH is defined as: C++ [[fallthrough]]
========================
== CLANG++ c++98 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG++ c++03 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG++ c++11 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG++ c++17 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG++ c++20 ==
SLJIT_FALLTHROUGH is defined as: C++ [[fallthrough]]
== CLANG++ c++23 ==
SLJIT_FALLTHROUGH is defined as: C++ [[fallthrough]]
== CLANG++ gnu++98 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG++ gnu++03 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG++ gnu++11 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG++ gnu++17 ==
SLJIT_FALLTHROUGH is defined as: __attribute__((fallthrough))
== CLANG++ gnu++20 ==
SLJIT_FALLTHROUGH is defined as: C++ [[fallthrough]]
== CLANG++ gnu++23 ==
SLJIT_FALLTHROUGH is defined as: C++ [[fallthrough]]
```

carenas:
> * We keep the existing `/* fallthrough */` comments, so that we guarantee not to regress on warnings for any customers using old compilers

As I said before, this is no longer needed.

[gcc >= 7](https://gcc.gnu.org/gcc-7/changes.html#c-family) uses comments or `__attribute__((fallthrough))` to mark the places were to suppress the warning.

NWilson:
> > * We keep the existing `/* fallthrough */` comments, so that we guarantee not to regress on warnings for any customers using old compilers
> 
> As I said before, this is no longer needed.
> 
> [gcc >= 7](https://gcc.gnu.org/gcc-7/changes.html#c-family) uses comments OR `__attribute__((fallthrough))` to mark the places were to suppress the warning.

Some people may use older compilers, or other compilers. Keeping the comment, which is harmless and universal, ensures that all tools or linters that used to be happy, will continue to be satisfied.

NWilson:
> > I wasn't able to take the change to downgrade the C requirement from C23 to C17. Clang issues a warning if you use the `[[...]]` syntax for C17.
> 
> I think that is why the original code was checking the return value from `__has_c_attribute()`, if it says 202311L then we are in c2x mode and there is no warning.

Well, checking `__STDC_VERSION__` is a clearer way to see if you're in C23 mode. The version number in the `__has_c_attribute()` is _less_ reliable.

> And of course, I tested icc and nvc, before adding those exceptions.

Fantastic, you do have a lot of compilers. Great!

carenas (approved):
Although the comments could be removed if we were to instead trust `__GNUC__` >= 7 and it removes support for older clang compilers, it is good enough.

carenas:
> I don't know what you were testing where you saw compilation fail without this

I never claimed that compilation would fail, only (as explained in the original commit message) that because there are so many compilers that report themselves as `__GNUC__` (for example clang, who also had a broken implementation of the `__has_attribute((fallthrough))` macro in older versions) and the syntax used (`__attribute__((fallthrough))`) is non standard, but a GNU extension, the safest thing to do was to make sure that only the derivatives we actually care about and that need this, should see that code.

Neither the Legacy Intel Compiler (icc, that identifies with `__INTEL_COMPILER`, unlike the newer one that is clang based), Nvidia's or Embarcadero's have a way to enable `-Wimplicit-fallthrough`. and at least two of them will never do, so it is safer to exempt them (bcc was exempted by the version check as it is based on clang 3.3). Nvidia is also unlikely to ever add this flag, as it is meant to pass the non CUDA code to the host compiler (which will be `cl` in Windows, and something that is not visible in Compiler Explorer).

Hope this clarifies that apparent misunderstanding, and in the same line lets make sure we are not also misunderstanding why I keep saying the second patch is not needed.

gcc 7, added support for `-Wimplicit-fallthrough` and all that is needed to suppress it was the attribute that is now in place. I noticed in one of the links you referenced earlier that someone made a mistake by saying that `__has_attribute()` might had not been supported by then, bit it was added two versions earlier and so, all versions of gcc that support the flag can use ONLY the attribute to trigger the suppression and even the current code that checks for its support will work.

any other tool that might had been looking for comments might as well see the macro, so frankly I don't think there is any need for that anymore (something that Zoltan also seems to prefer).

on that last point, squashing all commits into one (as I mentioned earlier) is another of his preferences for PR.

NWilson:
Thank you for reminding me to squash the commits. Done. (GitHub does this anyway when you merge, but I don't mind doing it.)

carenas:
One advantage of squashing them yourself is to make sure the commit message gets updatated and reflects the "why" for the whole change.

On that line, since I am hoping now it is clear no old compiler needs the comments, might be worth updating that part of the rationale for adding them back?

zherczeg (approved):
LGTM

PR #340, opened 2025-09-23
loongarch: Implement compare select (min/max) operation

lrzlin:
This implement compare select (min/max) operation for loongarch64, due to it's base instruction set lack of integer min/max ability (however, they do have fmin/fmax instruction), so use plain branch instruction instead. Tested on Loongson 3A6000.

BTW, I think the RISC-V way of implementing min/max with branch instructions (the fallback way) may have some issue, it didn't really compare `src1` with `src2_reg` IMO, though I haven't test it on a RISC-V QEMU or real machine.


zherczeg:
Thank you for trying the code on a real hardware. I have no access to riscv device unfortunately.

PR #343, opened 2025-10-04
x86: avoid SIGILL in test25 for older CPUs

carenas:
failing in at least NetBSD/i386 10.1 running in a Pentium CPU


zherczeg (approved):
LGTM

PR #344, opened 2025-10-04
netbsd: allow mprotect exception for PAX in wxalloc

carenas:
Allow running in NetBSD with the WX executable allocator.


zherczeg (approved):
LGTM

PR #345, opened 2025-10-06
config: prevent aarch64_be from being misdetected

carenas:
aarch64_be uses little endian instructions, and therefore the current generator will generate invalid instructions.

while it is possible to fix that, I am not sure how useful it would be as the architecture is fairly niche, so make sure that it will error out instead.

### DIFF
--- a/sljit_src/sljitConfig.h
+++ b/sljit_src/sljitConfig.h
@@ -133,7 +133,7 @@ extern "C" {
       nonzero value - FPU is present.
 */
 
-/* For further configurations, see the beginning of sljitConfigInternal.h */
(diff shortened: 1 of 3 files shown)


carenas:
yes, it affects at least the aarch64eb variant of NetBSD/evbarm, that can be get running with QEMU in an aarch64 host (one of those Apple CPU macs)

carenas:
correct, and my argument was that making the generator aware of that wasn't worth it, since the architecture is pretty niche. BTW, I think the same might apply to other big endian variants like (RISCV and loongarch).

what I think was needed instead, was not generating the bad instructions to begin with.

zherczeg (approved):
LGTM

PR #355, opened 2026-02-03
Fix build on non-macOS platforms such as iOS and tvOS

nirbheek:


carenas:
This will cause crashes. SLJIT doesn't build on those platforms because it can't work unless the code is changed to be able to switch patches, which is explicitly forbidden AFAIK.

PR #357, opened 2026-05-27
Fix TARGET_OS_IPHONE support

Coeur:
TARGET_OS_IPHONE support was lost around #151 due to the missing definition of SLJIT_UPDATE_WX_FLAGS


zherczeg:
@carenas what do you think about this change?

PR #359, opened 2026-06-05
Cast mask to sljit_sw in TEST_UI_IMM macro

### LINKED ISSUE #358: PPC64LE: 64-bit immediate XOR emits only lower 32 bits
uname -a: Linux c824f7490683 ... ppc64le GNU/Linux

``` c
#include <stdio.h>

#include "sljit/sljitLir.h"

typedef sljit_uw (SLJIT_FUNC* entry)(void);

int main(int argc, char* argv[]) {

  struct sljit_compiler* compiler = sljit_create_compiler(NULL);

  sljit_emit_enter(compiler, 0, SLJIT_ARG_RETURN(SLJIT_ARG_TYPE_W), 1, 0, 0);

  sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_IMM, 0x2D358DCCAA6C78A6ULL);
  sljit_emit_op2(compiler, SLJIT_XOR, SLJIT_R0, 0, SLJIT_R0, 0, SLJIT_IMM, 0x8BB84B93962EACC9ULL);
  sljit_emit_return(compiler, SLJIT_MOV, SLJIT_R0, 0);

  entry f = (entry)sljit_generate_code(compiler, 0, NULL);
  printf("%lx\n", f());

  sljit_free_code(f, NULL);
  sljit_free_compiler(compiler);

}

```

## Expected Behavior
```
a68dc65f3c42d46f
```
## Current Behavior
```
2d358dcc3c42d46f
```

jwhere:
Change the TEST_UI_IMM macro to cast the 0xffffffff mask to (sljit_sw), ensuring the mask has the same width as sljit_sw.
Fix #358

### DIFF
--- a/sljit_src/sljitNativePPC_common.c
+++ b/sljit_src/sljitNativePPC_common.c
@@ -554,7 +554,11 @@ static void reduce_code_size(struct sljit_compiler *compiler)
 
 			if (!(jump->flags & SLJIT_REWRITABLE_JUMP)) {
 				if (jump->flags & JUMP_ADDR) {
-					if (jump->u.target <= 0x01ffffff)
+					if (jump->u.target <= 0x01ffffff
+#if (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64) && (defined SLJIT_PASS_ENTRY_ADDR_TO_CALL && SLJIT_PASS_ENTRY_ADDR_TO_CALL)
+							&& !(jump->flags & IS_CALL)
+#endif
+							)
 						total_size = 1 - 1;
 #if (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64)
 					else if (jump->u.target < 0x80000000l)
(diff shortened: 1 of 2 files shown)


zherczeg:
Thank you! Could you add a test for it?

jwhere:
Done!

sljit_src/sljitNativePPC_common.c
```diff
@@ -554,7 +554,11 @@ static void reduce_code_size(struct sljit_compiler *compiler)
 
 			if (!(jump->flags & SLJIT_REWRITABLE_JUMP)) {
 				if (jump->flags & JUMP_ADDR) {
-					if (jump->u.target <= 0x01ffffff)
+					if (jump->u.target <= 0x01ffffff
+#if (defined SLJIT_CONFIG_PPC_64 && SLJIT_CONFIG_PPC_64) && (defined SLJIT_PASS_ENTRY_ADDR_TO_CALL && SLJIT_PASS_ENTRY_ADDR_TO_CALL)
+							&& !(jump->flags & IS_CALL)
```
zherczeg: Is this a different issue? What is the problems with calls?

jwhere: Yes, this is a separate issue from #358.

I found it while running the test suite on PPC64LE — test42 triggered an ASSERT, and tracing it led to reduce_code_size().

The root cause appears to be an inconsistency between reduce_code_size() and detect_jump_type(). On PPC64 with SLJIT_PASS_ENTRY_ADDR_TO_CALL, detect_jump_type() already skips optimization for IS_CALL jumps (goto keep_address), because the ELFv2 ABI requires r12 to hold the callee's entry address — something a single direct b/bl branch cannot provide. However, reduce_code_size() was missing the same guard, so it calculated that IS_CALL jumps could be reduced to one instruction. The resulting size mismatch triggered the ASSERT.

Note that this analysis was assisted by AI, so I may have missed something. If this fix looks incorrect or out of place to you, I will submit a new PR with it removed.

zherczeg: Probably valid. But it should go to a different bugfix PR with a test.

zherczeg:
Why this PR is closed?

jwhere:
> Why this PR is closed?

I apologize for the late response. I have submitted a new PR (#360)

PR #364, opened 2026-07-15
alpha: Add 64-bit JIT backend

mattst88:
Add a complete JIT code generator for DEC Alpha (AXP) 64-bit to sljit, structured as five commits:

1. Architecture detection and configuration.

2. 64-bit JIT backend: full coverage of the sljit LIR: integer ALU (32/64-bit), shifts, loads/stores, branches and calls, FP arithmetic, conditional moves, bitcount via CIX, byte/halfword ops via BWX, atomics via LDQ_L/STQ_C, and IMB for icache coherence. Runtime feature detection uses `__builtin_alpha_amask` (preferred over `getauxval` since qemu-alpha doesn't populate `AT_HWCAP`). 64-bit constants beyond 32-bit range use a four-instruction LDAH/LDA/SLL/ LDAH/LDA sequence; values requiring more than 32 bits in the high half return `SLJIT_ERR_UNSUPPORTED`. Float-to-integer conversion emits range-check fixups because Alpha CVTTQ wraps on overflow.

3. Unaligned memory synthesis: uses LDQ_U/STQ_U + EXT/INS/MSK families. FP unaligned accesses use FIX-extension ITOFS/ITOFT/FTOIS/ FTOIT to bypass the FP register file. Falls back to the kernel fixup path on CPUs without FIX. Tested with zero kernel unaligned traps on a real EV7.

4. BWX synthesis for pre-EV56 CPUs: synthesizes LDBU/LDWU/STB/STW using base-ISA LDQ_U + EXTBL/EXTWL/INSBL/INSWL/MSKBL/MSKWL + BIS + STQ_U. Taken only when ALPHA_HWCAP_BWX is absent.

5. PREFETCH: LDL to R31, the architecturally preferred prefetch form on all Alpha implementations.

Tested under qemu-alpha (EV67 emulation) and on a real HP AlphaServer ES47 EV7 (1 GHz): `sljit_test` passes all tests. PCRE2 JIT is 5.4x faster than the interpreter on the EV7.


sljit_src/sljitNativeALPHA_64.c
```diff
@@ -0,0 +1,154 @@
...
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER(S) OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR
+ * BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/* Alpha 64-bit arch dependent functions. */
```
zherczeg: If I understand correctly, Alpha is 64 bit only, so it is not worth to create such file.

sljit_src/sljitNativeALPHA_64.c
```diff
@@ -0,0 +1,154 @@
...
+		return emit_hi_lo(compiler, dst_r, TMP_ZERO, (sljit_s32)imm);
+
+	/* Full 64 bit constant: build the high half, shift and add the low half. */
+	low = (sljit_s32)imm;
+	high = imm >> 32;
+	if (low < 0)
+		high++;
+
+	if (high < S32_MIN || high > S32_MAX)
+		return SLJIT_ERR_UNSUPPORTED;
```
zherczeg: Can `load_immedate` fail?

zherczeg (commented):
This looks like a very nice contribution.
Could you extend the CI to check this cpu as well?

carenas:
> sljit_test passes all tests.

Tested it using `qemu-alpha` user emulation on Linux and full system qemu emulation (latest HEAD + small patch) with a NetBSD/alpha VM, and it fails several tests on both.

IMHO, adding tests that check the edge cases of the new implementation are also helpful

mattst88:
>  Could you extend the CI to check this cpu as well?

Yes, good idea. What I've added runs the test suite under qemu with the various ISA extensions -- which would have caught one of the problems @carenas noticed.

sljit_src/sljitNativeALPHA_64.c
mattst88: Sorry, was a bit lazy here. I've added support for generating all immediates now.

sljit_src/sljitNativeALPHA_64.c
mattst88: Oh, yes. Thanks, cleaned up.

sljit_src/sljitNativeALPHA_64.c
```diff
@@ -0,0 +1,3658 @@
...
+	if (imm >= S32_MIN && imm <= S32_MAX)
+		return emit_hi_lo(compiler, dst_r, TMP_ZERO, (sljit_s32)imm);
+
+	/* Full 64 bit constant: build the high half, shift and add the low half.
+	   The low half is added back sign extended, so the high half is biased by
+	   one when it is negative. That bias can push high up to 0x80000000, which
+	   no longer fits in a signed 32 bit value, but only the low 32 bits of the
+	   high half survive the shift, so computing it modulo 2^32 is equivalent. */
+	low = (sljit_s32)imm;
+	high = imm >> 32;
```
carenas: to avoid UB, need to cast `imm` to unsigned before shifting

sljit_src/sljitNativeALPHA_64.c
```diff
@@ -0,0 +1,3658 @@
...
+
+	low = (sljit_s32)addr;
+	high = (sljit_sw)addr >> 32;
+	if (low < 0)
+		high++;
+
+	inst[0] = LDAH | RA(reg) | RB(TMP_ZERO) | ((((sljit_ins)(high >> 16)) + (((high) & 0x8000) ? 1 : 0)) & 0xffff);
+	inst[1] = LDA | RA(reg) | RB(reg) | ((sljit_ins)high & 0xffff);
+	inst[2] = SLL | RA(reg) | ALPHA_LIT(32) | RC(reg);
+	inst[3] = LDAH | RA(reg) | RB(reg) | ((((sljit_ins)(low >> 16)) + (((low) & 0x8000) ? 1 : 0)) & 0xffff);
```
carenas: `(sljit_u32)low >> 16` and similarly for `ALPHA_HI16`

sljit_src/sljitNativeALPHA_64.c
```diff
@@ -0,0 +1,3658 @@
...
+}
+
+static sljit_s32 getput_arg(struct sljit_compiler *compiler, sljit_s32 flags, sljit_s32 reg, sljit_s32 arg, sljit_sw argw, sljit_s32 next_arg, sljit_sw next_argw)
+{
+	sljit_s32 base = arg & REG_MASK;
+	sljit_s32 tmp_r = (flags & MEM_USE_TMP2) ? TMP_REG2 : TMP_REG1;
+	sljit_sw offset;
+
+	SLJIT_UNUSED_ARG(next_arg);
+	SLJIT_UNUSED_ARG(next_argw);
```
carenas: wouldn't make more sense, to NOT pass those 2 parameters instead?

sljit_src/sljitNativeALPHA_64.c
```diff
@@ -0,0 +1,3658 @@
...
+
+	return const_;
+}
+
+SLJIT_API_FUNC_ATTRIBUTE struct sljit_jump* sljit_emit_op_addr(struct sljit_compiler *compiler, sljit_s32 op,
+	sljit_s32 dst, sljit_sw dstw)
+{
+	struct sljit_jump *jump;
+	sljit_s32 dst_r, target_r;
+	SLJIT_UNUSED_ARG(op);
```
carenas: delete

carenas (changes requested):
all tests pass in the emulated NetBSD/Alpha

sljit_src/sljitNativeALPHA_64.c
mattst88: Oops, done. Thanks!

sljit_src/sljitNativeALPHA_64.c
mattst88: Thanks, done.

sljit_src/sljitNativeALPHA_64.c
mattst88: Yeah, definitely. Good call. Done.

sljit_src/sljitNativeALPHA_64.c
mattst88: Thanks, done.

zherczeg:
The CI complains

mattst88:
Should be fixed now.

zherczeg (approved):
LGTM

zherczeg:
The patch is landed. @mattst88 you will be the maintainer of this port

mattst88:
> The patch is landed. @mattst88 you will be the maintainer of this port

Sounds great. Thank you very much!

PR #365, opened 2026-07-24
ubsan: avoid UB with << of a signed integer

carenas:
While running `sljit_test` with the undefined sanitizer enabled, noticed a failure inside `sljit_emit_op2_shift()` that was
found in most implementations.

The impact is mostly theoretical, as most compilers would normally do a logical shift left as was intended, and because
the test affected wasn't a realistic use case, with an index so close to MAX_LONG.


zherczeg (approved):
LGTM

PR #367, opened 2026-07-28
ppc: add SLJIT_BREAKPOINT

carenas:
helpful when debugging the generated code in PowerPC.

tested in POWER9 (little endian) and POWER10 (big endian, 32/64bit)

while at it, update documentation to remove refence to ppc as an architecture that is missing implementation.


sljit_src/sljitNativePPC_common.c
```diff
@@ -1422,6 +1422,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_op0(struct sljit_compiler *compile
 	op = GET_OPCODE(op);
 	switch (op) {
 	case SLJIT_BREAKPOINT:
+		return push_inst(compiler, 0x7fe00008);
```
zherczeg, line 1425: Do we need to add an ifdef to old ppc?

zherczeg (commented):
I have been thinking for some time that ppc is outdated, and we should move to power, which is still maintained.

sljit_src/sljitNativePPC_common.c
carenas: If you know what version that comment was referring to, AFAIK all powerpc CPUs had the `t/tw/trap` instruction[1]

[1] https://www.nxp.com/docs/en/application-note/AN2491.pdf

zherczeg: As far as I remember they used invalid instructions for trap similar to s390x

carenas: The spec for the first PowerPC CPU (601) has them defined with the HI/LO I used:

https://www.ibm.com/docs/en/aix/7.3.0?topic=reference-appendix-g-powerpc-601-risc-microprocessor-instructions

zherczeg (approved):
LGTM

PR #368, opened 2026-08-11
Fix build warnings by avoid function pointer to object pointer casts

NWilson:
Casting between function pointers and object (struct/int) pointers is not allowed under ANSI pedantic rules.

However, casting between pointers and integers (uintptr_t) is allowed, so we can fix these PPC warnings by doing that.


zherczeg (approved):
LGTM

Issue #204, opened 2023-09-15
Drop old MIPS support


zherczeg:
Is it worth to support MIPS III? @carenas what do you think?

carenas:
If the objective is to clean up the code, then there is little benefit by dropping MIPS III and would need instead to drop everything before `r2` which is IMHO too drastic considering r6 (likely the last release and that is itself 5 years old) doesn't even compile.

mattst88:
FWIW, I checked the status of the MIPS code in sljit and pcre2, and everything looks good:

Tested MIPS support across all four endian/width combinations using cross-compilation and qemu user-mode emulation.

**Toolchain notes:**
- MIPS64: `mips64{,el}-unknown-linux-gnu-gcc -mabi=64` (GCC 15.2)
- MIPS32: `mips64{,el}-unknown-linux-gnu-gcc -mabi=32 -march=mips32r2` (the `mips32-unknown-linux-gnu` Gentoo toolchains default to n32 ABI with no o32 sysroot, so the mips64 compiler is used instead)
- qemu 10.2.2 (`qemu-mips{,el,64,64el}`)

**sljit test suite:**

| Target | Result |
|---|---|
| MIPS III 64bit big endian | all tests PASSED |
| MIPS III 64bit little endian | all tests PASSED |
| MIPS32-R2 32bit big endian | all tests PASSED |
| MIPS32-R2 32bit little endian | all tests PASSED |

**pcre2 `RunTest` (with JIT, 8/16/32-bit modes):**

| Target | Result |
|---|---|
| MIPS64 big endian | All tests passed |
| MIPS64 little endian | All tests passed |
| MIPS32 big endian | All tests passed |
| MIPS32 little endian | All tests passed |

zherczeg:
Unfortunately MIPS III (and even newer ones) has some delayed result computation (i.e. the next instruction cannot use its result) for some instructions. Those instructions cannot be tested on qemu.

mattst88:
I don't think that's the case, practically speaking. MIPS I (R2000 and R3000) from the mid-to-late 80's certainly don't have hardware interlocks, but MIPS III definitely does. See [MIPS® Architecture For Programmers Volume III](https://s3-eu-west-1.amazonaws.com/downloads-mips/documents/MD00091-2B-MIPS64PRA-AFP-05.04.pdf). The word `interlock` only appears three times, all in the description of CP0 hazards that userspace software doesn't need to care about.

[...]

PR #369, opened 2026-08-11
Fix build & test failures on MIPS R6

NWilson:
The code did not even compile, so it's certain to be untested.
    
  - Fix some op encoding issues
  - Provide a new emit_prefetch for R6
  - Provide the missing sljit_emit_fmem on R6, which prevented linking and running the tests.

---------

I didn't write much of the code in this PR! I just let GPT run stuff with qemu and fix things on its own.

It _claims_ to know all the MIPS opcodes, and I also asked it to use gnu-as to write out binary data and disassemble it for each MIPS subarchitecture to verify its knowledge.

The tests didn't pass before; now, they do. I simply used a combination of your own `sljit_test -v` tests, and the PCRE2 ones as well.

### DIFF
--- a/sljit_src/sljitConfigInternal.h
+++ b/sljit_src/sljitConfigInternal.h
@@ -147,7 +147,7 @@ extern "C" {
 #endif
 
 /***********************************************************/
-/* Intel Control-flow Enforcement Technology (CET) spport. */
+/* Intel Control-flow Enforcement Technology (CET) support.*/
 /***********************************************************/
 
 #ifdef SLJIT_CONFIG_X86
@@ -423,9 +423,15 @@ typedef double sljit_f64;
 #define SLJIT_CONV_MIN_FLOAT SLJIT_CONV_RESULT_MIN_INT
 #define SLJIT_CONV_NAN_FLOAT SLJIT_CONV_RESULT_ZERO
 #elif (defined SLJIT_CONFIG_MIPS && SLJIT_CONFIG_MIPS)
+#if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
+#define SLJIT_CONV_MAX_FLOAT SLJIT_CONV_RESULT_MAX_INT
+#define SLJIT_CONV_MIN_FLOAT SLJIT_CONV_RESULT_MIN_INT
+#define SLJIT_CONV_NAN_FLOAT SLJIT_CONV_RESULT_ZERO
+#else /* SLJIT_MIPS_REV < 6 */
 #define SLJIT_CONV_MAX_FLOAT SLJIT_CONV_RESULT_MAX_INT
 #define SLJIT_CONV_MIN_FLOAT SLJIT_CONV_RESULT_MAX_INT
 #define SLJIT_CONV_NAN_FLOAT SLJIT_CONV_RESULT_MAX_INT
+#endif /* SLJIT_MIPS_REV >= 6 */
 #elif (defined SLJIT_CONFIG_PPC && SLJIT_CONFIG_PPC)
 #define SLJIT_CONV_MAX_FLOAT SLJIT_CONV_RESULT_MAX_INT
 #define SLJIT_CONV_MIN_FLOAT SLJIT_CONV_RESULT_MIN_INT
--- a/sljit_src/sljitNativeMIPS_common.c
+++ b/sljit_src/sljitNativeMIPS_common.c
@@ -249,8 +249,13 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define LDL		(HI(26))
 #define LDR		(HI(27))
 #define LDC1		(HI(53))
+#if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
+#define LL		(HI(31) | LO(54))
+#define LLD		(HI(31) | LO(55))
+#else /* SLJIT_MIPS_REV < 6 */
 #define LL		(HI(48))
 #define LLD		(HI(52))
+#endif /* SLJIT_MIPS_REV >= 6 */
 #define LUI		(HI(15))
 #define LW		(HI(35))
 #define LWL		(HI(34))
@@ -286,8 +291,13 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define ROTR		(HI(0) | (1 << 21) | LO(2))
 #define ROTRV		(HI(0) | (1 << 6) | LO(6))
 #endif /* SLJIT_MIPS_REV >= 2 */
+#if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
+#define SC		(HI(31) | LO(38))
+#define SCD		(HI(31) | LO(39))
+#else /* SLJIT_MIPS_REV < 6 */
 #define SC		(HI(56))
 #define SCD		(HI(60))
+#endif /* SLJIT_MIPS_REV >= 6 */
 #define SD		(HI(63))
 #define SDL		(HI(44))
 #define SDR		(HI(45))
@@ -317,10 +327,11 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define XORI		(HI(14))
 
 #if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 1)
-#define CLZ		(HI(28) | LO(32))
 #if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
-#define DCLZ		(LO(18))
+#define CLZ		((1 << 6) | LO(16))
+#define DCLZ		((1 << 6) | LO(18))
 #else /* SLJIT_MIPS_REV < 6 */
+#define CLZ		(HI(28) | LO(32))
 #define DCLZ		(HI(28) | LO(36))
 #define MOVF		(HI(0) | (0 << 16) | LO(1))
 #define MOVF_S		(HI(17) | FMT_S | (0 << 16) | LO(17))
@@ -332,8 +343,13 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define MOVZ_S		(HI(17) | FMT_S | LO(18))
 #define MUL		(HI(28) | LO(2))
 #endif /* SLJIT_MIPS_REV >= 6 */
+#if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
+#define PREF		(HI(31) | LO(53))
+#define PREF_OFFSET(offset)	(((sljit_ins)(offset) & 0x1ff) << 7)
+#else /* SLJIT_MIPS_REV < 6 */
 #define PREF		(HI(51))
 #define PREFX		(HI(19) | LO(15))
+#endif /* SLJIT_MIPS_REV >= 6 */
 #if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 2
 #define SEB		(HI(31) | (16 << 6) | LO(32))
 #define SEH		(HI(31) | (24 << 6) | LO(32))
@@ -2118,8 +2134,8 @@ static SLJIT_INLINE sljit_s32 emit_single_op(struct sljit_compiler *compiler, sl
 		}
 
 #if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
-		FAIL_IF(push_inst(compiler, SELECT_OP(DMUL, MUL) | S(src1) | T(src2) | D(dst), DR(dst)));
 		FAIL_IF(push_inst(compiler, SELECT_OP(DMUH, MUH) | S(src1) | T(src2) | DA(EQUAL_FLAG), EQUAL_FLAG));
+		FAIL_IF(push_inst(compiler, SELECT_OP(DMUL, MUL) | S(src1) | T(src2) | D(dst), DR(dst)));
 #else /* SLJIT_MIPS_REV < 6 */
 		FAIL_IF(push_inst(compiler, SELECT_OP(DMULT, MULT) | S(src1) | T(src2), MOVABLE_INS));
 		FAIL_IF(push_inst(compiler, MFHI | DA(EQUAL_FLAG), EQUAL_FLAG));
@@ -2516,6 +2532,37 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_op0(struct sljit_compiler *compile
 static sljit_s32 emit_prefetch(struct sljit_compiler *compiler,
         sljit_s32 src, sljit_sw srcw)
 {
+#if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
+	sljit_s32 base = src & REG_MASK;
+	sljit_s32 offset_reg;
+
+	if (!(src & OFFS_REG_MASK)) {
+		if (srcw <= 0xff && srcw >= -0x100)
+			return push_inst(compiler, PREF | S(base) | PREF_OFFSET(srcw), MOVABLE_INS);
+
+		FAIL_IF(load_immediate(compiler, DR(TMP_REG1), srcw));
+
+		if (base != 0)
+			FAIL_IF(push_inst(compiler, ADDU_W | S(base) | T(TMP_REG1) | D(TMP_REG1), DR(TMP_REG1)));
+
+		return push_inst(compiler, PREF | S(TMP_REG1), MOVABLE_INS);
+	}
+
+	srcw &= 0x3;
+	offset_reg = OFFS_REG(src);
+
+	if (SLJIT_UNLIKELY(srcw != 0)) {
+		FAIL_IF(push_inst(compiler, SLL_W | T(offset_reg) | D(TMP_REG1) | SH_IMM(srcw), DR(TMP_REG1)));
+		offset_reg = TMP_REG1;
+	}
+
+	if (base != 0) {
+		FAIL_IF(push_inst(compiler, ADDU_W | S(base) | T(offset_reg) | D(TMP_REG1), DR(TMP_REG1)));
+		offset_reg = TMP_REG1;
+	}
+
+	return push_inst(compiler, PREF | S(offset_reg), MOVABLE_INS);
+#else /* SLJIT_MIPS_REV < 6 */
 	if (!(src & OFFS_REG_MASK)) {
 		if (srcw <= SIMM_MAX && srcw >= SIMM_MIN)
 			return push_inst(compiler, PREF | S(src & REG_MASK) | IMM(srcw), MOVABLE_INS);
@@ -2532,6 +2579,7 @@ static sljit_s32 emit_prefetch(struct sljit_compiler *compiler,
 	}
 
 	return push_inst(compiler, PREFX | S(src & REG_MASK) | T(OFFS_REG(src)), MOVABLE_INS);
+#endif /* SLJIT_MIPS_REV >= 6 */
 }
 #endif /* SLJIT_MIPS_REV >= 1 */
 
@@ -4270,8 +4318,6 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_mem(struct sljit_compiler *compile
 #endif /* SLJIT_MIPS_REV >= 6 */
 }
 
-#if !(defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
-
 SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fmem(struct sljit_compiler *compiler, sljit_s32 type,
 	sljit_s32 freg,
 	sljit_s32 mem, sljit_sw memw)
@@ -4279,6 +4325,9 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fmem(struct sljit_compiler *compil
 	CHECK_ERROR();
 	CHECK(check_sljit_emit_fmem(compiler, type, freg, mem, memw));
 
+#if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
+	return sljit_emit_fmem_unaligned(compiler, type, freg, mem, memw);
+#else /* !(SLJIT_MIPS_REV >= 6) */
 	FAIL_IF(update_mem_addr(compiler, &mem, &memw, SIMM_MAX - ((type & SLJIT_32) ? 3 : 7)));
 	SLJIT_ASSERT(FAST_IS_REG(mem) && mem != TMP_REG2);
 
@@ -4358,9 +4407,8 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fmem(struct sljit_compiler *compil
 	FAIL_IF(push_inst(compiler, NOP, UNMOVABLE_INS));
 #endif /* MIPS III */
 	return SLJIT_SUCCESS;
+#endif /* SLJIT_MIPS_REV >= 6 */
 }
-
-#endif /* !SLJIT_MIPS_REV || SLJIT_MIPS_REV < 6 */
 
 #undef IMM_16_SECOND
 #undef IMM_16_FIRST


sljit_src/sljitConfigInternal.h
```diff
@@ -423,9 +430,15 @@ typedef double sljit_f64;
 #define SLJIT_CONV_MIN_FLOAT SLJIT_CONV_RESULT_MIN_INT
 #define SLJIT_CONV_NAN_FLOAT SLJIT_CONV_RESULT_ZERO
 #elif (defined SLJIT_CONFIG_MIPS && SLJIT_CONFIG_MIPS)
+#if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
+#define SLJIT_CONV_MAX_FLOAT SLJIT_CONV_RESULT_MAX_INT
+#define SLJIT_CONV_MIN_FLOAT SLJIT_CONV_RESULT_MIN_INT
+#define SLJIT_CONV_NAN_FLOAT SLJIT_CONV_RESULT_ZERO
```
NWilson, line 429: GPT believes that the saturation behaviour was changed in MIPS R6, and this single change here converts a test failure into a pass.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -249,8 +249,13 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define LDL		(HI(26))
 #define LDR		(HI(27))
 #define LDC1		(HI(53))
+#if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
+#define LL		(HI(31) | LO(54))
+#define LLD		(HI(31) | LO(55))
```
NWilson, line 254: GPT claims that several instruction encodings were changed in R6.

Before making these changes, the tests actually crashed with SIGILL (illegal instruction). The gnu-as disassembly with these changes confirms these new constants really are emitting LL/LLD correctly (as far as I can tell - I know **nothing** about MIPS)

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -336,8 +347,13 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define MOVZ_S		(HI(17) | FMT_S | LO(18))
 #define MUL		(HI(28) | LO(2))
 #endif /* SLJIT_MIPS_REV >= 6 */
+#if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
+#define PREF		(HI(31) | LO(53))
+#define PREF_OFFSET(offset)	(((sljit_ins)(offset) & 0x1ff) << 7)
```
NWilson, line 352: Apparently PREFX doesn't exist in R6. GPT wrote some entirely new code to perform its functionality.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -4129,10 +4034,10 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_mem(struct sljit_compiler *compile
 	sljit_s32 reg,
 	sljit_s32 mem, sljit_sw memw)
 {
-	sljit_s32 op = type & 0xff;
-	sljit_s32 flags = 0;
 	sljit_ins ins;
 #if !(defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
+	sljit_s32 op = type & 0xff;
+	sljit_s32 flags = 0;
```
NWilson, line 4183: I did this. It's just silencing a compiler warning.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -4278,16 +4183,17 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_mem(struct sljit_compiler *compile
...
 SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_fmem(struct sljit_compiler *compiler, sljit_s32 type,
 	sljit_s32 freg,
 	sljit_s32 mem, sljit_sw memw)
 {
 	CHECK_ERROR();
 	CHECK(check_sljit_emit_fmem(compiler, type, freg, mem, memw));
 
-	FAIL_IF(update_mem_addr(compiler, &mem, &memw, SIMM_MAX - (type & SLJIT_32) ? 3 : 7));
+#if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
+	return sljit_emit_fmem_unaligned(compiler, type, freg, mem, memw);
```
NWilson, line 4337: Without this change, the tests don't even link on MIPS R6, because sljit_emit_fmem needs to be defined.

I have followed GPT's advice and made it fallback to the generic unaligned version.

sljit_src/sljitConfigInternal.h
```diff
@@ -423,9 +423,15 @@ typedef double sljit_f64;
...
```
mattst88, line 429: Confirmed against the architecture manuals, so this is right on both sides.

Table 6.3 in Vol I-A (MD00083 rev 6.01, "Value Supplied When a New Quiet NaN Is Created") gives the fixed-point defaults directly. With `FCSR.NAN2008=1`: too-big -> 0x7fffffff, NaN -> 0x00000000, too-small -> 0x80000000, i.e. exactly MAX_INT / ZERO / MIN_INT. With `NAN2008=0` all three are 0x7fffffff, which is the existing pre-R6 branch.

And the compile-time split is legitimate because Release 6 hardwires the 2008 bits read-only to 1 (Vol I-A section 2.3.6: "IEEE 2008: NaN behavior is required. ... FCSR.Has2008=FIR.ABS2008=FIR.NAN2008=1 are hardwired read-only"), so on R6 there is no runtime FCSR state that could select the other behavior.

Worth putting those two references in the commit message.

Aside, not something this PR needs to fix: a pre-R6 core with Has2008 and `FCSR.NAN2008=1` gets the 2008 results too, and the `#else` branch would then be wrong. Not detectable at compile time, and it predates this change.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -321,10 +331,11 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define XORI		(HI(14))
 
 #if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 1)
-#define CLZ		(HI(28) | LO(32))
 #if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
-#define DCLZ		(LO(18))
+#define CLZ		((1 << 6) | LO(16))
```
mattst88, line 335: Worth noting in the commit message why the call site at line 1854 drops `T(dst)`: the R6 encoding fixes rt to 00000 (SPECIAL | rs | 00000 | rd | 00001 | 010000, MD00087 rev 6.06), so this isn't just tidying -- the old form was encoding a reserved instruction.

The old `#define DCLZ (LO(18))` was SPECIAL funct 18, i.e. `mflo`, which is literally the instruction `sljit_test` died on before this patch.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -871,7 +887,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 		if (!cpu_feature_list)
 			get_cpu_features();
 
-		return (cpu_feature_list & CPU_FEATURE_FR) != 0;
+		return (cpu_feature_list & CPU_FEATURE_FR) == 0;
```
mattst88, line 890: The inversion is correct -- `CPU_FEATURE_FR` means FR=1 (the `MTHC1` path in sljitNativeMIPS_32.c), and `F64_AS_F32_PAIR` is the FR=0 even/odd pairing mode. I confirmed it: reverting only this hunk on top of the rest of the PR makes `test70 case 3` fail on MIPS32-R6, and MIPS32-R2 (FR=0) still passes with it.

One problem though: `get_cpu_features()` only sets `CPU_FEATURE_FR` inside `if (FPU && FP64)`, so on a MIPS32 with no FPU at runtime there is no FR bit and this now returns 1. The enclosing `#if` is compile-time only, so it doesn't cover that case. Something like:

```c
	return (cpu_feature_list & (CPU_FEATURE_FPU | CPU_FEATURE_FR)) == CPU_FEATURE_FPU;
```

mattst88 (commented):
I tested this under QEMU and checked the encodings against the architecture manuals. Short version: the R6 fixes are correct, with one behavioral concern noted inline.

I verified every new opcode macro two ways: assembled with `gas -march=mips64r6` and compared bit-for-bit, and checked against the encoding diagrams in MD00087 rev 6.06. LL/LLD/SC/SCD and PREF are SPECIAL3 with the 9-bit offset at bits 15-7 (funct 110110/110111/100110/100111/110101), and CLZ/DCLZ are SPECIAL with rt=0 and sa=00001 (funct 010000/010010). All match. The LL/SC call sites emit offset 0, so the narrower R6 offset field is fine.

The MUH-before-MUL reordering is right: the old order clobbered `dst` before MUH read its sources when `dst` aliased `src1`/`src2`.

Test results with the patch applied (`sljit_test` and `regex_test`, warning-clean at the default flags):

| target | sljit_test | regex_test |
| --- | --- | --- |
| MIPS64-R6 LE (qemu `-cpu I6400`) | PASS | PASS |
| MIPS64-R6 BE (qemu `-cpu I6400`) | PASS | PASS |
| MIPS32-R6 LE (qemu `-cpu mips32r6-generic`) | PASS | PASS |
| MIPS64-R2 LE | PASS | PASS |
| MIPS32-R2 LE | PASS | - |
| LOONGARCH64 | PASS | - |

There is no R6 glibc for the toolchains I have, so the R6 runs use a musl I built for mips64r6el/mips64r6/mips32r6el plus R6-compiled soft-fp routines; stock r2 libgcc is unusable on an R6 core because it emits `jr`, which R6 removed. Happy to write that setup up if it's useful to anyone else.

For reference, on master the R6 build fails outright with an undefined reference to `sljit_emit_fmem`; stubbing that out to get a link, `sljit_test` then dies with SIGILL on the `mflo` mentioned in the CLZ comment below.

Two review notes beyond the inline comments:

- The LOONGARCH prefetch fix is a real bug -- the old code fed an uninitialized `TMP_REG1` to `ADD_D` when `OFFS_REG` was present with `srcw == 0`. Worth pointing out that the test suite cannot catch it, since a prefetch to a garbage address is architecturally harmless; it passes either way. That hunk stands on inspection alone.
- The `update_mem_addr` parenthesization fix is likewise real (the old `SIMM_MAX - (type & SLJIT_32) ? 3 : 7` always evaluated to 3), but it affects non-R6 MIPS and has nothing to do with the rest.

Which brings me to the main request: could you split this up? Right now it's a CET comment typo, a LOONGARCH codegen bug, a non-R6 MIPS precedence bug, and the R6 work in one commit, under a title that says MIPS. Separate commits would make each one bisectable, which matters more than usual here given how the patch was generated.

Minor nit: `#define CLZ ((1 << 6) | LO(16))` drops the `HI(0)` that the neighboring SPECIAL-encoded macros (`SEB`, `SEH`) spell out.

NWilson:
Great feedback, thanks! I can split up the PR. It's not much admin to do.

And thanks for looking it all up in the manuals. Are there other discontinued CPUs you specialise in? ;)

I guess we should also ask GPT to review the big-hitters like ARM and x64 as well, in case there are small things hiding in there.

mattst88:
> And thanks for looking it all up in the manuals. Are there other discontinued CPUs you specialise in? ;)

I can probably help with PowerPC :)

sljit_src/sljitConfigInternal.h
zherczeg: So they follow ARM in R6. Btw, is there a real R6 CPU in the market?

sljit_src/sljitNativeLOONGARCH_64.c
```diff
@@ -2249,9 +2249,13 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_op_src(struct sljit_compiler *comp
...
 			srcw &= 0x3;
-			if (SLJIT_UNLIKELY(srcw))
+			if (SLJIT_UNLIKELY(srcw)) {
 				FAIL_IF(push_inst(compiler, SLLI_D | RD(TMP_REG1) | RJ(OFFS_REG(src)) | IMM_I12(srcw)));
-			FAIL_IF(push_inst(compiler, ADD_D | RD(TMP_REG1) | RJ(base) | RK(TMP_REG1)));
+				FAIL_IF(push_inst(compiler, ADD_D | RD(TMP_REG1) | RJ(base) | RK(TMP_REG1)));
+			} else if (base != 0)
+				FAIL_IF(push_inst(compiler, ADD_D | RD(TMP_REG1) | RJ(base) | RK(OFFS_REG(src))));
+			else
+				return push_inst(compiler, PRELD | RD(0) | RJ(OFFS_REG(src)));
```
zherczeg, line 2258: @Xiao-Tao could you check these changes? I would put the return part first.

sljit_src/sljitNativeMIPS_common.c
zherczeg: These are atomic instructions. Probably the code needs to be changed as well.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -2124,8 +2140,8 @@ static SLJIT_INLINE sljit_s32 emit_single_op(struct sljit_compiler *compiler, sl
 		}
 
 #if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
-		FAIL_IF(push_inst(compiler, SELECT_OP(DMUL, MUL) | S(src1) | T(src2) | D(dst), DR(dst)));
 		FAIL_IF(push_inst(compiler, SELECT_OP(DMUH, MUH) | S(src1) | T(src2) | DA(EQUAL_FLAG), EQUAL_FLAG));
+		FAIL_IF(push_inst(compiler, SELECT_OP(DMUL, MUL) | S(src1) | T(src2) | D(dst), DR(dst)));
```
zherczeg, line 2144: Why is this change?

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -2522,6 +2538,37 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_emit_op0(struct sljit_compiler *compile
 static sljit_s32 emit_prefetch(struct sljit_compiler *compiler,
         sljit_s32 src, sljit_sw srcw)
 {
+#if (defined SLJIT_MIPS_REV && SLJIT_MIPS_REV >= 6)
```
zherczeg, line 2541: The other code path is created because the existence of `PREFX`. Without it, this is a normal memory instruction. Could it go through the normal addressing code?

sljit_src/sljitNativeMIPS_common.c
zherczeg: The `update_mem_addr` has side effects. Is this work without them?

zherczeg (changes requested):

sljit_src/sljitConfigInternal.h
mattst88: It seems like it was used in some embedded products, and I see that MIPS developers had evaluation boards from Linux kernel commit messages and mailing list threads, but otherwise MIPS R6 indeed seems to not exist in a way you or I could use.

sljit_src/sljitNativeMIPS_common.c
mattst88: I checked the encodings and both call sites. The constants are right, and I think the code is fine as it stands, though it could use a comment saying why.

R6 moved LL/SC out of their own major opcodes into SPECIAL3 (op = 0b011111) with a 9-bit signed offset in bits 15..7, bit 6 reserved as 0, and the function field in bits 5..0 (MD00087 rev 6.06). The new funct values are LL = 0x36 (54), LLD = 0x37 (55), SC = 0x26 (38), SCD = 0x27 (39), matching what the patch defines.

Nothing else has to change because the offset is always zero here. `sljit_emit_atomic_load()` emits `ins | T(dst_reg) | S(mem_reg)` and `sljit_emit_atomic_store()` emits `ins | TA(OTHER_FLAG) | S(mem_reg)`; neither uses `IMM()`, and the sljit atomic API only takes a register address, so there is no displacement to encode. A zero offset encodes identically in the pre-R6 16-bit form and the R6 9-bit form.

That does leave a trap for later: if anyone adds a displacement, `IMM()` would write into bits 15..0 and silently corrupt the funct field on R6. Either a short comment at the defines or an `LL_OFFSET()` macro alongside the new `PREF_OFFSET()` would spell the R6 field layout out rather than leaving it implied.

Separately, R6 does tighten the rules for what may appear between the LL and the SC (the forbidden-instruction list around the LL/SC loop), but that is a property of what the sljit *caller* emits between `sljit_emit_atomic_load()` and `sljit_emit_atomic_store()`, not of this backend, and it is unchanged by this patch.

sljit_src/sljitNativeMIPS_common.c
mattst88: It fixes a real bug, so it probably wants its own commit.

In the overflow path `dst` really can be the same register as `src1` or `src2`. `emit_op()` assigns `dst_r = dst` for `FAST_IS_REG(dst)`, `src1_r = src1` for `FAST_IS_REG(src1)` and `src2_r = src2` for `FAST_IS_REG(src2)`; the only place it substitutes a temporary is the `MOVE_OP` path, which `SLJIT_MUL` does not take. So `sljit_emit_op2(compiler, SLJIT_MUL | SLJIT_SET_OVERFLOW, R1, 0, R0, 0, R1, 0)` reaches `emit_single_op()` with `dst == src2`.

With `MUL` emitted first, `dst` is written before `MUH` reads `S(src1)` / `T(src2)`, so `MUH` computes the high word from the low half of the product instead of from the original operands, and `EQUAL_FLAG` (and therefore the overflow result) is garbage. Emitting `MUH` first reads both operands while they are still intact, and `MUL` writing `dst` afterwards is harmless because `MUH` has already consumed them.

The existing suite already covers this: `test56` has both `MUL | SLJIT_SET_OVERFLOW` with `dst == src2` (`R1 = R0 * R1`) and with `dst == src1 == src2` (`R1 = R1 * R1`), which is presumably one of the failures this PR turns into a pass.

The pre-R6 path does not have the problem because `MULT` writes HI/LO before either `MFHI` or `MFLO` touches `dst`, which is why this only shows up on R6.

sljit_src/sljitNativeMIPS_common.c
mattst88: Not quite, unfortunately. R6 removed PREFX (it was a COP1X instruction, and COP1X is gone), but it also moved PREF itself: it is now SPECIAL3 (op = 0b011111), funct 0x35, with a 9-bit signed offset in bits 15..7, rather than the old 16-bit-offset I-type form. So the base+reg form is gone, and the base+offset form no longer has the same immediate range or field position as an ordinary load/store.

The generic helpers therefore cannot be reused as-is: `IMM()` and everything built on it assume `SIMM_MIN`..`SIMM_MAX`.

The closest thing to what you are asking for would be `update_mem_addr()`, which already folds `OFFS_REG` plus shift into `TMP_REG1` and handles out-of-range offsets, and would collapse the whole R6 branch to roughly:

```c
	FAIL_IF(update_mem_addr(compiler, &src, &srcw, 0xff));
	return push_inst(compiler, PREF | S(src) | PREF_OFFSET(srcw), MOVABLE_INS);
```

The blocker is that `update_mem_addr()` only bounds the offset from above (`argw <= max_offset && argw >= SIMM_MIN`); it would leave `srcw == -1000` in place, which does not fit the 9-bit field. It needs a lower bound too, either a second parameter or a `min_offset` derived from `max_offset`. If you would take that change the dedicated R6 path here almost disappears, and `sljit_emit_fmem()` and friends get a tighter check as a side effect. Otherwise some dedicated path is unavoidable, and the one in the PR looks correct to me, just larger than it needs to be.

sljit_src/sljitNativeMIPS_common.c
mattst88: Yes, nothing is lost. `update_mem_addr()` exists to reduce a general `mem`/`memw` down to a base register plus a small offset so that the LDL/LDR and SDL/SDR pairs can be emitted at `memw` and `memw + 3` / `memw + 7`. The R6 path never emits those pairs; it calls `sljit_emit_fmem_unaligned()`, which just forwards the original `mem`/`memw` to `sljit_emit_fop1()`, and `fop1` runs the normal addressing path (`getput_arg()`) itself. So the address computation still happens, just one level down instead of here. Since the R6 branch returns immediately, there is also no temp-register state left over that the rest of the function would have depended on.

I also checked the premise, since the fallback is only valid if R6 guarantees misaligned support for the COP1 load/store forms and not only the integer ones. It does, explicitly. MD00083 rev 6.01 (Vol I-A), section 7.2.1.1 "Data Alignment in Loads, Stores, and Moves":

> pre-Release 6: All coprocessor loads and stores operate on naturally-aligned data items. An attempt to load or store to an address that is not naturally aligned for the data item causes an Address Error exception.
>
> Release 6 requires that a Release 6 compliant system support misaligned memory accesses (possibly by trapping and emulating) for all ordinary memory access instructions, including coprocessor loads and stores. Nevertheless, it is recommended that software naturally align FPU data in memory, for improved performance.

and Appendix B, "Misaligned Memory Accesses":

> Release 6 requires misaligned memory access support for all ordinary memory access instructions (for example, LW/SW, LWC1/SWC1). However, misaligned support is not provided for certain special memory accesses such as atomics (for example, LL/SC).

Note the "possibly by trapping and emulating" qualifier though. The guarantee is at the system level, not the hardware level, so on a core that does not do it in hardware every misaligned LDC1 here turns into a kernel trap. LLVM's `-mno-unaligned-access`/`-mstrict-align` support for r6 exists for that reason ("MIPSr6 ISA requires normal load/store instructions support misaligned memory access, while it is not always do so by hardware. On some microarchitectures or some corner cases it may need support by OS"). Correct, but potentially very slow, and the pre-R6 path here deliberately avoided traps, so it probably deserves a comment.

The same appendix also carves atomics out of the guarantee, which lines up with the LL/SC thread: sljit only emits those with a zero displacement off a register, so it never relies on the part of the architecture R6 does not provide.

Unrelated to R6, but on the same hunk: the old line

```c
	FAIL_IF(update_mem_addr(compiler, &mem, &memw, SIMM_MAX - (type & SLJIT_32) ? 3 : 7));
```

has a precedence bug. `-` binds tighter than `?:`, so it parses as `(SIMM_MAX - (type & SLJIT_32)) ? 3 : 7`, which is always `3` because the left operand is never zero. The intent was `SIMM_MAX - 3` or `SIMM_MAX - 7`. The added parentheses fix it. That one affects every pre-R6 MIPS target, i.e. configurations this PR is otherwise not touching, so it should be split out and described on its own.

sljit_src/sljitNativeLOONGARCH_64.c
mattst88: Not my area, but for whoever reviews it: the existing code does look wrong, independent of the restructuring. In the `src & OFFS_REG_MASK` branch, the `SLLI_D` into `TMP_REG1` is conditional on `srcw` being non-zero, but the following `ADD_D ... RK(TMP_REG1)` is unconditional, so when `srcw == 0` it adds an uninitialized `TMP_REG1` to the base. The patch fixes that by handling the `srcw == 0` case explicitly.

On the "return part first" suggestion: agreed, and the `base == 0 && srcw == 0` case reducing to a bare `PRELD` on the index register reads much better as an early return than as the tail of an if/else chain.

zherczeg:
Thank you for the explanation. I can land this patch with the proposed minor changes. It would be good to move the loongarch changes to a separate patch, and wait for its maintainer.

sljit_src/sljitNativeLOONGARCH_64.c
zherczeg: @lrzlin what do you think?

lrzlin: Overall LGTM, nice catch! but the final else branch seems to be redundant?
```C++
else
    return push_inst(compiler, PRELD | RD(0) | RJ(OFFS_REG(src)));
```
It could be removed, we will do this at the bottom of this case.

sljit_src/sljitConfigInternal.h
carenas: FYI, clang support for mips was always problematic (OpenBSD just dropped loongson, which was the last platform that was running mipsel, because clang 22 wouldn't work).

I would recommend using `gcc` based crosscompilers for testing (even older versions work fine, in systems were I couldn't get clang to behave), and need to build for a base of MIPS III (both big and little endian).

zherczeg:
@NWilson there was a small change request for the patch (loongarch part). Do you plan to do it?

carenas:
tested in OpenBSD/Octeon 7.9 using egcc (GCC) 8.4.0:

```
SLJIT tests: all tests PASSED on MIPS III 64bit (big endian + aligned) (with fpu)
```

The system has 4 CPUs with FPU and returns `0x739600` as `fir` and shouldn't skip test70

sljit_src/sljitNativeLOONGARCH_64.c
carenas: @irzlin: can you make a different PR with your change instead, so this part could be dropped from here?, hadn't look at it deeply, but I think you need that last part, and shouldn't be in an `else`, otherwise the prefetch is just a NOP, right?

sljit_src/sljitNativeMIPS_common.c
carenas: I think this breaks anything below R1, because the CLZ instruction was added with r1, and so when used in MIPS III should cause a SIGILL. I didn't see where it is being used, but this likely needs adjusting or at least a comment.

I am not surprised that the other instruction inside the SLJIT_MIPS_REV >= 6 branch wasn't even right. There was never a push to get mipsr6 working, since there are no available CPUs and only a few "distributions" that support it, and as you all have found, it is not even compatible with the older versions. 

The fact that it wouldn't even compile was spelled out in #204

sljit_src/sljitNativeMIPS_common.c
carenas: I don't think this is correct (and inddeed if a CPU without an FPU returns 1, this will cause a crash), but I can't found the documentation I got a hold of long ago.

I think CPU_FEATURE_FR is enabled when 64-bit register are found (which comes probably from `fir` bit 17.

There might be something else to consider (for example, mips III should return 0, as I think the ABI force all floating point registers to be even)

sljit_src/sljitNativeMIPS_common.c
carenas: everything inside a >=R6 was NOT production, so this doesn't need an independent "fix" IMHO

NWilson:
> @NWilson there was a small change request for the patch (loongarch part). Do you plan to do it?

I'm sorry, I can do the changes. It's just not high on my priorities so I haven't done it yet.

sljit_src/sljitNativeLOONGARCH_64.c
mattst88: Loongson fix merged in https://github.com/zherczeg/sljit/pull/371

NWilson:
I have broken out two changes from this PR:
* #372 
* #373

@mattst88 and others on this thread may want to see those changes.

As and when those are reviewed and merged, I'll rebase this PR with the remainder (R6 changes only).

NWilson:
@zherczeg This PR is rebased, and now affects only MIPS R6 (unless I messed up). Because this CPU never really shipped many units, this code is not much use to anyone.

However, it lets Matt's qemu tests pass, so it's worth taking.

NWilson:
GPT confirms I haven't done anything stupid. It says:

...

The new/changed behavior is confined to builds targeting **MIPS Release 6**:
- R6 encodings for LL/LLD/SC/SCD and CLZ/DCLZ,
- R6 prefetch emission,
- R6 `sljit_emit_fmem` availability/implementation,
- R6 multiply-with-overflow instruction ordering,
- R6 float-conversion result defaults.

So your statement is substantively correct, and can be made slightly more precise as:

> “Apart from a comment typo fix, this PR changes generated code and runtime behavior only for builds with `SLJIT_MIPS_REV >= 6`; non-MIPS targets and pre-R6 MIPS retain their prior behavior.”

carenas (approved):
Mostly does what was copy and pasted from GPT, except that the encoding for the CLZ  instruction is no longer only defined for R1 or newer, but also for any legacy CPUs that could trigger SIGILL if they are ever instructed to use it.

PR #370, opened 2026-08-16
aarch64: set bit 14 for encoding [S|U]MULH

carenas:
The documentation[1] and at least GNU `as` matches the new encoding.

[1] https://support.arm.com/documentation/ddi0602/2026-06/Base-Instructions/SMULH--Signed-multiply-high-?lang=en


zherczeg:
New encoding? Why is that changed? Is it compatible with old cpus?

carenas:
I suspect that the old encoding was just a bug that happens to work in real hardware, as it makes sense that it will encode in RA (bits 10 to 14) the value for the ZERO register for the "high" bits multiplication output

Indeed, I haven't found any old documentation indicating that bit was ever 0 (but bit15 is, so maybe it was mistaken originally).

Tested in Cortex A55, M1 and M2 and both encodings seem to work.

carenas:
FWIW, built GNU assembler (GNU Binutils) 2.23.2 (which was the first version that supported aarch64 from 2012) and it uses the "new" encoding as well:

```
Disassembly of section .text:

0000000000000000 <.text>:
   0:	9b4a7c3e 	smulh	x30, x1, x10
```

until this change JIT will output `9b4a3c3e` instead.

zherczeg (approved):
LGTM

PR #371, opened 2026-08-25
loongarch: fix prefetch with an index register and no shift

mattst88:
Split out of #369, which is otherwise about MIPS.

In `sljit_emit_op_src()`, the `SLJIT_PREFETCH_*` path with a `SLJIT_MEM2()`
address only initializes `TMP_REG1` when the offset shift is non-zero, but
feeds it to `ADD_D` unconditionally. With a shift of 0 the prefetch address is
computed from whatever `TMP_REG1` happened to hold.

Add the index register directly when there is no shift, matching the RISC-V
backend (`sljitNativeRISCV_common.c`).

Note that the test suite cannot catch this: a prefetch of a garbage address is
architecturally harmless, so the tests pass either way.

Compared to the version in #369, this drops the handling for `base == 0`.
That case is unreachable: `function_check_src_mem()` rejects an address with
`OFFS_REG` and no base register, and no `SLJIT_MEM2()` form can produce one.
This matches what @lrzlin pointed out in
https://github.com/zherczeg/sljit/pull/369#discussion_r3782167006, and is the
same assumption the RISC-V backend already makes.

I have left the surrounding code alone rather than doing the "put the return
part first" restructuring @zherczeg asked for in
https://github.com/zherczeg/sljit/pull/369#discussion_r3763527894, so that this
stays a minimal bug fix. Happy to fold that in here, or leave it for a separate
cleanup.

Cross-compiled for loongarch64 and tested under qemu; `sljit_test` and
`regex_test` both pass. Disassembly of the generated code for
`SLJIT_MEM2(SLJIT_R0, SLJIT_R1)` with a shift of 0 is now:

```
  14:	0010948d 	add.d       	$t1, $a0, $a1
  18:	2ac001a0 	preld       	0x0, $t1, 0
```


mattst88:
@lrzlin @Xiao-Tao @zherczeg @NWilson -- splitting this out of #369 as requested. I cannot request reviews on this repo, so mentioning you here instead.

zherczeg:
The patch is ok for me

lrzlin:
LGTM, Thanks!

zherczeg (approved):

PR #372, opened 2026-09-08
One small MIPS bugfix in instruction generation

NWilson:
I have split this out of #369 as requested. It affects actual MIPS CPUs (<=R5).

I've also fixed an unused-variable warning on those CPUs.


zherczeg (approved):
LGTM

PR #373, opened 2026-09-08
Fix MIPS runtime FR mode handling

NWilson:
- In autodetection, distinguish native FR mode from FRE compatibility mode
- Fix a typo in the ifdef for PR_GET_FP_MODE
- Fix the build for SLJIT_DETECT_FR==2 (needs 'zero' variable)
- Enable runtime use of MFHC1/MTHC1 paths when build targets R1, but R2 is present at runtime
- Report F64-as-F32 pairs correctly based on the FPU emulation mode

This is also split out of #369, but includes some changes that weren't discussed in that PR.


mattst88:
While reviewing this, I happened to notice that `HAVE_PRCTL` is used in `sljit_src/sljitNativeMIPS_common.c` (added by @carenas in https://github.com/zherczeg/sljit/commit/3dcdb25754363db2cd34dbd1cd9a6a6b880cadae) but I don't see where it's defined by any build system?

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -871,7 +866,7 @@ SLJIT_API_FUNC_ATTRIBUTE sljit_s32 sljit_has_cpu_feature(sljit_s32 feature_type)
 		if (!cpu_feature_list)
 			get_cpu_features();
 
-		return (cpu_feature_list & CPU_FEATURE_FR) != 0;
+		return (cpu_feature_list & (CPU_FEATURE_FPU | CPU_FEATURE_FR)) == CPU_FEATURE_FPU;
```
mattst88, line 870: This was inverted, agreed. `freg_map` maps `SLJIT_F64_SECOND(x)` to hardware register `even+1`, which is the double's high half only under FR=0. Under FR=1 it is an independent register, so the old return value advertised the pairing exactly when it did not hold.

This flips the answer on every MIPS32 host, so FR=0 and FRE machines now run `sljitTestFloat.h:2645` and `sljitTest.c:8544` for the first time. Has the suite been run on an FR=0 target, hardware or qemu?

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -257,9 +257,7 @@ static const sljit_u8 freg_map[SLJIT_NUMBER_OF_FLOAT_REGISTERS + 4] = {
 #define LWR		(HI(38))
 #define LWC1		(HI(49))
 #define MFC1		(HI(17))
-#if defined(SLJIT_MIPS_REV) && SLJIT_MIPS_REV >= 2
 #define MFHC1		(HI(17) | (3 << 21))
```
mattst88, line 260: Dropping the `REV >= 2` guards is safe under autodetection, since `CPU_FEATURE_FR` requires `CPU_FEATURE_FP64`, which comes from `FIR` bit 22, an R2 bit. But the `SLJIT_IS_FPU_AVAILABLE == 64` override at line 416 sets `CPU_FEATURE_FP64` unconditionally, so an R1 build using that override will now emit `MTHC1` on a CPU without it. The old `#if` made that unreachable. Worth either masking `CPU_FEATURE_FR` off when `SLJIT_MIPS_REV < 2`, or commenting the invariant in `get_cpu_features`.

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -442,18 +438,17 @@ static void get_cpu_features(void)
...
-#else /* PR_GET_FP_MODE */
+#ifdef PR_GET_FP_MODE
 		flag = prctl(PR_GET_FP_MODE);
 
-		if (flag > 0)
+		if ((flag & (PR_FP_MODE_FR | PR_FP_MODE_FRE)) == PR_FP_MODE_FR)
 			feature_list |= CPU_FEATURE_FR;
-#endif /* FP_GET_PR_MODE */
+#endif /* PR_GET_FP_MODE */
 #if ((defined(SLJIT_DETECT_FR) && SLJIT_DETECT_FR == 2) \
```
mattst88, line 448: `SLJIT_DETECT_FR` is documented nowhere outside this file, and now that the prctl branch actually compiles, the values matter. What is `== 2` meant to do? It builds the probe in even when prctl is available, but `if (flag < 0)` means it only runs when prctl fails. Fallback, or force-probe?

sljit_src/sljitNativeMIPS_common.c
```diff
@@ -442,18 +438,17 @@ static void get_cpu_features(void)
 #endif /* SLJIT_MIPS_REV >= 5 */
 #else
 		sljit_s32 flag = -1;
```
mattst88, line 440: Pre-existing, but this hunk is the place to fix it: with `PR_GET_FP_MODE` undefined and the probe also excluded (non-GCC, or `__mips < 2`), `flag` is set and never read, tripping `-Wunused-but-set-variable`. Same shape as the `zero` problem this PR fixes.

sljit_src/sljitNativeMIPS_common.c
NWilson: No, sadly I haven't done any hardware or emulation runs yet. I'm hoping CI will catch it. It's clearly broken currently.

sljit_src/sljitNativeMIPS_common.c
NWilson: I don't know, I just wanted to leave things a little better than I found them. SLJIT_DETECT_FR is likely used by no clients, and can be considered effectively dead code (maybe).

sljit_src/sljitNativeMIPS_common.c
NWilson: That's an easy fix, I'll fold it in.

sljit_src/sljitNativeMIPS_common.c
NWilson: The reason I had to remove these R2 guards is because I think they are positively wrong: the code could be compiled for R1 minimum, but run on R2 at runtime, and autodetect the FR features. Hence, runtime support for MTHC1 is needed regardless of what the (minimum-required) MIPS_REV says.

GPT also noticed the issue with SLJIT_IS_FPU_AVAILABLE. However, I chose to just let that one stay as it is.

SLJIT_IS_FPU_AVAILABLE  is an explicit user-override. Customers are very unlikely to be using it, but if they do, then it's their job to set it correctly and plan carefully what CPU they are compiling for and running on.

carenas:
> While reviewing this, I happened to notice that `HAVE_PRCTL` is used in `sljit_src/sljitNativeMIPS_common.c` (added by @carenas in [3dcdb25](https://github.com/zherczeg/sljit/commit/3dcdb25754363db2cd34dbd1cd9a6a6b880cadae)) but I don't see where it's defined by any build system?

if your using a recent enough linux system to build then can use:

```
make EXTRA_CPPFLAGS=-DHAVE_PRCTL
```

mattst88:
> if your using a recent enough linux system to build then can use:
> 
> make EXTRA_CPPFLAGS=-DHAVE_PRCTL

Right... but is that really how the code was intended to be enabled?

NWilson:
> Right... but is that really how the code was intended to be enabled?

Actually, yes.

The sljit project was extracted from PCRE2, with a fairly minimal build system for the independent project. I think there might be some other users, outside of PCRE2, but that is the largest consumer of sljit.

PCRE2 has Autoconf and CMake and systems that can detect `prctl()` and set `HAVE_PRCTL`. But sljit's own Makefile just doesn't do any of that. The sljit CI is more like a smoke test, and it exercises parts of the project that PCRE2 doesn't use.

zherczeg:
Please let me know when the changes are final.

About the history: sljit was created as a simple jit compiler, and pcre used it later. The inspiration come from v8/yarr, which also used jit to accelerate pattern matching. Although it turned out that pcre is far more complex than JavaScript regexp.

NWilson:
I think it's final from me. Matt had some comments about further improvements, but I don't think any of the things he commented were regressions. The existing SLJIT_DETECT_FR and SLJIT_IS_FPU_AVAILABLE overrides are delicate, and I haven't fixed or changed that.

mattst88:
Works for me. I'm happy to follow up myself.

zherczeg (approved):
LGTM

PR #374, opened 2026-09-16
mips: restrict better PREFX to avoid SIGILL

carenas:
Running NetBSD evbmips/mips 11.99.8 in a QEMU emulated malta board with a 4Kc CPU (MIPS32) triggered a SIGILL in test47.

I am little surprised this wasn't raised before as it is even reproducible with QEMU and the Ubuntu cross compiler that was recently added to PCRE2 CI, and the documentation[1] that was quoted when adding support for `R6`, but hopefully the organic refactoring I came with covers all codepaths going forward.

[1] https://docs.alexrp.com/mips/


zherczeg (approved):
LGTM

